PR #1, opened 2021-08-25
Project scaffolding

r00ta:
This PR aims to scaffold the project repository. 

4 separated modules have been added: 
1) manager
2) shard
3) executor
4) runner

The `runner` module aims to use the `manager`, the `shard` and the `executor` modules to build the single application to be deployed. 

See below the project structure

Screenshot from 2021-08-25 09-16-12


.github/workflows/CI.yaml
```diff
@@ -0,0 +1,43 @@
...
+        run: mvn test
+      - name: Publish Test Report
+        if: ${{ always() }}
+        uses: scacap/action-surefire-report@v1.0.10
+        with:
+          skip_publishing: true
+          check_name: Test Results
+          fail_on_test_failures: true
+          fail_if_no_tests: false
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
danielezonca, line 43: This is not accessible when PR starts from a fork (like in this case with `pull_request` event so you can remove this part

.github/workflows/CI.yaml
```diff
@@ -0,0 +1,43 @@
...
+    name: Build and Test
+    steps:
+      - name: Set up GraalVM
+        uses: DeLaGuardo/setup-graalvm@4.0
+        with:
+          graalvm: '21.1.0'
+          java: 'java11'
+      - name: Install native-image Binary
+        shell: bash
+        run: gu install native-image
```
danielezonca, line 21: As far as I can see there is no native compilation so this should not be necessary. Moreover using Quarkus lets you build native in a container [without the need](https://quarkus.io/guides/building-native-image#container-runtime) to install GraalVM

.github/workflows/CI.yaml
```diff
@@ -0,0 +1,43 @@
...
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+      - name: Compile and Validate Formating
+        shell: bash
+        run: mvn -Dmaven.test.skip=true -Pvalidate-formatting package
+      - name: Run Test Suite
+        shell: bash
+        run: mvn test
```
danielezonca, line 34: Let's use `verify` as goal so that it also runs integration tests

build-parent/pom.xml
```diff
@@ -0,0 +1,224 @@
...
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+    <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
+    <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
+    <quarkus.platform.version>2.1.3.Final</quarkus.platform.version>
+    <surefire-plugin.version>3.0.0-M5</surefire-plugin.version>
+    <version.compiler.plugin>3.8.1</version.compiler.plugin>
+    <version.surefire.plugin>2.22.2</version.surefire.plugin> <!-- minimum required by JUnit 5 -->
+    <version.install.plugin>2.5.2</version.install.plugin>
+    <kogito.version>1.10.0.Final</kogito.version>
```
danielezonca, line 36: As far as I can see this version is only used to download the code formatting (while everything else is via quarkus platform bom) so please rename the property to avoid misunderstanding. What about using something like `kogito-formatter.version` or `kogito-codestyle.version`?

pom.xml
```diff
@@ -0,0 +1,21 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.redhat.developer</groupId>
+  <artifactId>event-connect</artifactId>
+  <version>0.0.1</version>
```
danielezonca, line 8: Please use `0.0.1-SNAPSHOT` as version here and in all other poms. Fixed version becomes a problem as soon as we will configure nightly maven deploy

shard/README.md
```diff
@@ -0,0 +1,54 @@
+# shard Project
```
danielezonca, line 1: What about remove ~all the content of this file? This is just the default Quarkus readme that is not really necessary for us. A generic 
```suggestion
# shard Project

WIP
```
Should be enough. Wdyt?

runner/README.md
```diff
@@ -0,0 +1,54 @@
+# runner Project
```
danielezonca, line 1: What about remove ~all the content of this file? This is just the default Quarkus readme that is not really necessary for us. A generic 
```suggestion
# runner Project

WIP
```
Should be enough. Wdyt?

manager/README.md
```diff
@@ -0,0 +1,54 @@
+# manager Project
```
danielezonca, line 1: What about remove ~all the content of this file? This is just the default Quarkus readme that is not really necessary for us. A generic 
```suggestion
# manager Project

WIP
```
Should be enough. Wdyt?

executor/README.md
```diff
@@ -0,0 +1,54 @@
+# executor Project
```
danielezonca, line 1: What about remove ~all the content of this file? This is just the default Quarkus readme that is not really necessary for us. A generic 
```suggestion
# executor Project

WIP
```
Should be enough. Wdyt?

build-parent/pom.xml
```diff
@@ -0,0 +1,224 @@
...
+    <dependencies>
+      <dependency>
+        <groupId>${quarkus.platform.group-id}</groupId>
+        <artifactId>${quarkus.platform.artifact-id}</artifactId>
+        <version>${quarkus.platform.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
```
danielezonca, line 55: I see this import both here and also in all the other poms. It is not necessary this duplication so we can decide if to have this here only or in all the child modules.
Personally I would keep it here only but the drawback is that it will not be possible to simple override the platform with a `-Dquarkus....` command but you will have to change it in the parent. I don't think we expect to test each module with a different version of the platform so it should be safe to keep it only here. Wdyt?

.github/workflows/CI.yaml
```diff
@@ -0,0 +1,43 @@
+name: Event Connect - CI
+on: [pull_request]
+jobs:
+  event-connect-build:
+    env:
```
danielezonca, line 5: Please add [`concurrency` element](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency) to cancel old jobs when pushing new code
```suggestion
    concurrency: 
      group: ${{ github.head_ref }}
      cancel-in-progress: true
    env:
```

r00ta: concurrency element already present

.github/workflows/CI.yaml
```diff
@@ -0,0 +1,39 @@
...
+      cancel-in-progress: true
+    timeout-minutes: 45
+    runs-on: ubuntu-latest
+    name: Build and Test
+    steps:
+      - name: Set up GraalVM
+        uses: DeLaGuardo/setup-graalvm@4.0
+        with:
+          graalvm: '21.1.0'
+          java: 'java11'
```
danielezonca, line 18: Please replace this step with [these other steps](https://github.com/kiegroup/kogito-runtimes/blob/main/.github/workflows/apps-pr.yml#L31-L51) to:
- free a lot of space on the executor
- setup java 11 with the "blessed" `setup-java` action

danielezonca (approved):
Just a final comment but LGTM

robpblake (approved):

PR #2, opened 2021-08-25
Scaffolding addons

r00ta:
This PR aims to configure the first bunch of addons in the modules: 

1) manager: 
  - postgres
  - prometheus
  - swagger-ui
2) shard
  - prometheus
  - swagger-ui
3) executor
  - prometheus
  - swagger-ui


manager/pom.xml
```diff
@@ -27,12 +28,45 @@
...
 
     <dependency>
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-junit5</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-jdbc-h2</artifactId>
+      <scope>test</scope>
```
danielezonca, line 68: I would prefer to use (maybe using testcontainer or similar) the _real_ db also during the test instead of a different and in memory one.
It should be enough to add the following dependencies
```xml
<dependency>
    <groupId>org.testcontainers</groupId>
    <artifactId>testcontainers</artifactId>
    <version>1.16.0</version>
    <scope>test</scope>
</dependency>
<dependency>
    <groupId>org.testcontainers</groupId>
    <artifactId>postgresql</artifactId>
    <version>1.16.0</version>
    <scope>test</scope>
</dependency>
```

executor/pom.xml
```diff
@@ -22,6 +23,18 @@
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-resteasy</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-resteasy-jackson</artifactId>
+    </dependency>
+    <dependency>
```
robpblake, line 30: What's the need for OpenAPI and swagger-ui on the Executor and Shard?

r00ta: @robpblake I think the executor and the shard will have rest api, wdyt?

robpblake: I can only see the need for a REST APIs on the Manager (same as we did for DaaS). I imagine there will be two on the Manager:

- Customer Facing API that people use to create things
- API that the Shard(s) use to fetch details of what they need to deploy
  - For the first iteration, we may not even expose this one as a REST API as at Manager and Shard are deployed as a single unit, so it may just be directly injected as a bean

I don't see there being a REST API on the Shard or Executor. What's the use case you're seeing for a REST API on Shard + Executor?

PR #3, opened 2021-08-31
MGDOBR-1

r00ta:
https://issues.redhat.com/projects/MGDOBR/issues/MGDOBR-1?filter=allopenissues

This PR aims to implement MGDOBR-1 and the following project architecture: 
1) The user can call POST `/connectors` to create a connector
2) The user can call GET `/connectors` to get the list of the connectors created with their status (REQUESTED, PROVISIONING, AVAILABLE)
3) The manager stores the connector information into the postgres database
4) The shard asks the manager every 30s the connectors to be deployed
5) With an async implementation, the shard deploys the Ingress (mocked at the minute, will be replaced with specific CRD) and sends the status update of the connectors (PROVISIONING)
6) Once the  Ingress has been deployed (mocked at the minute) the shard sends the status update of the connector (AVAILABLE)
7) The ingress exposes an endpoint where the user can submit events only if they are valid Cloud Events. The ingress wraps the event with another Cloud Event (where it stores some internal information) and then sends it to a kafka topic.

Tests have been added together with a docker-compose to play with the application locally. 

In order to start it, run `mvn clean install -DskipTests && mvn clean compile -f runner/pom.xml quarkus:dev`

In conclusion, this PR implements MGDOBR-1 and prepare the work for the following tasks.

This is the project structure
Screenshot from 2021-09-01 14-03-30


shard/src/main/java/com/redhat/developer/shard/ManagerSyncServiceImpl.java
```diff
@@ -0,0 +1,84 @@
...
+    Vertx vertx;
+
+    @Inject
+    OperatorService operatorService;
+
+    WebClient webClient;
+
+    @PostConstruct
+    void init() {
+        webClient = WebClientUtils.getClient(vertx, eventConnectManagerUrl);
```
robpblake, line 45: I think it makes more sense to inject the `WebClient` from the IoC context via a `Producer` rather than doing an `@PostConstruct` here.

manager/src/main/java/com/redhat/developer/manager/ConnectorsServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+public class ConnectorsServiceImpl implements ConnectorsService {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ConnectorsServiceImpl.class);
+
+    @Inject
+    ConnectorDAO connectorDAO;
+
+    @Override
+    public Connector createConnector(String customerId, ConnectorRequest connectorRequest) {
+        if (connectorDAO.findByNameAndCustomerId(connectorRequest.getName(), customerId).size() > 0) {
+            throw new AlreadyExistingItemException("Element already present in storage.");
```
robpblake, line 28: It would be nice to have a clearer error message e.g. `Connector with name 'foo' already exists for customer with id 'bar'`

manager/src/main/resources/application.properties
```diff
@@ -26,4 +26,5 @@ quarkus.flyway.create-schemas=true
 %dev.quarkus.datasource.username=event-connect
 %dev.quarkus.datasource.password=event-connect
 %dev.quarkus.hibernate-orm.log.sql=true
-%dev.quarkus.hibernate-orm.log.bind-parameters=true
\ No newline at end of file
+%dev.quarkus.hibernate-orm.log.bind-parameters=true
+%dev.quarkus.hibernate-orm.database.generation=drop-and-create
```
robpblake, line 30: We should move this to use Flyway. We can copy the setup and approach that we had from DaaS, apart from using PostGres rather than H2.

There is a section in the DaaS Fleet Manager README.md that I think we can use for our approach to Flyway: https://github.com/kiegroup/baaas-decision-fleet-manager#database-development

WDYT?

manager/src/main/java/com/redhat/developer/manager/requests/ConnectorRequest.java
```diff
@@ -0,0 +1,19 @@
+package com.redhat.developer.manager.requests;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.redhat.developer.manager.models.Connector;
+
+public class ConnectorRequest {
```
robpblake, line 6: At some point we'll probably want to break this out into an `api` package e.g `eventthing-api` so that users of the service can easily import just the public facing API into their Java apps.

So we might want to be a little careful about having the internal domain model in the public facing API. It's not a huge probably right now, but it might be easier to start with a mapper that converts the API objects into internal domain objects and vice-versa.

manager/src/main/java/com/redhat/developer/manager/dao/ConnectorDAO.java
```diff
@@ -0,0 +1,29 @@
...
+@Transactional
+public class ConnectorDAO implements PanacheRepositoryBase<Connector, String> {
+
+    public List<Connector> findByStatus(ConnectorStatus status) {
+        Parameters params = Parameters
+                .with("status", status);
+        return find("#CONNECTOR.findByStatus", params).list();
+    }
+
+    public List<Connector> findByNameAndCustomerId(String name, String customerId) {
```
robpblake, line 24: I think this should return a `Connector`. We should add a unique constraint on the DB level that we can't store two Connectors with the same `customerId` and `name` so this query either returns `null` or the `Connector`.

manager/src/main/java/com/redhat/developer/manager/api/user/ConnectorsAPI.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Inject
+    CustomerIdResolver customerIdResolver;
+
+    @Inject
+    ConnectorsService connectorsService;
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getConnectors() {
```
robpblake, line 31: We have the requirement to follow the guidelines here https://gitlab.cee.redhat.com/service/api-guidelines for any API we expose to our customers. So at some point I think we're going to want to create a `ConnectorsResponse` object that includes things like the HATEOS links, versions etc.

So I'd suggest we do that now. Take a look at the guidelines I've linked ^ and also how we did it in DaaS here: https://github.com/kiegroup/baaas-decision-fleet-manager/tree/main/baaas-decision-fleet-manager-api/src/main/java/org/kie/baaas/dfm/api/decisions

You can then see how we mapped between API objects and internal domain model in DaaS here: https://github.com/kiegroup/baaas-decision-fleet-manager/blob/main/baaas-decision-fleet-manager/src/main/java/org/kie/baaas/dfm/app/controller/modelmappers/DecisionMapper.java

manager/src/main/java/com/redhat/developer/manager/models/ConnectorStatus.java
```diff
@@ -0,0 +1,41 @@
...
+    PROVISIONING("PROVISIONING"),
+    AVAILABLE("AVAILABLE");
+
+    private String status;
+
+    ConnectorStatus(String status) {
+        this.status = status;
+    }
+
+    public ConnectorStatusDTO toDTO() {
```
robpblake, line 16: This feels like a bit of a code smell...is there a reason we can't use a single enum?

pom.xml
```diff
@@ -16,6 +16,8 @@
     <module>runner</module>
     <module>bom</module>
     <module>build-parent</module>
+    <module>ingress</module>
+    <module>infra</module>
```
robpblake, line 20: Is the `infra` module the Fleet Shard?

shard/src/main/java/com/redhat/developer/shard/OperatorServiceInMemoryImpl.java
```diff
@@ -0,0 +1,53 @@
...
+    ManagerSyncService managerSyncService;
+
+    //TODO: remove after we move to k8s
+    @Inject
+    IngressService ingressService;
+
+    @Override
+    public ConnectorDTO createConnectorDeployment(ConnectorDTO connector) {
+        connectors.add(connector);
+        LOGGER.info("[shard] Request deployment of Connector with ID " + connector.getId() + "... MOCKED!");
```
robpblake, line 35: Can we add some more info to the logging messages to make it easier for us to debug e.g:

`Processing deployment of Connector with id 'foo' and name 'bar' for customer 'baz'`

shard/src/main/java/com/redhat/developer/shard/utils/WebClientUtils.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.net.URI;
+
+import io.vertx.ext.web.client.WebClientOptions;
+import io.vertx.mutiny.core.Vertx;
+import io.vertx.mutiny.ext.web.client.WebClient;
+
+public class WebClientUtils {
+
+    public static WebClient getClient(Vertx vertx, String url) {
```
robpblake, line 11: Does Quarkus offer us an out-of-the-box way to configure the `WebClient`?

robpblake (commented):
Thanks @r00ta . Looks like a great start!

Quite a few comments from me, but this is a big PR.

cheers,
Rob

manager/src/test/java/com/redhat/developer/manager/ConnectorsServiceTest.java
```diff
@@ -0,0 +1,87 @@
...
+    @TestTransaction
+    public void testGetEmptyConnectorsToDeploy() {
+        List<Connector> connectors = connectorsService.getConnectorsToDeploy();
+        Assertions.assertEquals(0, connectors.size());
+    }
+
+    @Test
+    @TestTransaction
+    public void testGetEmptyConnectors() {
+        List<Connector> connectors = connectorsService.getConnectors("jrota");
```
robpblake, line 33: Can you inject the `CustomerIdResolver` into the test class and call that to get the `id` rather than hard-coding `jrota` directly

ingress/src/main/java/com/redhat/developer/ingress/api/IngressAPI.java
```diff
@@ -0,0 +1,36 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(IngressAPI.class);
+
+    @Inject
+    IngressService ingressService;
+
+    @POST
+    @Path("/events/{name}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    public Response notifyDeployment(@PathParam("name") @NotNull String name, @NotNull JsonNode event) {
```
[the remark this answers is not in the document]
r00ta: It was a copy and paste mistake

infra/src/main/java/com/redhat/developer/infra/utils/CloudEventUtils.java
```diff
@@ -0,0 +1,102 @@
...
+import io.cloudevents.CloudEvent;
+import io.cloudevents.core.builder.CloudEventBuilder;
+import io.cloudevents.jackson.JsonCloudEventData;
+import io.cloudevents.jackson.JsonFormat;
+
+public class CloudEventUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CloudEventUtils.class);
+
+    public static final String UNKNOWN_SOURCE_URI_STRING = urlEncodedStringFrom("__UNKNOWN_SOURCE__")
```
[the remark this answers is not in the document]
r00ta: Copy and paste leftover

infra/src/main/java/com/redhat/developer/infra/dto/ConnectorDTO.java
```diff
@@ -0,0 +1,60 @@
+package com.redhat.developer.infra.dto;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class ConnectorDTO {
```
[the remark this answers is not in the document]
r00ta: Currently I return this DTO in the "shard" and in the "user" api.

manager/src/main/resources/application.properties
r00ta: Actually I tried to remove this to let flyway create the schemas but it did not work out. Do you have any suggestion? I've also looked at the baaas stuff but the configuration is the same (except for this line)

manager/src/main/java/com/redhat/developer/manager/dao/ConnectorDAO.java
r00ta: The unique constraint is already present at schema level

pom.xml
r00ta: Nope, it is a module that contains DTOs and other common utilities. The `shard` module is called `shard`

manager/src/test/java/com/redhat/developer/manager/ConnectorsServiceTest.java
r00ta: The `CustomerIdResolver` is just hardcoding `jrota` for the time being, but the idea is that we will use it to extract the customerId from the JWT token and use it in the platform. 
I think the tests should not call `CustomerIdResolver`, otherwise we will have to change it everywhere when we do this change

ingress/src/main/java/com/redhat/developer/ingress/producer/KafkaEventPublisher.java
```diff
@@ -0,0 +1,35 @@
...
+public class KafkaEventPublisher {
+    private final Logger logger = Logger.getLogger(KafkaEventPublisher.class);
+
+    @Inject
+    @Channel("events-out")
+    Emitter<String> emitter;
+
+    public boolean sendEvent(CloudEvent cloudEvent) {
+        logger.info("Sending event to event queue");
+        Optional<String> serializedCloudEvent = CloudEventUtils.encode(cloudEvent);
```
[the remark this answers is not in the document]
r00ta: This `CloudEventUtils` is a class copied from https://github.com/kiegroup/kogito-runtimes/blob/main/addons/common/messaging/utils/src/main/java/org/kie/kogito/cloudevents/CloudEventUtils.java . I will change it a bit

manager/src/main/resources/application.properties
robpblake: Is there any log from Flyway at all when Quarkus starts?

shard/src/main/java/com/redhat/developer/shard/utils/WebClientUtils.java
r00ta: Actually it is not quarkus but vertx. Anyway nope afaik

manager/src/main/java/com/redhat/developer/manager/api/user/ConnectorsAPI.java
r00ta: > We have the requirement to follow the guidelines here https://gitlab.cee.redhat.com/service/api-guidelines for any API we expose to our customers. So at some point I think we're going to want to create a `ConnectorsResponse` object that includes things like the HATEOS links, versions etc.
> 
> So I'd suggest we do that now. 

So let me summarize what we have to implement in this PR to follow the guidelines:
1) authentication 
2) Representation (kind, id, href) with a reference to the error description - something we did not have in baaas
3) pagination 
4) search 
5) Documentation with OpenAPI

From my side it is not a big deal to include those in this PR, but, If we want to do this in this PR, I suggest we edit the MGDOBR-1 ticket so to include all of those tasks. This PR will become bigger and I need probably some more days to do the job. 
Let me know how you prefer to proceed @robpblake

ingress/src/main/java/com/redhat/developer/ingress/IngressServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+    KafkaEventPublisher kafkaEventPublisher;
+
+    @Override
+    public boolean processEvent(String name, JsonNode event) {
+        //TODO: remove after we move to k8s
+        if (!deployments.contains(name)) {
+            throw new RuntimeException("Ingress with name " + name + " is not deployed.");
+        }
+
+        CloudEvent cloudEvent = CloudEventUtils.build(UUID.randomUUID().toString(), name,
```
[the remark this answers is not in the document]
r00ta: Actually the idea was to wrap the ingress event with our cloud event so to add more information about the event itself useful for our purposes (for example when it was inserted etc.. ). I've implemented the requirement that ALSO the ingress event must be a cloud event. 
If you think we should just take the event and send it to kafka directly let me know @robpblake

manager/src/main/resources/application.properties
r00ta: @robpblake 
```
2021-09-01 11:14:21,909 INFO  [org.fly.cor.int.lic.VersionPrinter] (Quarkus Main Thread) Flyway Community Edition 7.11.2 by Redgate
2021-09-01 11:14:21,910 INFO  [org.fly.cor.int.dat.bas.BaseDatabaseType] (Quarkus Main Thread) Database: jdbc:postgresql://localhost:5432/event-connect (PostgreSQL 13.1)
2021-09-01 11:14:21,929 INFO  [org.fly.cor.int.com.DbValidate] (Quarkus Main Thread) Successfully validated 0 migrations (execution time 00:00.007s)
2021-09-01 11:14:21,929 WARN  [org.fly.cor.int.com.DbValidate] (Quarkus Main Thread) No migrations found. Are your locations set up correctly?
2021-09-01 11:14:21,938 INFO  [org.fly.cor.int.com.DbMigrate] (Quarkus Main Thread) Current version of schema "public": << Empty Schema >>
2021-09-01 11:14:21,941 INFO  [org.fly.cor.int.com.DbMigrate] (Quarkus Main Thread) Schema "public" is up to date. No migration necessary.
```

r00ta: Ah wait... I see that in baaas we have the manual schema creation...

manager/src/test/java/com/redhat/developer/manager/ConnectorsServiceTest.java
r00ta: Moved to a constant btw

r00ta:
Hi @robpblake , I've resolved most of the comments and I'm waiting on some others. Let me know wdyt

infra/src/main/java/com/redhat/developer/infra/dto/ConnectorDTO.java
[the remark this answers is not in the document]
robpblake: Yeah we have the requirement to return something that is in alignment with the RH API specs for the user facing API. For the internal API that is consumed by the Fleet Shard(s), we can do what we want so this DTO is fine there.

infra/src/main/java/com/redhat/developer/infra/utils/CloudEventUtils.java
```diff
@@ -0,0 +1,69 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.redhat.developer.infra.utils.exceptions.CloudEventDeserializationException;
+import com.redhat.developer.infra.utils.exceptions.CloudEventSerializationException;
+
+import io.cloudevents.CloudEvent;
+import io.cloudevents.core.builder.CloudEventBuilder;
+import io.cloudevents.jackson.JsonCloudEventData;
+import io.cloudevents.jackson.JsonFormat;
+
+public class CloudEventUtils {
```
robpblake, line 19: Any reason not to make this an `@ApplicationScope` bean? That way we could inject the `ObjectMapper` into this class and then inject the `CloudEventUtils` where it is needed. Removes the need for `static` methods and gives us more control over our usage of `ObjectMapper` throughout the code.

ingress/src/main/java/com/redhat/developer/ingress/RegisterCloudEventModuleCustomizer.java
```diff
@@ -0,0 +1,16 @@
...
+
+import javax.inject.Singleton;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import io.cloudevents.jackson.JsonFormat;
+import io.quarkus.jackson.ObjectMapperCustomizer;
+
+@Singleton
+public class RegisterCloudEventModuleCustomizer implements ObjectMapperCustomizer {
```
robpblake, line 11: This would be used in combination with my comment about making CloudEventUtil a managed bean in the CDI context. We'd keep the registration of the CloudEvent Jackson Module in a single location.

ingress/src/main/java/com/redhat/developer/ingress/api/exceptions/IngressRuntimeException.java
```diff
@@ -0,0 +1,19 @@
+package com.redhat.developer.ingress.api.exceptions;
+
+import javax.ws.rs.core.Response;
+
+public class IngressRuntimeException extends IngressException {
```
robpblake, line 5: Do we need this? `IngressException` is already a `RuntimeException`.

infra/src/main/java/com/redhat/developer/infra/utils/exceptions/CloudEventSerializationException.java
```diff
@@ -0,0 +1,12 @@
+package com.redhat.developer.infra.utils.exceptions;
+
+public class CloudEventSerializationException extends RuntimeException {
```
robpblake, line 3: Should this extend `IngressException`?

ingress/src/main/java/com/redhat/developer/ingress/IngressServiceImpl.java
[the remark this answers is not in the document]
robpblake: @r00ta So the original `CloudEvent` is set as the `data` field on the a new `CloudEvent` before being sent to Kafka? Have I followed correctly?

ingress/src/main/java/com/redhat/developer/ingress/IngressServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+import com.redhat.developer.infra.utils.CloudEventUtils;
+import com.redhat.developer.ingress.api.exceptions.IngressRuntimeException;
+import com.redhat.developer.ingress.producer.KafkaEventPublisher;
+
+import io.cloudevents.CloudEvent;
+
+@ApplicationScoped
+public class IngressServiceImpl implements IngressService {
+
+    private static final URI URI_PRODUCER = URI.create("ingreservice/IngressServiceImpl");
```
robpblake, line 20: Small typo `ingreservice` I think should be `ingressservice` or `ingress_service` perhaps for readability.

ingress/src/main/java/com/redhat/developer/ingress/producer/KafkaEventPublisher.java
```diff
@@ -0,0 +1,38 @@
...
+
+    private final Logger logger = Logger.getLogger(KafkaEventPublisher.class);
+
+    private final BroadcastProcessor<String> eventSubject = BroadcastProcessor.create();
+
+    public void sendEvent(CloudEvent cloudEvent) {
+        logger.info("Sending event to event queue");
+        String serializedCloudEvent;
+        try {
+            serializedCloudEvent = CloudEventUtils.encode(cloudEvent);
```
robpblake, line 26: We can make `CloudEventUtils` throw a subclass of `IngressException` then we don't need to catch and re-throw here.

ingress/src/main/java/com/redhat/developer/ingress/producer/KafkaEventPublisher.java
```diff
@@ -0,0 +1,38 @@
...
+
+@ApplicationScoped
+public class KafkaEventPublisher {
+
+    private final Logger logger = Logger.getLogger(KafkaEventPublisher.class);
+
+    private final BroadcastProcessor<String> eventSubject = BroadcastProcessor.create();
+
+    public void sendEvent(CloudEvent cloudEvent) {
+        logger.info("Sending event to event queue");
```
robpblake, line 23: Might be useful to have some detail about the event we're sending in this log message to help us debug. Is there an `id` or something else than can help us uniquely identify the event we're sending?

manager/src/main/java/com/redhat/developer/manager/ConnectorsServiceImpl.java
robpblake: @r00ta I still think this is a valid comment. Do you disagree?

manager/src/test/java/com/redhat/developer/manager/ConnectorsServiceTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Inject
+    ConnectorsService connectorsService;
+
+    @Inject
+    DatabaseManagerUtils databaseManagerUtils;
+
+    @BeforeEach
+    public void cleanUp() {
+        databaseManagerUtils.cleanDatabase();
```
robpblake, line 29: Any particular reason you needed to add this over using `@TestTransaction`?

shard/src/main/java/com/redhat/developer/shard/ManagerSyncServiceImpl.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private List<ConnectorDTO> deserializeConnectors(String s) {
+        try {
+            return mapper.readValue(s, new TypeReference<List<ConnectorDTO>>() {
+            });
+        } catch (JsonProcessingException e) {
+            LOGGER.warn("[shard] failed to deserialize connectors to deploy", e);
+        }
+        return null;
```
robpblake, line 73: If I read the code correctly, this will cause a NPE in the `fetchAndProcessConnectorsFromManager` if this happens. I'd say let's just throw one of our own `RuntimeExceptions` with the `JsonProcessingException` as the `cause`. That way we get a nice big stack trace in the log with the additional NPE.

ingress/src/main/java/com/redhat/developer/ingress/IngressServiceImpl.java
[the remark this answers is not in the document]
r00ta: Yes, correct

shard/src/test/java/com/redhat/developer/shard/ManagerSyncServiceTest.java
```diff
@@ -0,0 +1,114 @@
...
+        stubConnectorsToDeploy(connectorDTOs);
+        stubConnectorUpdate();
+        String expectedJsonUpdateRequest = "{\"id\": \"myId-1\", \"name\": \"myName-1\", \"endpoint\": \"myEndpoint\", \"customerId\": \"myCustomerId\", \"status\": \"PROVISIONING\"}";
+
+        CountDownLatch latch = new CountDownLatch(2); // Two updates to the manager are expected
+        addConnectorUpdateRequestListener(latch);
+
+        managerSyncService.fetchAndProcessConnectorsFromManager().await().atMost(Duration.ofSeconds(5));
+
+        Assertions.assertTrue(latch.await(30, TimeUnit.SECONDS));
```
robpblake, line 67: Anyway we can reduce this time for the tests? We're going to be running these a lot, so having to wait 30 seconds each time is going to add up over time.

robpblake (commented):
@r00ta This is really nice work 🔥 

Just some minor comments from me, but looking good 👍

infra/src/main/java/com/redhat/developer/infra/utils/exceptions/CloudEventSerializationException.java
r00ta: The idea of the CloudEventUtils and its related exceptions (CloudEventSerializationException, CloudEventDeserializationException) is that they are re-usable in different modules (`ingress`, `executor` and others if needed). According to this design, this class cloud not extend `IngressException`, wdyt?

shard/src/test/java/com/redhat/developer/shard/ManagerSyncServiceTest.java
r00ta: The time you see in this test is the maximum time we wait for the latch to go to `0`. So actually the test runs very fast, we never wait 30 seconds!

ingress/src/main/java/com/redhat/developer/ingress/IngressServiceImpl.java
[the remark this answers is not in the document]
robpblake: OK, thanks for the explanation 👍 

I like your idea of adding metadata related to our handling to the received Event. I am not 100% sure on the idea of putting the original event into the `data` field of a new CloudEvent as I wonder if it makes our filtering logic a bit more complex later on? I also wonder if it will make things easier for us to preserve the original `source`, `type` and `id` on the CloudEvent. 

The other approach for metadata I'm thinking about is that we add a set of namespaced attributes to the received CloudEvent e.g:

- com.redhat.eventthing/time-received
- com.redhat.eventthing/another-useful-attribute

This would leave the original CloudEvent intact.

My instinct right now is to just send the `CloudEvent` as-is onto Kafka and then we can revisit that decision once we get further into the implementation.

WDYT?

infra/src/main/java/com/redhat/developer/infra/utils/CloudEventUtils.java
r00ta: The idea is the following: this util class is inside the `infra` package so that it can be used by every module that needs to encode/decode a cloud event (for example `ingress` and `executor`). 
If the module that imports it does not register the  `cloudEventJacksonModule` in the ObjectMapper, this class will not work properly: this is why I would keep this class in this package with its own ObjectMapper. 
The other modules can register or not the `cloudEventJacksonModule` without impacting this utility class. 
For example, the ingress do register it so that we can get the CloudEvent obj directly in the rest endpoint and let jackson deserialize it like magic. But, probably the `executor` do not need rest api and will not work with the serialization of the cloud event outside this class: it will probably not register the  `cloudEventJacksonModule` in its object mapper. 

In conclusion, for the current design I think we should keep this class as it is, self-contained. wdyt?

shard/src/test/java/com/redhat/developer/shard/ManagerSyncServiceTest.java
robpblake: OK thanks 👍

ingress/src/main/java/com/redhat/developer/ingress/producer/KafkaEventPublisher.java
r00ta: Related to the comments above: `CloudEventUtils` is an utility in the `infra` package that will be used by the `executor` as well so we should not throw `IngressException` or similar in the utility class

infra/src/main/java/com/redhat/developer/infra/utils/CloudEventUtils.java
robpblake: Ah OK. That makes sense to me. Thanks.

infra/src/main/java/com/redhat/developer/infra/utils/exceptions/CloudEventDeserializationException.java
```diff
@@ -0,0 +1,12 @@
+package com.redhat.developer.infra.utils.exceptions;
+
+public class CloudEventDeserializationException extends RuntimeException {
```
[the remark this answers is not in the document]
robpblake: I wonder if this should extend `Exception` rather than `RuntimeException` otherwise we'll probably forget to catch the `RuntimeException` when we use it elsewhere in the code?

manager/src/test/java/com/redhat/developer/manager/ConnectorsServiceTest.java
r00ta: @robpblake yes, `@TestTransaction` is pretty dangerous. See here https://github.com/quarkusio/quarkus/issues/15436 . In particular we have tests with restassured -> the transaction is gone and it is not rolled back when the test is over. This is why we have to clean the database before each test. 
If we had only this `ConnectorsServiceTest` we could remove this `@BeforeEach` and add `@TestTransaction` to all the tests, but since we have other tests that leave data in the database the `@BeforeEach` approach is more safe imo. 

Unless we add also a 
```
    @AfterEach
    public void cleanUp() {
        databaseManagerUtils.cleanDatabase();
```
in the restassured test classes or a 
```
    @BeforeAll
    public void cleanUp() {
        databaseManagerUtils.cleanDatabase();
```
in this class we can not use `@TestTransaction`. We might give a try to this https://github.com/radcortez/flyway-junit5-extensions plugin that should clean up the db after each test. 
My main concern is about how do we pass this information to those who are not 100% familiar with @TestTransaction and the testing framework that does not start a new fresh database after each test. wdyt?

robpblake: Your solution works for me 👍  

I'd say let's just add a bit of info into the `DatabaseManagerUtils` class as Javadoc or comments as to why it exists, capturing your points ^. That way when new team members come along, they don't have to ask the same question as me.

manager/src/main/java/com/redhat/developer/manager/ConnectorsServiceImpl.java
r00ta: Sorry, I probably closed it by mistake. Fixed now

ingress/src/main/java/com/redhat/developer/ingress/api/exceptions/IngressRuntimeException.java
r00ta: removed

ingress/src/main/java/com/redhat/developer/ingress/producer/KafkaEventPublisher.java
r00ta: done

shard/src/main/java/com/redhat/developer/shard/ManagerSyncServiceImpl.java
r00ta: done

manager/src/test/java/com/redhat/developer/manager/ConnectorsServiceTest.java
r00ta: Makes sense, done

infra/src/main/java/com/redhat/developer/infra/utils/exceptions/CloudEventSerializationException.java
r00ta: It depends.. some people do not like checked exceptions and some others do. 

My personal preference is always unchecked exception. But if you think this is a good fit for a checked exception, I can change it. Let me know

ingress/src/main/java/com/redhat/developer/ingress/IngressServiceImpl.java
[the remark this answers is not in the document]
r00ta: My idea was to use the cloud event envelop just for the transport layer. The `executor` would then extract the user/incoming/real cloud event and apply the logic there. 

I'm pretty sure we will have to add some metadata to the event sooner or later, and we can achieve that wrapping the original event (like in this proposal) or modifying the original event adding some metadata there. 
Personally, I prefer the first option like I did in this PR, but I'm fine to postpone it when we will have the need (in case). 

I'll change the implementation so to send the original event as it is to kafka

infra/src/main/java/com/redhat/developer/infra/utils/exceptions/CloudEventSerializationException.java
robpblake: Let's not worry about it for now 👍

infra/src/main/java/com/redhat/developer/infra/utils/exceptions/CloudEventDeserializationException.java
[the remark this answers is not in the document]
r00ta: See comment above

manager/pom.xml
```diff
@@ -63,15 +66,8 @@
...
-      <artifactId>testcontainers</artifactId>
-      <version>1.16.0</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.testcontainers</groupId>
-      <artifactId>postgresql</artifactId>
-      <version>1.16.0</version>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-jdbc-h2</artifactId>
```
r00ta, line 70: fyi I've removed test container since I've managed to run the tests without any mock for the persistence layer using h2, which is much faster than running a testcontainer

manager/src/main/java/com/redhat/developer/manager/api/user/ConnectorsAPI.java
r00ta: I've added the fields according to the specification. the `href` field is present in the definition but not populated yet because I'll implement the GET /connectors/{id} endpoint in the next pull request (this one is already big)

robpblake: Thanks @r00ta. Works for me 👍

robpblake (approved):
Looks good to me @r00ta . Nicely done!

shard/pom.xml
```diff
@@ -38,5 +66,11 @@
       <artifactId>rest-assured</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.tomakehurst</groupId>
+      <artifactId>wiremock-jre8</artifactId>
+      <version>2.30.1</version>
```
danielezonca, line 72: DependencyManagement

manager/pom.xml
danielezonca: How big is this difference? Because it could be slower but at least we prevent errors related to different SQL syntax and similar that we already faced in the past. Personally I would probably prefer to use _the real_ env

shard/src/test/resources/application.properties
```diff
@@ -0,0 +1 @@
+quarkus.scheduler.enabled=false
```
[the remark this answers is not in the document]
r00ta: @danielezonca The scheduled methods (for example the job that fetches the Bridges to deploy every 30s) are tested with specific tests, the quarkus application should not start them otherwise they might mess up the interactions with the mocks/update the status of the objects without control

ingress/src/main/java/com/redhat/developer/ingress/IngressService.java
```diff
@@ -0,0 +1,13 @@
+package com.redhat.developer.ingress;
+
+import io.cloudevents.CloudEvent;
+
+public interface IngressService {
+
+    void processEvent(String name, CloudEvent event);
+
+    // TODO: remove after we move to k8s
+    String deploy(String name);
```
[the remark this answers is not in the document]
r00ta: @danielezonca It is a mock because the `Ingress` application will be deployed by k8s according to a specific `Deployment` or CRD. 
I expect that the "core" of the ingress application will remain the same: it will receive a cloud event on a specific endpoint and then it will push it to a kafka topic. I commented all the "infrastructural" parts that will be handed over to k8s, for example the deployment of the instance (in this class we keep a list of "deployed" instances, and if the user tries to push an event to an undeployed instance we return an error. When we move to k8s, we just delete all this logic because the application is deployed by k8s)

manager/pom.xml
r00ta: Honestly I did not measure it, but **it is** faster because h2 is running in memory and testcontainer has to start a docker container (after having downloaded it in case), init the application and so on and so forth. I've created https://issues.redhat.com/browse/MGDOBR-26 so that we can move forward with this PR, is that ok for you @danielezonca ?

infra/src/main/java/com/redhat/developer/infra/utils/CloudEventUtils.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(CloudEventUtils.class);
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper().registerModule(JsonFormat.getCloudEventJacksonModule());
+
+    public static CloudEvent build(String id, String topic, URI source, String subject, JsonNode data) {
```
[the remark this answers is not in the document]
r00ta: Created https://issues.redhat.com/browse/MGDOBR-27

danielezonca (approved):

ingress/src/main/java/com/redhat/developer/ingress/IngressService.java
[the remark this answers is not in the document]
danielezonca: ok thanks for the clarification 👍

PR #6, opened 2021-09-08
MGDOBR-25: create CD pipeline

r00ta:
Jira: https://issues.redhat.com/browse/MGDOBR-25

This PR aims to add the CD pipeline so to create a new docker image when a PR is merged into main. Since at the moment we have one simple application that contains multiple modules, the pipeline will only create one image called `event-bridge-all-in-one`. 

When the different modules will be moved to different and separated applications we will deprecate `event-bridge-all-in-one` and create ad-hoc images.

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -3,6 +3,15 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - 'LICENSE'
+      - '**/.gitignore'
+      - '**.md'
+      - '**.adoc'
+      - '*.txt'
+      - '.github/*'
+      - 'kustomize/*'
+      
 jobs:
   event-bridge-build:
     env:
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -9,8 +9,8 @@ on:
       - '**.md'
       - '**.adoc'
       - '*.txt'
-      - '.github/*'
-      - 'kustomize/*'
+      - '.github/**'
+      - 'kustomize/**'
       
 jobs:
   event-bridge-build:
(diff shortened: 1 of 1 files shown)


robpblake:
@r00ta Is it possible to exclude the container image build if the only changeset in the merge to `main` is to the Kustomize templates?

.github/workflows/CD.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Cache Maven Packages
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+      - name: Build JARs
+        shell: bash
+        run: mvn --batch-mode package -Dmaven.test.skip=true -Dcheckstyle.skip
```
danielezonca, line 46: What about create a `quickly` profile like in Kogito/Quarkus that disable all checks? For example it can skip code formatting etc

.github/workflows/CD.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Login to Container Registry
+        shell: bash
+        env:
+          CONTAINER_REGISTRY_PASSWORD: ${{ secrets.CONTAINER_REGISTRY_PASSWORD }}
+          CONTAINER_REGISTRY_USER: ${{ secrets.CONTAINER_REGISTRY_USER }}
+        run: echo "$CONTAINER_REGISTRY_PASSWORD" | docker login quay.io --username $CONTAINER_REGISTRY_USER --password-stdin
```
danielezonca, line 52: Do you need this echo command?

.github/workflows/CD.yml
r00ta: https://issues.redhat.com/browse/MGDOBR-31

danielezonca (approved):

PR #7, opened 2021-09-08
[MGBOBR-6] Part 1: Manager-side API for adding Processor to a Bridge

robpblake:
This PR adds an additional endpoint to our API to add support for adding a Processor to a Bridge. The supported endpoint is `/api/v1/bridges/{id}/processors`. A Processor can be created on the given Bridge instance by `POST`-ing a request to this endpoint.

The shard side of the implementation is deliberately excluded from this PR. That will be in a follow-on PR.

Other minor changes in this PR include:

- Ensuring we return a HTTP 201 Created Response code from our API in response to POST requests as per guidelines

### DIFF
--- a/manager/src/main/java/com/redhat/developer/manager/ProcessorService.java
+++ b/manager/src/main/java/com/redhat/developer/manager/ProcessorService.java
@@ -8,11 +8,9 @@ import javax.transaction.Transactional;
 
 import com.redhat.developer.infra.dto.BridgeStatus;
 import com.redhat.developer.manager.api.models.requests.ProcessorRequest;
-import com.redhat.developer.manager.dao.BridgeDAO;
 import com.redhat.developer.manager.dao.ProcessorDAO;
 import com.redhat.developer.manager.exceptions.AlreadyExistingItemException;
 import com.redhat.developer.manager.exceptions.BridgeLifecycleException;
-import com.redhat.developer.manager.exceptions.ItemNotFoundException;
 import com.redhat.developer.manager.models.Bridge;
 import com.redhat.developer.manager.models.Processor;
 
@@ -23,15 +21,11 @@ public class ProcessorService {
     ProcessorDAO processorDAO;
 
     @Inject
-    BridgeDAO bridgeDAO;
+    BridgesService bridgesService;
 
     @Transactional
-    public Processor createProcessor(String customerId, String bridgeId, ProcessorRequest processorRequest) {
-        Bridge bridge = bridgeDAO.findByIdAndCustomerId(bridgeId, customerId);
-        if (bridge == null) {
-            throw new ItemNotFoundException("Bridge with id '" + bridgeId + "' does not exist for customer '" + customerId + "'");
-        }
-
+    public Processor createProcessor(String bridgeId, String customerId, ProcessorRequest processorRequest) {
+        Bridge bridge = bridgesService.getBridge(bridgeId, customerId);
         if (BridgeStatus.AVAILABLE != bridge.getStatus()) {
             /* We cannot deploy Processors to a Bridge that is not Available */
             throw new BridgeLifecycleException("Bridge with id '" + bridge.getId() + "' for customer '" + customerId + "' is not in the '" + BridgeStatus.AVAILABLE + "' state.");
@@ -46,7 +40,7 @@ public class ProcessorService {
         p.setName(processorRequest.getName());
         p.setSubmittedAt(ZonedDateTime.now());
         p.setStatus(BridgeStatus.REQUESTED);
-        bridge.addProcessor(p);
+        p.setBridge(bridge);
         processorDAO.persist(p);
         return p;
     }
--- a/manager/src/main/java/com/redhat/developer/manager/ProcessorService.java
+++ b/manager/src/main/java/com/redhat/developer/manager/ProcessorService.java
@@ -1,0 +1,55 @@
+package com.redhat.developer.manager;
+
+import java.time.ZonedDateTime;
+import java.util.Collections;
+import java.util.List;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import javax.transaction.Transactional;
+
+import com.redhat.developer.infra.dto.BridgeStatus;
+import com.redhat.developer.manager.api.models.requests.ProcessorRequest;
+import com.redhat.developer.manager.dao.BridgeDAO;
+import com.redhat.developer.manager.dao.ProcessorDAO;
+import com.redhat.developer.manager.exceptions.AlreadyExistingItemException;
+import com.redhat.developer.manager.exceptions.BridgeLifecycleException;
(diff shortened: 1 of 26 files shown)


manager/src/main/java/com/redhat/developer/manager/ProcessorService.java
```diff
@@ -0,0 +1,53 @@
...
+        Processor p = processorDAO.findByBridgeIdAndName(bridgeId, processorRequest.getName());
+        if (p != null) {
+            throw new AlreadyExistingItemException("Processor with name '" + processorRequest.getName() + "' already exists for bridge with id '" + bridgeId + "' for customer '" + customerId + "'");
+        }
+
+        p = new Processor();
+        p.setName(processorRequest.getName());
+        p.setSubmittedAt(ZonedDateTime.now());
+        p.setStatus(BridgeStatus.REQUESTED);
+        bridge.addProcessor(p);
```
r00ta, line 49:
```suggestion
```
According to our schema definition, I think you don't need to do this

r00ta (commented):
Hi @robpblake , few minor comments from my side

manager/src/main/java/com/redhat/developer/manager/models/Bridge.java
```diff
@@ -53,9 +58,22 @@
...
+    @OneToMany(fetch = FetchType.LAZY, mappedBy = "bridge", orphanRemoval = true, cascade = CascadeType.ALL)
+    private List<Processor> processors = new ArrayList<>();
+
     public Bridge() {
     }
 
+    public void addProcessor(Processor processor) {
+        this.processors.add(processor);
+        processor.setBridge(this);
+    }
```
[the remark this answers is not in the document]
robpblake: @r00ta It's a generally recognised pattern for managing relationships in `@OneToMany` in JPA. Check out : https://vladmihalcea.com/the-best-way-to-map-a-onetomany-association-with-jpa-and-hibernate/, particulary the section titled `BiDirectional @OneToMany`

manager/src/main/java/com/redhat/developer/manager/api/models/responses/ProcessorResponse.java
```diff
@@ -0,0 +1,98 @@
...
+    private final String kind = "Processor";
+
+    @JsonProperty("href")
+    private String href;
+
+    @JsonProperty("name")
+    private String name;
+
+    @JsonProperty("bridge")
+    private BridgeResponse bridge;
```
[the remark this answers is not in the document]
robpblake: Felt useful to be able to see details of the Bridge a Processor is attached to without having to make another request to the API.

manager/src/main/java/com/redhat/developer/manager/ProcessorService.java
robpblake: This is setting up the foreign key relationship between Bridge --> Processor as part of the data i.e. a Processor cannot be inserted into the DB without being associated with a valid Bridge.

manager/src/test/java/com/redhat/developer/manager/dao/ProcessorDAOTest.java
```diff
@@ -0,0 +1,113 @@
...
+        Processor p = createProcessor(b, "foo");
+
+        Processor byBridgeIdAndName = processorDAO.findByBridgeIdAndName(b.getId(), p.getName());
+        assertThat(byBridgeIdAndName, is(notNullValue()));
+        assertThat(byBridgeIdAndName.getName(), equalTo(p.getName()));
+        assertThat(byBridgeIdAndName.getBridge().getId(), equalTo(b.getId()));
+    }
+
+    @Test
+    @Transactional
```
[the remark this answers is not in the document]
robpblake: I need the `@Transactional` here for the `processorDAO.getEntityManager.merge()` calls as there is no active transaction otherwise.

manager/src/main/java/com/redhat/developer/manager/models/Bridge.java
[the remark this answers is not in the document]
danielezonca: I think the name of the method is a bit misleading: I usually don't expect a `add...(value)` to perform a side effect on the argument `value`.
Maybe rename as `register...`/`link....`/`bind...` could help 🤔

danielezonca (approved):
Just a minor comment

manager/src/main/java/com/redhat/developer/manager/ProcessorService.java
r00ta: You sure @robpblake ? I've just tried and it works

robpblake: You make me doubt myself now. Let me double check.

robpblake: @r00ta Just to double check: you're suggesting I remove the `bridge.addProcessor(p);` line?

When I do that I get (as I would expect):

```
Caused by: org.h2.jdbc.JdbcSQLException: NULL not allowed for column "BRIDGE_ID"; SQL statement:
insert into Processor (bridge_id, name, published_at, status, submitted_at, version, id) values (?, ?, ?, ?, ?, ?, ?) [23502-197]
```

Nothing else is setting up the relationship between the entities other that the `addProcessor(p)` call.

manager/src/test/java/com/redhat/developer/manager/utils/TestUtils.java
```diff
@@ -2,61 +2,58 @@
...
         return given()
                 .filter(new ResponseLoggingFilter())
                 .contentType(ContentType.JSON)
-                .when()
+                .when();
+    }
+
+    public static Response getBridges() {
+        return jsonRequest()
                 .get("/api/v1/bridges");
```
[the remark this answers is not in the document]
robpblake: 👍

manager/src/main/java/com/redhat/developer/manager/ProcessorService.java
r00ta: Let me try again

r00ta: Sorry @robpblake , I mean to replace `bridge.addProcessor(p);` with `p.setBridge(bridge);`. In this very moment `bridge.addProcessor(p);` is setting adding the processor `p` to its list and the bridge `bridge` to the processor `p`. We can remove this if you simply do `p.setBridge(bridge);` here

robpblake:
@r00ta / @danielezonca Updated. I decided to remove the bi-directional association between Bridge and Processor as thinking through the workflows we probably don't need it and it only really works if the number of Processors on a bridge remains low. This resolves the discussions around `addProcessor()/removeProcessor()` methods on the `Bridge`.

Factored in all other feedback from you both.

cheers,

Rob

r00ta (approved):

PR #8, opened 2021-09-08
MGDOBR-22-23: add kustomization project

r00ta:
https://issues.redhat.com/browse/MGDOBR-22
https://issues.redhat.com/browse/MGDOBR-23

This PR aims to add the kustomization project so to deploy the all-in-one artifact together with the kafka cluster and the postgresql database. 

When the modules will be moved to separated applications (ingress, executor, manager, shard), the kustomization project has to be updated accordingly


kustomize/base/all-in-one/resources/deploy.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+  selector:
+    matchLabels:
+      app: event-bridge
+  replicas: 1
+  template:
+    spec:
+      containers:
+        - image: event-bridge-all-in-one
+          imagePullPolicy: Always
```
robpblake, line 14: Can be `IfNotPresent`. We're using a unique image tag for each build of our container.

kustomize/base/all-in-one/resources/deploy.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: apps/v1
```
robpblake, line 1: Looks like we're missing:

- `Service` for the `Deployment`
- `Route` for the API on the Manager
- `Route` for the Ingress

kustomize/overlays/prod/all-in-one/deploy-config.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: ConfigMap
+metadata:
+  name: event-bridge-all-in-one-config
+data:
+  EVENT_BRIDGE_DB_USERNAME: "postgres"
+  EVENT_BRIDGE_DB_PASSWORD: "postgres"
+  EVENT_BRIDGE_DB_HOST: "event-bridge-db"
+  EVENT_BRIDGE_DB_PORT: "5432"
+  EVENT_BRIDGE_DB_SCHEMA: "event-bridge"
+  EVENT_BRIDGE_MANAGER_URL: "http://event-bridge:8080"
```
robpblake, line 11: I've created https://issues.redhat.com/browse/MGDOBR-32 as we should switch this to use a TLS secured connection as soon as possible. I'm OK with the `http` endpoint for this PR 👍

r00ta: @robpblake the service will always expose http endpoints, the api gateway is responsible for the secure connection with the clients outside the cluster. 
In other words, if the shard and the manager are on the same cluster, they will talk with http connections. If they are on different clusters, they will use https connections provided by the api gateway. 

MGDOBR-32 is only a consequence of exposing secure routes with OCP at the end of the day imo

kustomize/base/all-in-one/resources/deploy.yaml
r00ta: I've included the `Route` in the `prod` overlay. In that way we keep the `base` not OCP specific: we can create a `dev` or `minikube` overlay that deploys the stuff on a minikube cluster. Let me know if it's ok for you @robpblake

robpblake: Sounds good 👍

kustomize/overlays/prod/all-in-one/deploy-config.yaml
robpblake: I take your point about the gateway handling TLS for connection coming from outside the cluster, but any in-cluster HTTP communications will also need to be secured with TLS. 

We'll be covered by the data encryption standards as part of ESS (https://source.redhat.com/departments/it/it-information-security/wiki/enterprise_security_standard_80_essv8#SECDATAREQ2_Encryption_of_DatainTransit) so this will include things like:

- TLS on the connection to our database
- TLS to any HTTP endpoint in the same cluster 
- TLS to any HTTP endpoint external to the cluster

It's fine for this iteration until we start to break out the shard and manager, but we just need to keep it in mind as we go that we should default to TLS whenever possible.

r00ta: I honestly do not see the point of making https connections between services running on the same cluster. If we worry about the traffic **within the same cluster**, it means we want to protect the traffic by the people that have access to the cluster. And actually if somebody has access to the cluster he/she can do much more dangerous things (read secrets, redirect traffic and whatever). And if the cluster is shared, the cluster policies should take care of isolation between the projects/services and not the services themselves.

I did not go through the big document, but when it will be the moment I'll have a look

r00ta:
Merged as approved in gchat

PR #16, opened 2021-09-13
MGDOBR-34: Add integration tests

r00ta:
https://issues.redhat.com/browse/MGDOBR-34

This PR aims to add the Bridge integration tests.


build-parent/pom.xml
```diff
@@ -37,8 +37,10 @@
     <impsort.plugin.version>1.5.0</impsort.plugin.version>
 
     <version.io.cloudevents>2.2.0</version.io.cloudevents>
-    <version.org.testcontainer>1.16.0</version.org.testcontainer>
+    <version.org.testcontainer>1.15.3</version.org.testcontainer>
```
r00ta, line 40: quarkus bug https://github.com/quarkusio/quarkus/issues/16541#issuecomment-918004517

runner/src/test/java/com/redhat/developer/runner/it/End2EndTestIT.java
```diff
@@ -0,0 +1,149 @@
...
+
+import io.quarkus.test.common.QuarkusTestResource;
+import io.quarkus.test.junit.QuarkusTest;
+import io.restassured.filter.log.ResponseLoggingFilter;
+import io.restassured.http.ContentType;
+import io.restassured.specification.RequestSpecification;
+
+import static io.restassured.RestAssured.given;
+
+@QuarkusTest
```
danielezonca, line 32: Can you use [`@QuarkusIntegrationTest`](https://quarkus.io/guides/getting-started-testing#quarkus-integration-test) annotation? It should be the best fit for integration (and it should work for native too if necessary).

runner/pom.xml
```diff
@@ -32,6 +32,38 @@
...
+
+
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-junit5</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-jdbc-h2</artifactId>
```
danielezonca, line 44: As proper E2E test it should use Postgres and not h2

runner/src/test/java/com/redhat/developer/runner/it/End2EndTestIT.java
r00ta: @danielezonca I've tried that. But it's like the annotation builds the artifact and runs it with `java -jar ...`: I did not find a way to inject the configurations.

tbh I did not spend much time on that, so it's probably possible, but I'd suggest to stay with @QuarkusTest that works pretty well at least until we are using the All-In-One application.  https://quarkus.io/guides/getting-started-testing#quarkus-integration-test

runner/src/test/java/com/redhat/service/bridge/runner/it/End2EndTestIT.java
```diff
@@ -0,0 +1,153 @@
...
+
+    private static String bridgeId;
+
+    @ConfigProperty(name = "event-bridge.manager.url")
+    String managerUrl;
+
+    @Order(1)
+    @Test
+    public void getEmptyBridges() {
+        System.out.println(managerUrl);
```
danielezonca, line 47: Leftover?

runner/src/test/java/com/redhat/service/bridge/runner/it/PostgresResource.java
```diff
@@ -0,0 +1,44 @@
...
+
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
+
+public class PostgresResource implements QuarkusTestResourceLifecycleManager {
+
+    private final PostgreSQLContainer postgres = new PostgreSQLContainer<>("postgres:10.5")
+            .withDatabaseName("hibernate_orm_test")
+            .withUsername("hibernate_orm_test")
+            .withPassword("hibernate_orm_test")
+            .withExposedPorts(5432)
```
danielezonca, line 20: Is it possible to move to test application properties file? You can use [config programmatic API](https://quarkus.io/guides/config#programmatically-access-the-configuration) if necessary to retrieve them

runner/src/test/java/com/redhat/service/bridge/runner/it/PostgresResource.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override
+    public Map<String, String> start() {
+        postgres.start();
+
+        Map<String, String> config = new HashMap<>();
+        config.put("EVENT_BRIDGE_DB_USERNAME", postgres.getUsername());
+        config.put("EVENT_BRIDGE_DB_PASSWORD", postgres.getPassword());
+        config.put("EVENT_BRIDGE_DB_HOST", postgres.getHost());
+        config.put("EVENT_BRIDGE_DB_PORT", "5432");
```
danielezonca, line 35: Is it possible to get port from `postgres` instance like other params?

runner/src/test/java/com/redhat/service/bridge/runner/it/PostgresResource.java
r00ta: As per chat I did a quick investigation and unfortunately it does not work. I guess it's because of @QuarkusTestResource.

runner/src/test/java/com/redhat/developer/runner/it/End2EndTestIT.java
danielezonca: Ok fine for me 👍 
We can revisit in the future while looking at native compilation

danielezonca (approved):

PR #19, opened 2021-09-13
[MGDOBR-6] Attempt to fix flaky test and improve Exception message for clarity

robpblake:


r00ta (approved):

manager/src/main/java/com/redhat/service/bridge/manager/api/internal/ShardBridgesSyncAPI.java
```diff
@@ -34,7 +34,7 @@
 public class ShardBridgesSyncAPI {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ShardBridgesSyncAPI.class);
-    private static final List<BridgeStatus> statuses = Arrays.asList(BridgeStatus.REQUESTED, BridgeStatus.PROVISIONING, BridgeStatus.DELETION_REQUESTED);
```
danielezonca, line 37: Isn't this change just hiding in all the endpoints bridges in `PROVISIONING` status? Why this change? Can't this happen in real world?

robpblake: Yep it can happen in the real-world, but not really in the current implementation given that everything is tied into a single deployment and we just start from fresh whenever we deploy the components. 

As/when we start to split things out, we support restarting of components and also look into correct failing handling, then we'll need to revisit what is returned from this API.

danielezonca: Thanks for the clarification, do we have a ticket for that? So that we keep this in the backlog

danielezonca (approved):

PR #21, opened 2021-09-13
[MGDOBR-6] Updated communication between Shard and Manager for Processor Deployment

robpblake:
This PR updates the internal `/api/v1/shard/bridges/processors` endpoint to return all Processors that need to be deployed. This removes the need for the Shard to make a request-per-Bridge to determine what Processors to deploy. The Shard is updated to make use of this endpoint.

In addition, the general flow of handling the data from the Manager has been aligned further with how we do it for bridges, which should hopefully simplify things as/when we start to migrate to using K8S crds for state management.

Finally I've added `TODO` statements into the `ProcessorController`, just to reiterate that the in-memory solution is temporary for iteration 1.


shard/src/main/java/com/redhat/service/bridge/shard/ManagerSyncServiceImpl.java
```diff
@@ -40,11 +39,15 @@
...
-        LOGGER.info("[Shard] wakes up to get Bridges to deploy and delete");
+    void syncUpdatesFromManager() {
+        LOGGER.info("[Shard] Fetching updates from Manager for Bridges and Processors to deploy and delete");
         fetchAndProcessBridgesToDeployOrDelete().subscribe().with(
-                success -> LOGGER.info("[shard] has processed all the Bridges to deploy or delete"),
-                failure -> LOGGER.warn("[shard] something went wrong during the process of the Bridges to be deployed or deleted"));
+                success -> processingComplete(BridgeDTO.class),
+                failure -> processingFailed(BridgeDTO.class, failure));
+
+        fetchAndProcessorProcessorsToDeployOrDelete().subscribe().with(
```
r00ta, line 48: typo

r00ta (approved):

danielezonca (approved):

shard/src/main/java/com/redhat/service/bridge/shard/ManagerSyncServiceImpl.java
robpblake: Thanks!

PR #25, opened 2021-09-15
MGDOBR-37: refactoring with k8s facade

r00ta:
This PR aims to refactor between the modules that will be provisioned by k8s. 

- a new module `k8s` has been added. It implements the `KubernetesClient` interface defined in the `infra` module. The idea is that the shard will inject the `KubernetesClient` and the implementation bean will be provided by the `k8s` module (imported by the runner at the moment). 
- The `KubernetesClient` interface tries to be similar to the actual fabric8 kubernetes api (it is possible to create/delete/update custom resources, deployments and ingresses). They are not 100% similar because of course the purpose is not to implement another kubernetes-like sdk. 
- The `shard` will create the `ProcessorCustomResource` and the `BridgeCustomResource` from the `BridgeDTO`, then the `ProcessorController` and the `IngressController` will deploy the applications/ingresses depending on what they need. 
- The `ProcessorCustomResource` and the `BridgeCustomResource` CRDs are currently in the `infra` module. This is because the `k8s` needs to access them. This is going to be refactored in another pull request -> they will be moved to the `shard` module. 
- The reconcile loops of the `IngressController` and the `ProcessorController` can be summarized as: 
    - If the resource has been deleted, then undeploy all the dependent resources and notify the manager.
    - If a resource (a CR, deployment, ingress) has been modified, then the reconcile loop checks that all the resources are deployed. If any of the resource is missing, it deploys what it needs.
- The interactions between the `k8s` module and the `ingress`/`executor` modules will be refactored in another pull request


k8s/src/main/java/com/developer/service/bridge/k8s/DeploymentsManagerImpl.java
```diff
@@ -0,0 +1,82 @@
...
+        DeploymentStatus status = new DeploymentStatusBuilder()
+                .withConditions(new DeploymentConditionBuilder().withStatus("Ready").build())
+                .build();
+        deployment.setStatus(status);
+
+        String type = KubernetesUtils.extractTypeFromMetadata(deployment);
+        if (type.equals(K8SBridgeConstants.PROCESSOR_TYPE)) {
+            // hack for the time being
+            ProcessorCustomResource processorCustomResource = customResourceManager.getCustomResource(name, ProcessorCustomResource.class);
+            executorsService.createExecutor(processorCustomResource.toDTO());
```
r00ta, line 60: `executorsService.createExecutor` wants the `ProcessorDTO`, this is an hack until the next refactoring

shard/src/main/java/com/redhat/service/bridge/shard/OperatorServiceInMemoryImpl.java
```diff
@@ -1,87 +1,52 @@
...
-import com.redhat.service.bridge.shard.controllers.ProcessorController;
-
-import io.quarkus.scheduler.Scheduled;
+import com.redhat.service.bridge.infra.k8s.K8SBridgeConstants;
+import com.redhat.service.bridge.infra.k8s.KubernetesClient;
+import com.redhat.service.bridge.infra.k8s.crds.BridgeCustomResource;
+import com.redhat.service.bridge.infra.k8s.crds.ProcessorCustomResource;
 
 @ApplicationScoped
 public class OperatorServiceInMemoryImpl implements OperatorService {
```
danielezonca, line 17: Do we really need to preserve this class? It looks now like a wrapper of `KubernetesClient` so we can probably remove it

danielezonca (approved):
Very well done, only few minor comments 👍

infra/src/main/java/com/redhat/service/bridge/infra/k8s/KubernetesClient.java
```diff
@@ -0,0 +1,24 @@
...
+import io.fabric8.kubernetes.api.model.networking.v1.Ingress;
+
+public interface KubernetesClient {
+    void createOrUpdateDeployment(Deployment deployment);
+
+    void deleteDeployment(String name);
+
+    Deployment getDeployment(String name);
+
+    void createOrUpdateCustomResource(String name, Object customResource, String type);
```
robpblake, line 13: If `Object` here is instead `BridgeCustomResource` and that has a `getType()` method on it, does that simplify things? `BridgeCustomResource` and `ProcessorCustomResource` would then extend `BridgeCustomResource`.

I see we have method calls where we do things like `createOrUpdateCustomResource("foo, BridgeCustomResource.fromDTO(bridgeDTO), K8SBridgeConstants.BRIDGE_TYPE);`

I can easily do something like `createOrUpdateCustomResource("foo, BridgeCustomResource.fromDTO(bridgeDTO), K8SBridgeConstants.PROCESSOR_TYPE);`, whereas if we have an `BridgeCustomResource`, I cannot do that.

infra/src/main/java/com/redhat/service/bridge/infra/k8s/KubernetesClient.java
```diff
@@ -0,0 +1,24 @@
...
+
+    <T> T getCustomResource(String name, Class<T> tClass);
```
robpblake, line 15: If the above comment makes sense, this would become `<T extends BridgeCustomResource>`

infra/src/main/java/com/redhat/service/bridge/infra/k8s/KubernetesClient.java
```diff
@@ -0,0 +1,24 @@
+package com.redhat.service.bridge.infra.k8s;
+
+import io.fabric8.kubernetes.api.model.apps.Deployment;
+import io.fabric8.kubernetes.api.model.networking.v1.Ingress;
+
+public interface KubernetesClient {
```
robpblake, line 6: I'm not sure about calling this `KubernetesClient`. I can just see us ending up in conversations where we say things like "our KubernetesClient, not the Fabric8 KubernetesClient" or "this is our KubernetesClient which wraps another KubernetesClient" 

I understand the intent, but I wonder if there is a better name for this class that stops us having slightly awkward conversations in the future?

infra/src/main/java/com/redhat/service/bridge/infra/k8s/K8SBridgeConstants.java
```diff
@@ -0,0 +1,7 @@
+package com.redhat.service.bridge.infra.k8s;
+
+public class K8SBridgeConstants {
+    public static final String METADATA_TYPE = "type";
+    public static final String PROCESSOR_TYPE = "processor";
```
robpblake, line 5: Any reason why these can't be on the `ProcessorCustomResource` and `BridgeCustomResource`?

k8s/src/main/java/com/developer/service/bridge/k8s/CustomResourceManager.java
```diff
@@ -0,0 +1,9 @@
+package com.developer.service.bridge.k8s;
+
+public interface CustomResourceManager {
+    void createOrUpdateCustomResource(String name, Object customResource, String type);
```
robpblake, line 4: Same comment here about defining a `BridgeCustomResource` type

infra/src/main/java/com/redhat/service/bridge/infra/k8s/ResourceEvent.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class ResourceEvent {
+
+    private String subject;
+
+    private String resourceId;
+
+    private Action action;
+
+    public ResourceEvent(String subject, String resourceId, Action action) {
```
robpblake, line 11: Just to make sure I understand, `subject` here is the `type` e.g `processor` or `bridge`? If yes, can we rename to `type` for consistency.

infra/src/main/java/com/redhat/service/bridge/infra/k8s/ResourceEvent.java
```diff
@@ -0,0 +1,40 @@
+package com.redhat.service.bridge.infra.k8s;
+
+public class ResourceEvent {
```
robpblake, line 3: Any reason not to just pass in the entire resource, particularly if we have the `BridgeCustomResource` interface with `getType` and `getName` methods on it?

robpblake (commented):
@r00ta As always, the quality of the work here is great 👍 

Some minor implementation feedback from me.

A general comment: I do have some reservations about this. It seems like we're putting our own custom layer on top of functionality that I would expect to be mostly delivered by the Java operator-sdk.  I'm not sure I fully see (yet) the benefit of the extra layer of indirection  our own Events brings us over and above what the operator-sdk will provide. The shard will be an operator that works with K8S API, I don't see the need to try and abstract that too much. As we move more into our K8S implementation, I'd be keen to see what this brings us over simply `@Inject` directly our implementations into controllers that are invoked by events from the Kube API. When I see things like `// equivalent of ResourceEventSource for operator sdk` in the code, that makes me wonder...

I'm happy to add my +1 , but let's keep an eye on this, particularly as we move to using the operator-sdk for real. I just want to make sure this abstraction continues to provide benefit and doesn't add another layer for us to understand and maintain.

r00ta:
> @r00ta As always, the quality of the work here is great 
> 
> Some minor implementation feedback from me.
> 
> A general comment: I do have some reservations about this. It seems like we're putting our own custom layer on top of functionality that I would expect to be mostly delivered by the Java operator-sdk. I'm not sure I fully see (yet) the benefit of the extra layer of indirection our own Events brings us over and above what the operator-sdk will provide. The shard will be an operator that works with K8S API, I don't see the need to try and abstract that too much. As we move more into our K8S implementation, I'd be keen to see what this brings us over simply `@Inject` directly our implementations into controllers that are invoked by events from the Kube API. When I see things like `// equivalent of ResourceEventSource for operator sdk` in the code, that makes me wonder...
> 
> I'm happy to add my +1 , but let's keep an eye on this, particularly as we move to using the operator-sdk for real. I just want to make sure this abstraction continues to provide benefit and doesn't add another layer for us to understand and maintain.

@robpblake the k8s module is intended to be 100% replaced by the operator-sdk. The purpose of this refactoring is only to move to such type of interactions (the CRD is deployed -> the reconcile loop is triggered -> the reconcile loop checks and deploys what is expected to be the desired state). 

In other words, for example, the call to `kubernetesClient.createOrUpdateCustomResource` will be replaced by `fabric8` `kubernetesClient.customResource(..).createOrReplace(namespace, name)` (like this call here https://github.com/r00ta/baaas-decision-fleet-shard/blob/c173edb2fdd3d02abbb9e2cfc1fa23f71b99b1d6/decision-fleet-shard-operator/src/main/java/org/kie/baaas/dfs/service/KogitoService.java#L195)

robpblake:
> > @r00ta As always, the quality of the work here is great
> > Some minor implementation feedback from me.
> > A general comment: I do have some reservations about this. It seems like we're putting our own custom layer on top of functionality that I would expect to be mostly delivered by the Java operator-sdk. I'm not sure I fully see (yet) the benefit of the extra layer of indirection our own Events brings us over and above what the operator-sdk will provide. The shard will be an operator that works with K8S API, I don't see the need to try and abstract that too much. As we move more into our K8S implementation, I'd be keen to see what this brings us over simply `@Inject` directly our implementations into controllers that are invoked by events from the Kube API. When I see things like `// equivalent of ResourceEventSource for operator sdk` in the code, that makes me wonder...
> > I'm happy to add my +1 , but let's keep an eye on this, particularly as we move to using the operator-sdk for real. I just want to make sure this abstraction continues to provide benefit and doesn't add another layer for us to understand and maintain.
> 
> @robpblake the k8s module is intended to be 100% replaced by the operator-sdk. The purpose of this refactoring is only to move to such type of interactions (the CRD is deployed -> the reconcile loop is triggered -> the reconcile loop checks and deploys what is expected to be the desired state).
> 
> In other words, for example, the call to `kubernetesClient.createOrUpdateCustomResource` will be replaced by `fabric8` `kubernetesClient.customResource(..).createOrReplace(namespace, name)` (like this call here https://github.com/r00ta/baaas-decision-fleet-shard/blob/c173edb2fdd3d02abbb9e2cfc1fa23f71b99b1d6/decision-fleet-shard-operator/src/main/java/org/kie/baaas/dfs/service/KogitoService.java#L195)



> > @r00ta As always, the quality of the work here is great
> > Some minor implementation feedback from me.
> > A general comment: I do have some reservations about this. It seems like we're putting our own custom layer on top of functionality that I would expect to be mostly delivered by the Java operator-sdk. I'm not sure I fully see (yet) the benefit of the extra layer of indirection our own Events brings us over and above what the operator-sdk will provide. The shard will be an operator that works with K8S API, I don't see the need to try and abstract that too much. As we move more into our K8S implementation, I'd be keen to see what this brings us over simply `@Inject` directly our implementations into controllers that are invoked by events from the Kube API. When I see things like `// equivalent of ResourceEventSource for operator sdk` in the code, that makes me wonder...
> > I'm happy to add my +1 , but let's keep an eye on this, particularly as we move to using the operator-sdk for real. I just want to make sure this abstraction continues to provide benefit and doesn't add another layer for us to understand and maintain.
> 
> @robpblake the k8s module is intended to be 100% replaced by the operator-sdk. The purpose of this refactoring is only to move to such type of interactions (the CRD is deployed -> the reconcile loop is triggered -> the reconcile loop checks and deploys what is expected to be the desired state).
> 
> In other words, for example, the call to `kubernetesClient.createOrUpdateCustomResource` will be replaced by `fabric8` `kubernetesClient.customResource(..).createOrReplace(namespace, name)` (like this call here https://github.com/r00ta/baaas-decision-fleet-shard/blob/c173edb2fdd3d02abbb9e2cfc1fa23f71b99b1d6/decision-fleet-shard-operator/src/main/java/org/kie/baaas/dfs/service/KogitoService.java#L195)

OK thanks for clearing that up. 

In that case, I'm not sure how much of my feedback on the `k8s` module is valid here; I wouldn't want you to spend too much time on it if we're planning to replace it shortly with operator-sdk. Given the current speed of implementation, I don't think it will be that long before that work can start.

Can you take a call on the feedback I've provided and decide what is/isn't valid given ^.

cheers,

Rob

robpblake (approved):

r00ta:
> > > @r00ta As always, the quality of the work here is great
> > > Some minor implementation feedback from me.
> > > A general comment: I do have some reservations about this. It seems like we're putting our own custom layer on top of functionality that I would expect to be mostly delivered by the Java operator-sdk. I'm not sure I fully see (yet) the benefit of the extra layer of indirection our own Events brings us over and above what the operator-sdk will provide. The shard will be an operator that works with K8S API, I don't see the need to try and abstract that too much. As we move more into our K8S implementation, I'd be keen to see what this brings us over simply `@Inject` directly our implementations into controllers that are invoked by events from the Kube API. When I see things like `// equivalent of ResourceEventSource for operator sdk` in the code, that makes me wonder...
> > > I'm happy to add my +1 , but let's keep an eye on this, particularly as we move to using the operator-sdk for real. I just want to make sure this abstraction continues to provide benefit and doesn't add another layer for us to understand and maintain.
> > 
> > 
> > @robpblake the k8s module is intended to be 100% replaced by the operator-sdk. The purpose of this refactoring is only to move to such type of interactions (the CRD is deployed -> the reconcile loop is triggered -> the reconcile loop checks and deploys what is expected to be the desired state).
> > In other words, for example, the call to `kubernetesClient.createOrUpdateCustomResource` will be replaced by `fabric8` `kubernetesClient.customResource(..).createOrReplace(namespace, name)` (like this call here https://github.com/r00ta/baaas-decision-fleet-shard/blob/c173edb2fdd3d02abbb9e2cfc1fa23f71b99b1d6/decision-fleet-shard-operator/src/main/java/org/kie/baaas/dfs/service/KogitoService.java#L195)
> 
> > > @r00ta As always, the quality of the work here is great
> > > Some minor implementation feedback from me.
> > > A general comment: I do have some reservations about this. It seems like we're putting our own custom layer on top of functionality that I would expect to be mostly delivered by the Java operator-sdk. I'm not sure I fully see (yet) the benefit of the extra layer of indirection our own Events brings us over and above what the operator-sdk will provide. The shard will be an operator that works with K8S API, I don't see the need to try and abstract that too much. As we move more into our K8S implementation, I'd be keen to see what this brings us over simply `@Inject` directly our implementations into controllers that are invoked by events from the Kube API. When I see things like `// equivalent of ResourceEventSource for operator sdk` in the code, that makes me wonder...
> > > I'm happy to add my +1 , but let's keep an eye on this, particularly as we move to using the operator-sdk for real. I just want to make sure this abstraction continues to provide benefit and doesn't add another layer for us to understand and maintain.
> > 
> > 
> > @robpblake the k8s module is intended to be 100% replaced by the operator-sdk. The purpose of this refactoring is only to move to such type of interactions (the CRD is deployed -> the reconcile loop is triggered -> the reconcile loop checks and deploys what is expected to be the desired state).
> > In other words, for example, the call to `kubernetesClient.createOrUpdateCustomResource` will be replaced by `fabric8` `kubernetesClient.customResource(..).createOrReplace(namespace, name)` (like this call here https://github.com/r00ta/baaas-decision-fleet-shard/blob/c173edb2fdd3d02abbb9e2cfc1fa23f71b99b1d6/decision-fleet-shard-operator/src/main/java/org/kie/baaas/dfs/service/KogitoService.java#L195)
> 
> OK thanks for clearing that up.
> 
> In that case, I'm not sure how much of my feedback on the `k8s` module is valid here; I wouldn't want you to spend too much time on it if we're planning to replace it shortly with operator-sdk. Given the current speed of implementation, I don't think it will be that long before that work can start.
> 
> Can you make a call on the feedback I've provided and decide what is/isn't valid given ^.
> 
> cheers,
> 
> Rob

If it's fine for you, I'd move forward without refactoring the k8s module. 
Your comments were absolutely valid, but as you said it will be thrown away so I would not spend more time on that. I've resolved the comments without fixing them, I'll move forward :+1:

PR #29, opened 2021-09-17
MGDOBR-10 part 1 - Add filters

r00ta:
This part aims to add the FIlters to the REST API and the internal modules. 

1) The user can create filters by calling POST `/api/v1/bridges/{id}/processors` with a body like
```json
{
 "name" : "myProcessor",
 "filters": [
   { "key": "myJsonKey", "value": "jac", "type": "StringContains"}
 ]
}
```
2) ATM there are 3 Filters implemented: `StringContains`, `StringEquals`, `StringStartsWith` .
3) The endpoint `/api/v1/bridges/{id}/processors` accepts a LIST of Filters. If there are more than one Filters in the requests, those filters are ANDED in the evaluation. 
4) More filters will be implemented (see here https://github.com/r00ta/openbridge-PoC/tree/main/src/main/java/com/redhat/developer/models/filters for a more complete spoiler)
5) The general representation of the Filter is kept in the `infra` module. Only the manager has another representation of the Filter for the entity. (The `ProcessorRequest` uses the Filter of the Infra module, as well as the `ProcessorResponse`, `ProcessorDTO` and `ProcessorCustomResource`. I expect the modules do not have to duplicate all the Filters, the executor will add the Filter implementation logic with a factory.)
6) **IMPORTANT** The topic around polymorphism and JPA is pretty sensible. We could keep one single table with all the attributes in it (with external tables for Filters that have a value instance of List), or we could keep a table for each single Filter. 
I personally went for a custom single-table solution that should be much more efficient and simple: we convert the `value` of the Filter to string, we store such information in the DB, and when we need the Filter we simply build the Filter back using a Factory. 
In this way, we do not have to duplicate all the classes and do not overkill the DB. ** 

What is not included: 
1) The Filters are ignored by the executor at the moment


infra/src/main/java/com/redhat/service/bridge/infra/models/filters/FilterFactory.java
```diff
@@ -0,0 +1,16 @@
...
+            case StringBeginsWith.FILTER_TYPE_NAME:
+                return new StringBeginsWith(key, value);
+            case StringContains.FILTER_TYPE_NAME:
+                return new StringContains(key, value);
+            case StringEquals.FILTER_TYPE_NAME:
+                return new StringEquals(key, value);
+            default:
+                throw new IllegalArgumentException("Type " + type + " is not valid");
+        }
+    }
```
r00ta, line 15: The manager stores the `value` of a Filter as String in the database, then it needs to convert back to a Filter using this factory

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorService.java
```diff
@@ -41,16 +42,22 @@ public Processor getProcessor(String processorId, String bridgeId, String custom
...
+        if (processorDAO.findByBridgeIdAndName(bridgeId, processorRequest.getName()) != null) {
             throw new AlreadyExistingItemException("Processor with name '" + processorRequest.getName() + "' already exists for bridge with id '" + bridgeId + "' for customer '" + customerId + "'");
         }
 
-        p = new Processor();
+        final Processor p = new Processor();
+
+        Set<Filter> filters = null;
+        if (processorRequest.getFilters() != null) {
+            filters = processorRequest.getFilters().stream().map(x -> new Filter(x.getKey(), x.getType(), x.getStringValue(), p)).collect(Collectors.toSet());
```
r00ta, line 53: Here we convert the `ProcessorRequest` to the `Process` entity, and we store the `value` as string

manager/src/main/java/com/redhat/service/bridge/manager/api/user/ProcessorsAPI.java
```diff
@@ -37,6 +38,7 @@
 
 @Path(APIConstants.USER_API_BASE_PATH)
 @Produces(MediaType.APPLICATION_JSON)
+@Transactional
 public class ProcessorsAPI {
```
r00ta, line 42: `@Transactional` moved to the entry point of the application as Quarkus guide recommends. In this way, when the ProcessorResponse is built by Jackson it can `LAZY` fetch the filters

r00ta (commented):
See point `6` of the PR description

manager/src/main/java/com/redhat/service/bridge/manager/api/user/ProcessorsAPI.java
robpblake: Problem will be that with `list` methods, this leads to a n+1 select on the DB e.g:

- select the list of Processors
- foreach Processor, select the list of filters

Instead I'd suggest we do a `left join fetch`  in the queries that service these methods when loading our data from the DB. So the query would be something like:

`from Processor p join fetch p.bridge left join fetch p.filters where p.bridge.id=:bridgeId and p.bridge.customerId=:customerId`

This will mean that the Filters collection is eagerly initialised within the single DB query. By using the `left join fetch` in the queries where we need it, we can:

- Avoid the N+1 selects
- Keep the `@Transaction` boundary on the `ProcessorService`
- Keep the Filter collection marked as `fetch = FetchType.LAZY` on the Entity, so that we don't load it by default, unless you use a NamedQuery or call `getFilters()` within a Transaction.

manager/src/main/resources/db/migration/0.1/V0.1.3__create_filter.sql
```diff
@@ -0,0 +1,10 @@
+create table FILTER
+(
+    id                 varchar(255) NOT NULL PRIMARY KEY,
+    processor_id       varchar(255) NOT NULL,
+    key                varchar(255) NOT NULL,
+    type               varchar(255) NOT NULL,
+    value              TEXT NOT NULL,
+    version            integer NOT NULL default 0,
```
robpblake, line 8: I don't see the version column mapped in the `Filter` entity. Not 100% sure we need it on Filters. It's there to support Optimistic locking in JPA (see https://vladmihalcea.com/jpa-entity-version-property-hibernate/), so it's gonna depend a little bit on our strategy of updating Processors in the future, but I'd expect the `version` on the Processor to be enough.

infra/src/main/java/com/redhat/service/bridge/infra/models/filters/FilterFactory.java
robpblake: Just for visibility, JPA supports `@PostLoad` and `@PostUpdate` callbacks that might be another option for this: https://nullbeans.com/configuring-postload-and-postupdate-in-jpa-hibernate/

I think this is fine for now though as it's clear as to what's going on, but we might want to consider if it makes our life easier in the future.

manager/src/main/java/com/redhat/service/bridge/manager/models/Processor.java
```diff
@@ -168,4 +187,15 @@ public boolean equals(Object o) {
     public int hashCode() {
         return Objects.hash(id);
     }
+
+    private Set<com.redhat.service.bridge.infra.models.filters.Filter> buildFilters() {
```
danielezonca, line 191: I would like to avoid class name clash, can you please rename one of the two `Filter` class?

manager/src/main/resources/db/migration/0.1/V0.1.3__create_filter.sql
r00ta: Removed

manager/src/main/java/com/redhat/service/bridge/manager/api/internal/ShardBridgesSyncAPI.java
```diff
@@ -29,6 +30,7 @@
 @Path(APIConstants.SHARD_API_BASE_PATH)
 @Produces(MediaType.APPLICATION_JSON)
 @Consumes(MediaType.APPLICATION_JSON)
+@Transactional
```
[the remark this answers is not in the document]
r00ta: The idea was to move `@Transactional` to the entry points as quarkus suggests https://quarkus.io/guides/transaction . Anyway, since I've moved to `join fetch` I've reverted the changes

manager/src/main/java/com/redhat/service/bridge/manager/api/user/ProcessorsAPI.java
r00ta: @robpblake done, now the queries look like
```java
@NamedQueries({
        @NamedQuery(name = "PROCESSOR.findByBridgeIdAndName",
                query = "from Processor p left join fetch p.filters where p.name=:name and p.bridge.id=:bridgeId"),
        @NamedQuery(name = "PROCESSOR.findByStatus",
                query = "from Processor p join fetch p.bridge left join fetch p.filters where p.status in (:statuses) and p.bridge.status='AVAILABLE'"),
        @NamedQuery(name = "PROCESSOR.findByIdBridgeIdAndCustomerId",
                query = "from Processor p join fetch p.bridge left join fetch p.filters where p.id=:id and (p.bridge.id=:bridgeId and p.bridge.customerId=:customerId)"),
        @NamedQuery(name = "PROCESSOR.findByBridgeIdAndCustomerId",
                query = "from Processor p join fetch p.bridge left join fetch p.filters where p.bridge.id=:bridgeId and p.bridge.customerId=:customerId"),
        @NamedQuery(name = "PROCESSOR.countByBridgeIdAndCustomerId",
                query = "select count(p.id) from Processor p where p.bridge.id=:bridgeId and p.bridge.customerId=:customerId"),
        @NamedQuery(name = "PROCESSOR.idsByBridgeIdAndCustomerId",
                query = "select p.id from Processor p where p.bridge.id=:bridgeId and p.bridge.customerId=:customerId order by p.submittedAt asc"),
        @NamedQuery(name = "PROCESSOR.findByIds",
                query = "select p from Processor p join fetch p.bridge left join fetch p.filters where p.id in (:ids)")
})
```

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
```diff
@@ -67,6 +72,11 @@ public Bridge getBridge(String id, String customerId) {
 
     @Override
     public void deleteBridge(String id, String customerId) {
+        List<Processor> processors = processorService.getProcessors(id, customerId);
```
danielezonca, line 75: Same comment about a `count()` method or a list of Ids instead of the whole entity

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
```diff
@@ -67,6 +72,11 @@ public Bridge getBridge(String id, String customerId) {
...
+        if (processors.size() > 0) {
+            throw new BridgeLifecycleException("It is not possible to delete a Bridge instance with active Processors.");
```
danielezonca, line 77: I think this is true only until we implement a delete cascade/delete all logic right? If so, do we have a ticket to mention here in a comment?

manager/src/main/java/com/redhat/service/bridge/manager/models/Processor.java
danielezonca, line 191: I would move this logic to `ProcessorDAO` to keep the entity ~without logic. What do you think?

infra/src/main/java/com/redhat/service/bridge/infra/models/filters/FilterFactory.java
r00ta: Created https://issues.redhat.com/browse/MGDOBR-42

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
r00ta: https://issues.redhat.com/browse/MGDOBR-43

manager/src/main/java/com/redhat/service/bridge/manager/models/Processor.java
r00ta: @danielezonca wdym? This `ProcessorEntity` contains only `FilterEntity`s. 
We use this `buildFilters` method only in the `toDTO` and `toResponse` methods so to convert the `FilterEntity` to `Filter`.  If `ProcessorDAO` should return `ProcessEntity` instances which contain `FilterEntity`, not `Filter.

manager/src/test/java/com/redhat/service/bridge/manager/utils/DatabaseManagerUtils.java
```diff
@@ -23,12 +27,31 @@
...
-        bridgeDAO.deleteAll();
+        deleteAllProcessors();
+        deleteAllBridges();
+    }
+
+    /**
+     * processorDAO.deleteAll() does not cascade! https://github.com/quarkusio/quarkus/issues/13941
+     */
+    private void deleteAllProcessors() {
+        List<Processor> processors = processorDAO.listAll();
```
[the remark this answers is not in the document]
r00ta: I don't think we will experience such perfomances issues in this test utility, but I've implemented your suggestions

manager/src/main/java/com/redhat/service/bridge/manager/models/Processor.java
r00ta: Renamed the infra one to  `BaseFilter`

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
r00ta: Implemented with `getProcessorsCount` in the `ProcessorService` and `ProcessorDAO` accordingly

robpblake (approved):

danielezonca (approved):

PR #31, opened 2021-09-20
MGDOBR-41: refactor executor with new k8s module and deployment logic

r00ta:
This PR aims to refactor the executor logic so to separate the configuration provisioning from the core logic of the executor. 

The `ExecutorService` will get the `Executors` from an `ExecutionConfigProvider`: 
1) The vision is that the implementation `ExecutionConfigProviderImpl` will read the configuration from the `application.properties`/a file in the image/from a remote url and will create the `Executors`. 
2) Until we do not move away from the all-in-one application (i.e. switch to`ExecutionConfigProviderImpl`), the bean `ExecutionConfigProviderMock` will be used instead. This mock implements the interface `K8SDeploymentManagerMock` as well, so that the module `k8s` can use it to fake the deployment/configuration injection. `ExecutionConfigProviderMock` **will be removed** when we move to a standalone application for the executors.

In conclusion, the `ExecutorService` is not free from any fake deployment logic.


executor/src/main/java/com/redhat/service/bridge/executor/ExecutorConfigProvider.java
```diff
@@ -0,0 +1,17 @@
+package com.redhat.service.bridge.executor;
+
+import java.util.Set;
+
+public interface ExecutorConfigProvider {
```
danielezonca, line 5: I think the name is misleading because it mentions `Config` while it is a provider of the actual `Executor` instances so I would use `ExecutorProvider`. Wdyt?

executor/src/main/java/com/redhat/service/bridge/executor/ExecutorsService.java
```diff
@@ -31,49 +29,37 @@
...
+    public CompletionStage<Void> processBridgeEvent(final Message<String> message) {
         try {
-            CloudEvent cloudEvent = CloudEventUtils.decode(event);
+            CloudEvent cloudEvent = CloudEventUtils.decode(message.getPayload());
             BridgeCloudEventExtension bridgeCloudEventExtension = ExtensionProvider.getInstance().parseExtension(BridgeCloudEventExtension.class, cloudEvent);
-            Set<Executor> executors = bridgeToProcessorMap.get(bridgeCloudEventExtension.getBridgeId());
+            String bridgeId = bridgeCloudEventExtension.getBridgeId();
+            Set<Executor> executors = executorConfigProvider.getExecutors(bridgeId);
+            if (executors == null) {
+                LOG.info("[executor] A message for BridgeID {} has been received, but no executors were found.", bridgeId);
```
danielezonca, line 47: Can this happen? I expect it should be impossible to deploy a runner without any executor so I would at least move log from `info` to `error`. Wdyt?

executor/src/main/java/com/redhat/service/bridge/executor/ExecutorsService.java
```diff
@@ -31,49 +29,37 @@
...
+            }
             if (executors != null) {
                 for (Executor e : executors) {
                     try {
                         e.onEvent(cloudEvent);
                     } catch (Throwable t) {
                         // Inner Throwable catch is to provide more specific context around which Executor failed to handle the Event, rather than a generic failure
-                        LOG.error("Processor with id '{}' on bridge '{}' failed to handle Event.", e.getProcessor().getId(), e.getProcessor().getBridge().getId(), t);
+                        LOG.error("Processor with id '{}' on bridge '{}' failed to handle Event. The message is acked anyway.", e.getProcessor().getId(), e.getProcessor().getBridge().getId(), t);
```
danielezonca, line 55: Let's keep `[executor]` prefix also here. Wdyt?
```suggestion
                        LOG.error("[executor] Processor with id '{}' on bridge '{}' failed to handle Event. The message is acked anyway.", e.getProcessor().getId(), e.getProcessor().getBridge().getId(), t);
```

executor/src/main/java/com/redhat/service/bridge/executor/ExecutorsService.java
```diff
@@ -31,49 +29,37 @@
...
                     }
                 }
             }
         } catch (Throwable t) {
-            LOG.error("Failed to handle Event received on Bridge.", t);
+            LOG.error("Failed to handle Event received on Bridge. The message is acked anyway.", t);
```
danielezonca, line 60: Same here
```suggestion
            LOG.error("[executor] Failed to handle Event received on Bridge. The message is acked anyway.", t);
```

executor/src/main/java/com/redhat/service/bridge/executor/ExecutorsService.java
r00ta: @danielezonca at the moment, with the all-in-one application, it is possible: the user can 
1) create the Bridge instance
3) **NO PROCESSORS ARE CREATED FOR THE BRIDGE**
3) push events to the ingress
3) the all-in-one-executor will consume every message and fetch the Executors by Bridge ID -> if the user did not create any of them, we just ignore the message because no processors have been created. 

ATM, it is not a `warn` or an `error` because the user can upload events without having created a Processor. 
We might turn it into a `debug` message actually, but I think it's fine to have it as `info` in this early phase.

executor/src/main/java/com/redhat/service/bridge/executor/K8SDeploymentManagerMock.java
```diff
@@ -0,0 +1,9 @@
+package com.redhat.service.bridge.executor;
+
+import com.redhat.service.bridge.infra.dto.ProcessorDTO;
+
+public interface K8SDeploymentManagerMock {
```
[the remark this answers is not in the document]
r00ta: @danielezonca I don't think so. I'll rename `K8SDeploymentManagerMock` to `ExecutorDeploymentManager` because it might help to get the right idea. 

The application configuration is Injected by the bean that implements `ExecutorConfigProvider` (I will rename it soon). Until we use the all-in-one application we provide a dummy implementation of `ExecutorConfigProvider`: `ExecutorConfigProviderMock`. The `ExecutorConfigProviderMock` implements `K8SDeploymentManagerMock` as well, so that we can include the "deployment" logic behind the scenes/without touching any core part of the executor.
On the other side, the `k8s` module will inject a  `K8SDeploymentManagerMock` bean, so to control the deployment/undeployment of the resources.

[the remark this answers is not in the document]
r00ta: In other words, `K8SDeploymentManagerMock` is the bean that must stay in the module that has to be deployed/undeployed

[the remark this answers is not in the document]
r00ta: Sorry @danielezonca , now I got what you meant: you asked to move this interface to the k8s module and we keep the bean that implements it in the executor. 
At the moment the dependency is `k8s` -> `executor`. 
It would be ok to invert it, but the `shard` atm imports `k8s` as dependency and uses it in some tests: we would have to provide a mock for that bean and do some other stuff to make it work. 

Since the entire k8s module will be thrown away soonish I think it's fine as it is, but if the switch to k8s will take longer we can work on https://issues.redhat.com/browse/MGDOBR-48

danielezonca (approved):

PR #34, opened 2021-09-21
MGDOBR-10: add Executor filter evaluation with FEEL engine

r00ta:
This PR aims to evaluate the filters when a new event is received by the executor. We use the FEEL engine for the evaluation. 

 The executor uses the interfaces `TemplateFactory` and `FilterEvaluator`, and we just provide the feel implementation with `FilterEvaluatorFEEL` and `TemplateFactoryFEEL` to build and evaluate the templates.


executor/src/main/java/com/redhat/service/bridge/executor/Executor.java
```diff
@@ -15,14 +20,30 @@
...
 
-        //TODO - consider if the CloudEvent needs cleaning up from our extensions before it is handled by Actions
+        if (matchFilters(cloudEvent)) {
+            LOG.info("[executor] Filters of processor '{}' matched for event with id '{}'", processor.getId(), cloudEvent.getId());
+            // TODO - CALL ACTIONS;
+            // TODO - consider if the CloudEvent needs cleaning up from our extensions before it is handled by Actions
+            return;
+        }
+        LOG.debug("[executor] Filters of processor '{}' did not match for event with id '{}'", processor.getId(), cloudEvent.getId());
+        // DO NOTHING;
```
danielezonca, line 46: I think an `if`/`else` without the `return` is easier to read
```suggestion
        if (matchFilters(cloudEvent)) {
            LOG.info("[executor] Filters of processor '{}' matched for event with id '{}'", processor.getId(), cloudEvent.getId());
            // TODO - CALL ACTIONS;
            // TODO - consider if the CloudEvent needs cleaning up from our extensions before it is handled by Actions
        }
        else {
            LOG.debug("[executor] Filters of processor '{}' did not match for event with id '{}'", processor.getId(), cloudEvent.getId());
            // DO NOTHING;
        }
```

executor/pom.xml
```diff
@@ -30,6 +30,11 @@
       <groupId>com.redhat.service.bridge</groupId>
       <artifactId>infra</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.kie</groupId>
+      <artifactId>kie-dmn-feel</artifactId>
+      <version>7.57.0.Final</version>
```
danielezonca, line 36: Move version to depMan section.
Btw we can also consider to import `io.quarkus.platform:quarkus-kogito-bom` so that this dependency should be declared as transitive dependency

executor/src/main/java/com/redhat/service/bridge/executor/Executor.java
```diff
@@ -15,14 +20,30 @@
 
     private final ProcessorDTO processor;
 
-    public Executor(ProcessorDTO processor) {
+    private final Set<String> templates;
+
+    private final FilterEvaluator filterEvaluator;
+
+    private final TemplateFactory templateFactory;
```
danielezonca, line 27: I don't like this `template` intermediate step, I would hide the `TemplateFactory` inside the `FilterEvaluator` and accept `Set<BaseFilter>` directly.
If you want to "parse" the filter only once you can rename `FilterEvaluator` as `FilterEvaluatorBuilder` and adds a `FilterEvaluatorInstance` concept that is a specific instance tied to a set of filters.
Current decoupling implies that `TemplateFactory` and `FilterEvaluator` are related (both needs to be the FEEL version) but this is not enforced in the code.
Wdyt?

robpblake (approved):
Very nice!

executor/src/main/java/com/redhat/service/bridge/executor/Executor.java
r00ta: Fair point. I've refactored the code with the following logic: 

```java
public interface FilterEvaluatorFactory {
    FilterEvaluator build(Set<BaseFilter> filters);
}
```
and 
```java
public interface FilterEvaluator {
    boolean evaluateFilters(Map<String, Object> data);
}
```

And we provide a FEEL implementation `FilterEvaluatorFactoryFEEL` and `FilterEvaluatorFEEL`. The `Executor` is simply built with ` new Executor(processorDTO, filterEvaluatorFactory.build(processorDTO.getFilters()))` and it's completely trasparent for the consumer 

```java
        if (filterEvaluator.evaluateFilters(CloudEventUtils.getMapper().convertValue(cloudEvent, Map.class))) {
            LOG.info("[executor] Filters of processor '{}' matched for event with id '{}'", processor.getId(), cloudEvent.getId());
            // TODO - CALL ACTIONS;
            // TODO - consider if the CloudEvent needs cleaning up from our extensions before it is handled by Actions
            return;
        }
```

danielezonca (approved):

PR #44, opened 2021-09-22
Refactoring actions api

r00ta:


robpblake:
@r00ta As I mentioned in my PR yesterday, my only concern is that have we have add a sub-class for every Action that we support in this approach. That's fine when the number of actions is small, but longer-term I don't think we want to do that. My instinct is that we'd want to validate Actions and the user-submitted parameters by fetching a schema from a registry

There isn't a whole lot of need to work with the "typed" action class in the rest of the code. It's useful for readability in the tests for example, but in reality the rest of the code doesn't really need it

danielezonca:
@r00ta 
How do you want to proceed with this PR considering that https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/45 has been merged?
Do you want to update it based on [my proposal](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/45#discussion_r715471902) or do you prefer to close it for now?
I don't think it is compatible with what we already have.
Wdyt?

r00ta:
The code changed too much and it does not make much sense to rebase this. I'll refactor this in another PR in the future eventually

PR #45, opened 2021-09-22
[MGDOBR-11] Shard side of implementation for "Send to Kafka" Action

robpblake:
[MGDOBR-11](https://issues.redhat.com/browse/MGDOBR-11) 

This PR adds the shard-side implementation for the "send to Kafka" Action. 

In addition I've updated feedback from the original work to make it clearer as to the intent of how we support Actions in a generic fashion in OpenBridge. So I have now:

1.  Created an `actions` module in our project

2. Created an `ActionProvider` interface that requires that authors of an Action provide a way to validate the parameters of an Action and also a way to invoke the Action
3. Consolidated all Action-specific code into the `actions` module, so the Manager and Executor are now generic and simply request Actions by the `type` the user has sent in their request.

Using this approach, we can now add Actions to OB by only working in the `actions` module. Nothing in the manager/shard/executor needs to change when a new Action is added. 


executor/src/main/java/com/redhat/service/bridge/executor/actions/ActionInvokers.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public ActionInvoker build(ProcessorDTO processor) {
+        Optional<ActionInvokerFactory> invokerFactory = factories.stream().filter((f) -> f.accepts(processor.getAction())).findFirst();
+        if (invokerFactory.isPresent()) {
+            return invokerFactory.get().build(processor, processor.getAction());
+        }
+
+        throw new ActionInvokerException(String.format("No registered factory to build ActionInvoker for type '%s'", processor.getAction().getType()));
+    }
+}
```
r00ta, line 25: If I understand the code properly you might be looking for an abstract factory pattern. 

You can rename `ActionInvokers` with `ActionFactoryProducer` with a `getFactory` method and return the `ActionInvokerFactory` that matches it. Then you can pass `ActionInvokerFactory` into the `Executor` and so on.

executor/src/main/java/com/redhat/service/bridge/executor/actions/kafka/KafkaTopicActionInvokerFactory.java
```diff
@@ -0,0 +1,67 @@
...
+                throw new ActionInvokerException(
+                        String.format("The requested topic '%s' for Action on Processor '%s' for bridge '%s' does not exist", requiredTopic, processor.getId(), processor.getBridge().getId()));
+            }
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            throw new ActionInvokerException(String.format("Unable to list topics from Kafka to check requested topic '%s' exists for Action on Processor '%s' on bridge '%s'", requiredTopic,
+                    processor.getId(), processor.getBridge().getId()), e);
+        }
+
+        return new KafkaTopicActionInvoker(emitter, processor, requiredTopic);
+    }
```
r00ta, line 66: Just a small note... with the current implementation, if the user creates a Processor with a KafkaTopicAction with `topic=test` **before** the topic `test` is created manually by the user on the kafka cluster, the application will raise an exception. 
I think the pod will be marked as unhealthy and will be redeployed by the k8s controller, or we just put it as a requirement. Anyway, in the meanwhile we have just to make sure we don't crash the all-in-one application and we manage this case properly

executor/src/test/java/com/redhat/service/bridge/executor/actions/kafka/KafkaTopicActionInvokerFactoryTest.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @BeforeEach
+    public void beforeEach() throws Exception {
+        mockKafkaAdmin();
+    }
+
+    @Test
+    public void accepts() {
+
```
r00ta, line 82: remove newline or add it to every test (there are multiple tests where there is a newline at the beginning)

r00ta (commented):
Some comments from my side

executor/src/main/java/com/redhat/service/bridge/executor/Executor.java
```diff
@@ -19,9 +20,12 @@
 
     private final FilterEvaluator filterEvaluator;
 
-    public Executor(ProcessorDTO processor, FilterEvaluatorFactory factory) {
+    private final ActionInvoker actionInvoker;
+
+    public Executor(ProcessorDTO processor, FilterEvaluatorFactory factory, ActionInvoker actionInvoker) {
```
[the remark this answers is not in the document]
danielezonca: I prefer to keep the build method invocation here for both factories: filters, actions and transformations are related to processorDTO so it should not be possible to provide a set of filter or an action that is not defined in the processorDTO while now this is possible because the executor class has no control of actionInvoker

executor/src/main/java/com/redhat/service/bridge/executor/actions/kafka/KafkaTopicActionInvoker.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Override
+    public void onEvent(CloudEvent cloudEvent) {
+
+        /*
+         * As the user can specify their target topic in the Action configuration, we set
+         * it in the metadata of the message we are sending.
+         * 
+         * We use the Processor id for the key on the event to Kafka to ensure ordering of events
+         * emitted by actions on this processor.
```
danielezonca, line 38: If we make this choice we need to pay attention to some aspects: as far as I remember a topic can be configured to be [log compacted](https://towardsdatascience.com/log-compacted-topics-in-apache-kafka-b1aa1e4665a7) and in this case messages with same keys are removed preserving only the last occurrence (but if we control the creation of the topic this should not be critical). Moreover if a processor produces way more messages compared to the other we will have a single Kafka partition "overloaded" while other partitions are ~empty.
We can stay with this for now and maybe revisit in the future if necessary. We need to find a proper way to track similar decisions like a WIP design doc.
Wdyt?

executor/src/main/java/com/redhat/service/bridge/executor/actions/kafka/KafkaTopicActionInvokerFactory.java
```diff
@@ -0,0 +1,67 @@
...
+    public ActionInvoker build(ProcessorDTO processor, BaseAction baseAction) {
+
+        String requiredTopic = baseAction.getParameters().get(KafkaTopicAction.KAFKA_ACTION_TOPIC_PARAM);
+        if (requiredTopic == null) {
+            throw new ActionInvokerException(
+                    String.format("There is no topic specified in the parameters for Action on Processor '%s' on Bridge '%s'", processor.getId(), processor.getBridge().getId()));
+        }
+
+        try {
+            Set<String> strings = adminClient.listTopics().names().get(DEFAULT_LIST_TOPICS_TIMEOUT, DEFAULT_LIST_TOPICS_TIMEUNIT);
```
danielezonca, line 55: I'm not sure about this check: adminApi can be disabled in some Kafka installation and moreover this doesn't prevent user to removed a topic later 🤔

executor/src/main/java/com/redhat/service/bridge/executor/actions/kafka/KafkaClients.java
```diff
@@ -0,0 +1,37 @@
...
+    @Inject
+    @Identifier("default-kafka-broker")
+    Map<String, Object> config;
+
+    /*
+     * Taken from: https://quarkus.io/guides/kafka#kafka-bare-clients
+     */
+    @Produces
+    @ApplicationScoped
+    AdminClient adminClient() {
```
danielezonca, line 28: If we remove AdminClient usage we can remove this too of course

executor/src/main/resources/application.properties
```diff
@@ -6,3 +6,7 @@ mp.messaging.incoming.events-in.connector=smallrye-kafka
 mp.messaging.incoming.events-in.topic=events
 mp.messaging.incoming.events-in.value.serializer=org.apache.kafka.common.serialization.StringSerializer
 mp.messaging.incoming.events-in.auto.offset.reset=earliest
+
+# Configuring the outgoing channel (writing to Kafka) for Actions
+mp.messaging.outgoing.actions-out.connector=smallrye-kafka
+mp.messaging.outgoing.actions-out.value.serializer=org.apache.kafka.common.serialization.StringSerializer
```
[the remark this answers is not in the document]
r00ta: Fine for me, but does this impact your long term plan for Actions @robpblake ? We discussed it yesterday, but since I did not get it I'm wondering if this fits it

executor/src/main/java/com/redhat/service/bridge/executor/actions/kafka/KafkaTopicActionInvoker.java
robpblake: Thanks for the link! 

Giving it some thought, this actually feels like something we should expose as part of the Action parameters so that the user can decide what they want to do in terms of using a `key`. So I'll remove this behaviour for now so that we start with the default behaviour which is not to use a key and then gradually add the functionality when we see a need for it.

In terms of capturing how Actions will work, I'd hope that we'd start to produce documentation for Actions where we can capture behaviour like this (or show the user how to configure it).

executor/src/main/java/com/redhat/service/bridge/executor/actions/kafka/KafkaTopicActionInvokerFactory.java
robpblake: I think for this iteration it will help us get things up and running and capture obvious errors early. We also know that we can pretty much guarantee admin API is enabled for MK at this point, so I'd say +1 to keep it for now

Longer-term, I'd say we move it to a `param` on the Action e.g `checkTopicExists=true` which can then make the assumption that we have admin API capabilities from the credentials that the user has provided.

If that works for you, I'll add an issue to the backlog to make the topic check on Action provisioning a configurable thing?

executor/src/main/java/com/redhat/service/bridge/executor/actions/kafka/KafkaClients.java
robpblake: See previous comment.

executor/src/main/java/com/redhat/service/bridge/executor/actions/kafka/KafkaTopicActionInvokerFactory.java
robpblake: @r00ta The Exception is handled in the `DeploymentsManagerImpl` so that we return a `FAILED` state to the Fleet Manager if we fail to deploy the Action.

executor/src/main/resources/application.properties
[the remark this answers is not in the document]
robpblake: @danielezonca  / @r00ta I've refactored to address this now as I think it helps to make my approach to Actions clearer. Please take a look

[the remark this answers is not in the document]
r00ta: Hi @robpblake , I think your refactoring is not following @danielezonca 's suggestion. 

He is suggesting to create **one** module **for each** action. We can have a generic module `actions` where we keep interfaces/api, but then every action is in a separated module `kafka-action`, `webhook-action` (eventually splitted between `kafka-action-api` and `kafka-action-core` for example), where we provide the beans that implement the factory interface.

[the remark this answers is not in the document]
robpblake: @r00ta Yeah I decided not to go to that level of decomposition just yet, keeping everything in one module for now. I'd like to see what comes out of https://issues.redhat.com/browse/MGDOBR-55 as to what other actions might  look like before we start to fully break it down like that. 

So this is the half-way option: Everything is centralised in a single location to avoid it spreading throughout the codebase e.g. not in `manager`, `shard`, `infra`, and if we want to further decompose in the future, we can.

actions/src/main/java/com/redhat/service/bridge/actions/ActionProvider.java
```diff
@@ -0,0 +1,14 @@
+package com.redhat.service.bridge.actions;
+
+import com.redhat.service.bridge.infra.models.actions.BaseAction;
+import com.redhat.service.bridge.infra.models.dto.ProcessorDTO;
+
+public interface ActionProvider {
+
+    String getType();
```
danielezonca, line 8: What about moving here the `accept` logic instead of `ActionProviderFactory`?
```suggestion
    String getType();
    
    default boolean accept(String actionType) {
        return getType().equals(actionType);
    }
```

actions/src/main/java/com/redhat/service/bridge/actions/ActionProviderFactory.java
```diff
@@ -0,0 +1,23 @@
...
+import javax.inject.Inject;
+
+@ApplicationScoped
+public class ActionProviderFactory {
+
+    @Inject
+    Instance<ActionProvider> actionProviders;
+
+    public ActionProvider getActionProvider(String actionType) {
+        Optional<ActionProvider> ap = actionProviders.stream().filter((a) -> a.getType().equals(actionType)).findFirst();
```
danielezonca, line 16: In this case this becomes (please check if correct because I'm changing the code directly here on GitHub
```suggestion
        Optional<ActionProvider> ap = actionProviders.stream().filter(ActionProvider::accept).findFirst();
```

actions/src/main/java/com/redhat/service/bridge/actions/kafkatopic/KafkaTopicActionValidator.java
```diff
@@ -0,0 +1,19 @@
...
+public class KafkaTopicActionValidator implements ActionParameterValidator {
+
+    @Override
+    public boolean isValid(BaseAction baseAction) {
+        if (baseAction.getParameters() != null) {
+            String topic = baseAction.getParameters().get(KafkaTopicAction.TOPIC_PARAM);
+            return topic != null && !topic.isEmpty();
+        }
+        return false;
+    }
```
danielezonca, line 18: @robpblake @r00ta 
Here is where I was thinking at the possibility to plug an [Hibernate Validator based validator](https://quarkus.io/guides/validation#json-mapping-and-validation) (and this can be different for each action)
It marshall/unmarshall the object so it is a bit unefficient but it is done only once during the saving of the action in the Manager so I think it is not an issue and we preserve module isolation/modularisation.
Wdyt?
(I have also suggested to return the list of violation instead of a boolean in the next comment so that we don't lose proper error reporting)
```suggestion
    @Inject
    Validator validator;
    
    @Inject
    ObjectMapper om;

    @Override
    public boolean isValid(BaseAction baseAction) {
        CustomActionBean bean = om.readValue(om.writeValueAsString(baseAction), CustomActionBean.class);
        Set<ConstraintViolation<Book>> violations = validator.validate(bean);
        return violations.isEmpty();
    }
    
    public static class CustomActionBean {
        @NotNull
        ....
    }
```

actions/src/main/java/com/redhat/service/bridge/actions/ActionParameterValidator.java
```diff
@@ -0,0 +1,7 @@
+package com.redhat.service.bridge.actions;
+
+import com.redhat.service.bridge.infra.models.actions.BaseAction;
+
+public interface ActionParameterValidator {
+    boolean isValid(BaseAction baseAction);
```
danielezonca, line 6: I would return a collection of violation instead of a boolean so that we can provide meaningful error reporting.
Wdyt?
```suggestion
    Set<ValidationProblem> validate(BaseAction baseAction);
```

executor/src/main/resources/application.properties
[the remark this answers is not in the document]
danielezonca: Yes I expect each action will have a different module but more in general ~all this code will be changed when we split `manager` because validation of the action can be probably moved in a `my-action-api` module in the `manager` while the execution semantic can stay closer to the `executor` in a `my-action-runtime` but for now I think it is enough to have a single `actions` module 🤔

actions/src/main/java/com/redhat/service/bridge/actions/kafkatopic/KafkaTopicActionValidator.java
robpblake: Yep, works for me.

actions/src/main/java/com/redhat/service/bridge/actions/ActionParameterValidator.java
robpblake: I was thinking to return a `ValidationResult` with `isValid()` and a `getMessage()` method. The message contains the validation error string and can be constructed in any way by the `ActionParameterValidator`. So final code here would look something like:

```
ValidationResult validationResult = actionProvider.getParameterValidator().isValid(baseAction);
if(!validationResult.isValid()) {
   context.buildConstraintViolationWithTemplate(validationResult.getMessage()).addConstraintViolation();
}
return validationResult.isValid();
```

danielezonca: Fine for me 👍

danielezonca (approved):
Thanks for the PR 👍

r00ta (approved):

PR #47, opened 2021-09-23
MGDOBR-8: Add processor delete operation in manager and shard

r00ta:
[MGDOBR-8](https://issues.redhat.com/browse/MGDOBR-8) 

### DIFF
--- a/infra/src/main/java/com/redhat/service/bridge/infra/k8s/KubernetesResourceType.java
+++ b/infra/src/main/java/com/redhat/service/bridge/infra/k8s/KubernetesResourceType.java
@@ -1,0 +1,7 @@
+package com.redhat.service.bridge.infra.k8s;
+
+public enum KubernetesResourceType {
+    CUSTOM_RESOURCE,
+    DEPLOYMENT,
+    INGRESS
+}
--- a/manager/src/main/java/com/redhat/service/bridge/manager/ProcessorService.java
+++ b/manager/src/main/java/com/redhat/service/bridge/manager/ProcessorService.java
@@ -1,107 +1,26 @@
 package com.redhat.service.bridge.manager;
 
-import java.time.ZonedDateTime;
 import java.util.List;
-import java.util.Set;
-import java.util.stream.Collectors;
-
-import javax.enterprise.context.ApplicationScoped;
-import javax.inject.Inject;
-import javax.transaction.Transactional;
 
 import com.redhat.service.bridge.infra.models.dto.BridgeStatus;
 import com.redhat.service.bridge.infra.models.dto.ProcessorDTO;
 import com.redhat.service.bridge.manager.api.models.requests.ProcessorRequest;
-import com.redhat.service.bridge.manager.dao.ProcessorDAO;
-import com.redhat.service.bridge.manager.exceptions.AlreadyExistingItemException;
-import com.redhat.service.bridge.manager.exceptions.BridgeLifecycleException;
-import com.redhat.service.bridge.manager.exceptions.ItemNotFoundException;
-import com.redhat.service.bridge.manager.models.Action;
-import com.redhat.service.bridge.manager.models.Bridge;
-import com.redhat.service.bridge.manager.models.Filter;
 import com.redhat.service.bridge.manager.models.ListResult;
 import com.redhat.service.bridge.manager.models.Processor;
 
-@Transactional
-@ApplicationScoped
-public class ProcessorService {
+public interface ProcessorService {
 
-    @Inject
-    ProcessorDAO processorDAO;
+    Processor getProcessor(String processorId, String bridgeId, String customerId);
 
-    @Inject
-    BridgesService bridgesService;
+    Processor createProcessor(String bridgeId, String customerId, ProcessorRequest processorRequest);
 
-    public Processor getProcessor(String processorId, String bridgeId, String customerId) {
+    List<Processor> getProcessorByStatuses(List<BridgeStatus> statuses);
 
-        Bridge bridge = bridgesService.getBridge(bridgeId, customerId);
-        Processor processor = processorDAO.findByIdBridgeIdAndCustomerId(processorId, bridge.getId(), bridge.getCustomerId());
-        if (processor == null) {
-            throw new ItemNotFoundException(String.format("Processor with id '%s' does not exist on Bridge '%s' for customer '%s'", processorId, bridgeId, customerId));
-        }
+    Processor updateProcessorStatus(ProcessorDTO processorDTO);
 
-        return processor;
-    }
+    Long getProcessorsCount(String bridgeId, String customerId);
 
-    public Processor createProcessor(String bridgeId, String customerId, ProcessorRequest processorRequest) {
-        Bridge bridge = getAvailableBridge(bridgeId, customerId);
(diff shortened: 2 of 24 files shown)


manager/src/main/java/com/redhat/service/bridge/manager/ProcessorService.java
```diff
@@ -1,107 +1,26 @@
...
-        p.setStatus(BridgeStatus.REQUESTED);
-        p.setBridge(bridge);
-        p.setFilters(filters);
-
-        Action a = new Action();
-        a.setName(processorRequest.getAction().getName());
-        a.setParameters(processorRequest.getAction().getParameters());
-        a.setType(processorRequest.getAction().getType());
-        p.setAction(a);
+public interface ProcessorService {
```
r00ta, line 11: Moved to an interface... unfortunately github shows a complete rewrite, but this is not the case

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
```diff
@@ -0,0 +1,131 @@
...
+                    processorDTO.getBridge().getCustomerId()));
+        }
+        p.setStatus(processorDTO.getStatus());
+
+        if (processorDTO.getStatus().equals(BridgeStatus.DELETED)) {
+            processorDAO.deleteById(processorDTO.getId());
+        }
+
+        return p;
+    }
```
r00ta, line 101: This has been updated

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Override
+    public void deleteProcessor(String bridgeId, String processorId, String customerId) {
+        Processor processor = processorDAO.findByIdBridgeIdAndCustomerId(processorId, bridgeId, customerId);
+        processor.setStatus(BridgeStatus.DELETION_REQUESTED);
+        LOGGER.info("[manager] Processor with id '{}' for customer '{}' on bridge '{}' has been marked for deletion", processor.getId(), processor.getBridge().getCustomerId(),
+                processor.getBridge().getId());
+    }
+
```
r00ta, line 121: new method

infra/src/main/java/com/redhat/service/bridge/infra/k8s/KubernetesResourceType.java
```diff
@@ -0,0 +1,7 @@
+package com.redhat.service.bridge.infra.k8s;
+
+public enum KubernetesResourceType {
```
robpblake, line 3: @r00ta What is this enum used for? I can see you adding it to the `ResourceEvent`, but I'm not sure what we use it for?

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorService.java
danielezonca: Is this just a refactor or `ProcessorServiceImpl` contains also some changes?

shard/src/main/java/com/redhat/service/bridge/shard/controllers/IngressController.java
```diff
@@ -51,30 +51,39 @@
...
+            /*
+             * If the CRD is deleted, remove also all the other related resource
+             *
+             * If another dependent resource is deleted, the `reconcileExecutor` will catch the mismatch between the expected state and the current state.
+             * It will redeploy the resources so to reach the expected status at the end.
+             */
+            if (event.getAction().equals(Action.DELETED) && event.getResourceType().equals(KubernetesResourceType.CUSTOM_RESOURCE)) {
+                delete(event.getResourceId());
+                return;
+            }
```
danielezonca, line 66: If we properly set [CR ownership and cascading](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#ownership-and-finalizers) this should become OOTB when moved to K8s: if the BridgeCR owns all the ProcessorCR, when it is removed, all dependents resources are automatically removed.
What about adding a TODO comment to track this so that when we move to K8s we properly check the configuration?

shard/src/main/java/com/redhat/service/bridge/shard/controllers/ProcessorController.java
```diff
@@ -37,22 +39,31 @@
...
+            /*
+             * If the CRD is deleted, remove also all the other related resource
+             *
+             * If another dependent resource is deleted, the `reconcileExecutor` will catch the mismatch between the expected state and the current state.
+             * It will redeploy the resources so to reach the expected status at the end.
+             */
+            if (event.getAction().equals(Action.DELETED) && event.getResourceType().equals(KubernetesResourceType.CUSTOM_RESOURCE)) {
+                delete(event.getResourceId());
+                return;
+            }
```
danielezonca, line 52: Same comment as above (even if I'm not sure if an Action becomes a CR or not)

infra/src/main/java/com/redhat/service/bridge/infra/k8s/KubernetesResourceType.java
r00ta: @robpblake basically when a resource is updated/deleted the controllers get called
```
    void onEvent(@Observes ResourceEvent event) { // equivalent of ResourceEventSource for operator sdk
        if (event.getSubject().equals(K8SBridgeConstants.BRIDGE_TYPE)) {

            /*
             * If the CRD is deleted, remove also all the other related resource
             *
             * If another dependent resource is deleted, the `reconcileExecutor` will catch the mismatch between the expected state and the current state.
             * It will redeploy the resources so to reach the expected status at the end.
             */
            if (event.getAction().equals(Action.DELETED) && event.getResourceType().equals(KubernetesResourceType.CUSTOM_RESOURCE)) {
                delete(event.getResourceId());
                return;
            }
```
in order to properly notify that the custom resource has been deleted and that all the related resources have to be cleaned up, I had to include the `KubernetesResourceType` in the `ResourceEvent`. 

With the operator-sdk we already get called only when the CRD is deleted (see https://github.com/r00ta/baaas-decision-fleet-shard/blob/c173edb2fdd3d02abbb9e2cfc1fa23f71b99b1d6/decision-fleet-shard-operator/src/main/java/org/kie/baaas/dfs/controller/DecisionVersionController.java#L90). So our mock is not 100% matching the real api but it does not really matter since it will be thrown away and a small refactoring will be necessary.

manager/src/main/java/com/redhat/service/bridge/manager/api/internal/ShardBridgesSyncAPI.java
```diff
@@ -38,7 +38,7 @@
     BridgesService bridgesService;
 
     @Inject
-    ProcessorService processorService;
+    ProcessorServiceImpl processorService;
```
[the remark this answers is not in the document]
r00ta: wops, it's a leftover. thanks for the catch

shard/src/main/java/com/redhat/service/bridge/shard/controllers/IngressController.java
r00ta: Yes good point

danielezonca (approved):

robpblake (approved):

PR #81, opened 2021-10-20
MGDOBR-86: create an executable for each component

r00ta:
[MGDOBR-86](https://issues.redhat.com/browse/MGDOBR-86) 

This PR aims to create an executable for each component. The components are not supposed to be working already, they still have some cross dependencies that will cause a crash: we need to start creating an executable so to update the CD pipelines and push the docker images to quay. 

- [x] All new functionality is tested and uses org.assertj.core.api.Assertions for Assertions


danielezonca (commented):
I'm not sure about making each module as executable. What about create a new (empty) `executor-runner` module, a new `ingress-runner` etc?
We got issue in the past using a Quarkus runner as dependency (like jar renamed and similar) so I would like to avoid them considering that during this migration phase we will preserve the all-in-one executor and also these independent components.
Wdyt?

danielezonca (approved):
As agreed offline let's proceed 👍

PR #92, opened 2021-10-26
[MGDOBR-57] Add Error Catalog to Openbrige API

fjtirado:
Adding Error Catalog implementation

Refactoring existing code so it use inheritance and avoid duplication
of common response fields

[MGDOBR-57](https://issues.redhat.com/browse/MGDOBR-57)

### DIFF
--- a/manager/src/main/java/com/redhat/service/bridge/manager/api/user/ErrorsAPI.java
+++ b/manager/src/main/java/com/redhat/service/bridge/manager/api/user/ErrorsAPI.java
@@ -1,0 +1,59 @@
+package com.redhat.service.bridge.manager.api.user;
+
+import java.util.Optional;
+
+import javax.inject.Inject;
+import javax.validation.Valid;
+import javax.ws.rs.BeanParam;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
+
+import org.eclipse.microprofile.openapi.annotations.enums.SecuritySchemeType;
+import org.eclipse.microprofile.openapi.annotations.security.SecurityRequirement;
+import org.eclipse.microprofile.openapi.annotations.security.SecurityScheme;
+import org.eclipse.microprofile.openapi.annotations.security.SecuritySchemes;
+
+import com.redhat.service.bridge.manager.ErrorsService;
+import com.redhat.service.bridge.manager.api.models.responses.ErrorListResponse;
+import com.redhat.service.bridge.manager.api.models.responses.ErrorResponse;
+import com.redhat.service.bridge.manager.api.models.responses.ListResponse;
+import com.redhat.service.bridge.manager.models.Error;
+import com.redhat.service.bridge.manager.models.QueryInfo;
+
+import io.quarkus.security.Authenticated;
+
+import static com.redhat.service.bridge.infra.api.APIConstants.ERROR_API_BASE_PATH;
+
+@SecuritySchemes(value = {
+        @SecurityScheme(securitySchemeName = "bearer",
+                type = SecuritySchemeType.HTTP,
+                scheme = "Bearer")
+})
+@SecurityRequirement(name = "bearer")
+@Path(ERROR_API_BASE_PATH)
+@Consumes(MediaType.APPLICATION_JSON)
+@Produces(MediaType.APPLICATION_JSON)
+@Authenticated
+public class ErrorsAPI {
+
+    @Inject
+    private ErrorsService service;
+
+    @GET
+    public Response getErrors(@Valid @BeanParam QueryInfo queryInfo) {
+        return Response.ok(ListResponse.fill(service.getErrors(queryInfo), new ErrorListResponse(), ErrorResponse::from)).build();
+    }
+
+    @GET
+    @Path("{id}")
+    public Response getError(@PathParam("id") int id) {
+        Optional<Error> error = service.getError(id);
+        return (error.isPresent() ? Response.ok(ErrorResponse.from(error.get())) : Response.status(Status.NOT_FOUND)).build();
+    }
+}
--- a/manager/src/main/java/com/redhat/service/bridge/manager/api/user/ProcessorsAPI.java
+++ b/manager/src/main/java/com/redhat/service/bridge/manager/api/user/ProcessorsAPI.java
@@ -1,21 +1,16 @@
 package com.redhat.service.bridge.manager.api.user;
 
-import java.util.List;
-
 import javax.inject.Inject;
 import javax.validation.Valid;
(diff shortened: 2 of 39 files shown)


danielezonca (approved):
LGTM 👍

manager/src/main/java/com/redhat/service/bridge/manager/exceptions/AlreadyExistingItemException.java
```diff
@@ -4,6 +4,8 @@
 
 public class AlreadyExistingItemException extends EventBridgeManagerException {
 
+    private static final long serialVersionUID = 1L;
```
r00ta, line 7: Do we need `serialVersionUID` in the exception classes?

r00ta (approved):
LGTM, just a minor comment :) Well done @fjtirado !

manager/src/main/java/com/redhat/service/bridge/manager/api/user/ProcessorsAPI.java
```diff
@@ -77,19 +63,9 @@ public Response getProcessor(@NotEmpty @PathParam("bridgeId") String bridgeId, @
...
-        ListResult<Processor> processors = processorService.getProcessors(bridgeId, customerId, page, pageSize);
-        List<ProcessorResponse> px = processors.getItems().stream().map(processorService::toResponse).collect(toList());
-
-        ProcessorListResponse listResponse = new ProcessorListResponse();
-        listResponse.setItems(px);
-        listResponse.setPage(processors.getPage());
-        listResponse.setSize(processors.getSize());
-        listResponse.setTotal(processors.getTotal());
-        return Response.ok(listResponse).build();
+    public Response listProcessors(@NotEmpty @PathParam("bridgeId") String bridgeId, @BeanParam QueryInfo queryInfo) {
```
robpblake, line 66: Can you add a `@Valid` next to the `@BeanParam` to ensure it is validated.

manager/src/main/java/com/redhat/service/bridge/manager/api/user/ErrorsAPI.java
```diff
@@ -0,0 +1,58 @@
...
+@Consumes(MediaType.APPLICATION_JSON)
+@Produces(MediaType.APPLICATION_JSON)
+@Authenticated
+public class ErrorsAPI {
+
+    @Inject
+    private ErrorsService service;
+
+    @GET
+    public Response getErrors(@BeanParam QueryInfo queryInfo) {
```
robpblake, line 48: Same here, can we have `@Valid` on the `@BeanParam`.

robpblake (approved):
Thanks @fjtirado . Looks good!

Just a couple of minor comments from me 👍

manager/src/main/java/com/redhat/service/bridge/manager/exceptions/AlreadyExistingItemException.java
fjtirado: I did that to avoid an annoying warning that Eclipse displays because since Exception implements Serializable, theoretially it is a good practise to put a defaul serial in all the children. In practise, you are never going to serialize an Exception once RMI has dead for good, but nevertheless I prefer to put the line and avoid the yellow color in my screen ;)

manager/src/main/java/com/redhat/service/bridge/manager/api/user/ProcessorsAPI.java
fjtirado: sure good point thanks

manager/src/test/java/com/redhat/service/bridge/manager/dao/ProcessorDAOTest.java
```diff
@@ -164,7 +165,7 @@ public void findByBridgeIdAndCustomerId() {
     @Test
     public void findByBridgeIdAndCustomerId_noProcessors() {
         Bridge b = createBridge();
-        ListResult<Processor> listResult = processorDAO.findByBridgeIdAndCustomerId(b.getId(), TestConstants.DEFAULT_CUSTOMER_ID, 0, 100);
+        ListResult<Processor> listResult = processorDAO.findByBridgeIdAndCustomerId(b.getId(), TestConstants.DEFAULT_CUSTOMER_ID, new QueryInfo(0, 100));
         assertThat(listResult.getPage()).isZero();
```
pibizza, line 169: I would not use 0 and 100 since they have little evident semantics. Possibly you could use a constant like PAGE_0 and PAGE_SIZE_LARGE instead?

manager/src/test/java/com/redhat/service/bridge/manager/dao/ProcessorDAOTest.java
```diff
@@ -176,7 +177,7 @@ public void findByBridgeIdAndCustomerId_pageOffset() {
         Processor p = createProcessor(b, "foo");
         Processor p1 = createProcessor(b, "bar");
 
-        ListResult<Processor> listResult = processorDAO.findByBridgeIdAndCustomerId(b.getId(), TestConstants.DEFAULT_CUSTOMER_ID, 1, 1);
+        ListResult<Processor> listResult = processorDAO.findByBridgeIdAndCustomerId(b.getId(), TestConstants.DEFAULT_CUSTOMER_ID, new QueryInfo(1, 1));
         assertThat(listResult.getPage()).isEqualTo(1L);
```
pibizza, line 181: Same as above. I would use PAGE_ONE and PAGE_SIZE_ONE here. Besides you can reuse these constants in the following assertions, and make the binding between test setup and test assertions clearer.

Also you may want to consider using DEFAULT_CUSTOMER_ID as a static import, the TestConstants. prefix is just noise in this test.

manager/src/test/java/com/redhat/service/bridge/manager/dao/ProcessorDAOTest.java
```diff
@@ -164,7 +165,7 @@ public void findByBridgeIdAndCustomerId() {
     @Test
     public void findByBridgeIdAndCustomerId_noProcessors() {
         Bridge b = createBridge();
-        ListResult<Processor> listResult = processorDAO.findByBridgeIdAndCustomerId(b.getId(), TestConstants.DEFAULT_CUSTOMER_ID, 0, 100);
+        ListResult<Processor> listResult = processorDAO.findByBridgeIdAndCustomerId(b.getId(), TestConstants.DEFAULT_CUSTOMER_ID, new QueryInfo(0, 100));
```
pibizza, line 168: Two general comments on the test, even if they are not related to the PR per se.

- probably you should make Bridge b a instance variable for the test, this would remove a parameter from create processor and make the code clearer

- the assertion assertThat(listResult.getPage()).isEqualTo(1L) is very verbose and exposes lots of implementation details on how to test that the page of listResult is 1. I would replace it with a custom assertion like pageForListResultIs(1). The same for other assertions. In general I would try to make the test very terse, just providing the information that is specific for the test and hide all the other details.

pibizza (commented):
I have added some comments on one unit tests, to discuss possible improvements on unit test formats and approach. Feel free to comment and discuss.

PR #99, opened 2021-10-29
MGDOBR-69: Store whole processors as jsonb

kostola:
[MGDOBR-69](https://issues.redhat.com/browse/MGDOBR-69) 

## Scope of this PR

This PR changes the way processors are stored inside the PostgreSQL database: the whole processor definition is now stored in a single `jsonb` column inside the `processor` table. The previous implementation used separate tables to store the processor and its subcomponents (filters, action and action parameters).

The reason for this change is that with the old implementation the rigidity of such a database schema was not justified by a real need to query the internal structure of the processor while the new approach makes it easier and more flexible to store all the data.

I chose `jsonb` because it's a very efficient way to store JSON inside PostgreSQL databases and it still enables querying the data in case unexpected needs emerge.

## Changes introduced

* Added new migration that deletes the unused tables and adds the `definition` column (of type `jsonb`) to the `processor` table.
* Modified `manager` to handle the new data structure.
* Modified `manager` tests of service beans to use PostgreSQL istead of H2 because the new migration was failing. I think there's no point in using H2 now that with testcontainers it's so easy to spin up a PostgreSQL container since we want to apply migrations. Supporting H2 means either we only support the common SQL between PostgreSQL and H2 (excluding **a lot** of cool features) OR we must write different migrations whenever you encounter a difference (which is a waste of time imho).
* Created `test-utils` module to reuse common test resources between `manager` and `runner` (caused by the previous point).

## PR requirements


manager/src/main/java/com/redhat/service/bridge/manager/models/Processor.java
```diff
@@ -17,34 +14,41 @@
...
 import com.redhat.service.bridge.infra.models.dto.ProcessorDTO;
-import com.redhat.service.bridge.infra.models.filters.BaseFilter;
+import com.redhat.service.bridge.infra.models.processors.ProcessorDefinition;
 import com.redhat.service.bridge.manager.api.models.responses.ProcessorResponse;
 
-// The join fetch on the filters will produce duplicates that must be removed from the results (see https://developer.jboss.org/docs/DOC-15782#jive_content_id_Hibernate_does_not_return_distinct_results_for_a_query_with_outer_join_fetching_enabled_for_a_collection_even_if_I_use_the_distinct_keyword)
+import io.quarkiverse.hibernate.types.json.JsonBinaryType;
+import io.quarkiverse.hibernate.types.json.JsonTypes;
+
+// Outer join fetches will produce duplicates that must be removed from the results.
```
r00ta, line 32: You can remove this comment

infra/src/main/java/com/redhat/service/bridge/infra/models/processors/ProcessorDefinition.java
```diff
@@ -0,0 +1,71 @@
+package com.redhat.service.bridge.infra.models.processors;
+
+import java.util.Objects;
+import java.util.Set;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.redhat.service.bridge.infra.models.actions.BaseAction;
+import com.redhat.service.bridge.infra.models.filters.BaseFilter;
+
+public class ProcessorDefinition {
```
danielezonca, line 10: The idea of "schemaless" for processor definition is intended to be for the whole chain of processing: no Entity but also not "DTO" otherwise we are just moving the problem from DB to Manager code.
I think the conversion to an actual bean should happen (if necessary) only at Shard/Executor level while in the manager you can use JsonNode or similar

manager/src/test/resources/application.properties
```diff
@@ -1,30 +1,7 @@
-# Database
-quarkus.datasource.db-kind=h2
```
danielezonca, line 2: You should be able to remove [`quarkus-jdbc-h2`](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/manager/pom.xml#L80-L84) dependency

r00ta (commented):
Can you please update `DEMO.md` with the new processor request structure? Thanks!

infra/src/main/java/com/redhat/service/bridge/infra/models/processors/ProcessorDefinition.java
r00ta: my2c: the manager already serializes the ProcessorDefinition when the request comes (also for validation), so it already has a dependency to such structure. 
I'd keep using `ProcessorDefinition` instead of JsonNode in the DTOs, it's much simpler imo.

manager/src/main/java/com/redhat/service/bridge/manager/models/Processor.java
r00ta: Sorry, I just saw your reply in one of my previous comment. Resolving

infra/src/main/java/com/redhat/service/bridge/infra/models/processors/ProcessorDefinition.java
r00ta: Let me spend a couple of words on this topic: I think making use of the JsonNode in the manager is not enough. Let's assume we have a breaking change in the `ProcessorDefinition` from
```json
{
"filters": [ {"type": "StringEquals", "value": "jacopo", "key": "data.name"} ],
"transformationTemplate": "",
"actions": []
}
```
to
```json
"filters": [ {"type": "StringEquals", "filterValue": "jacopo", "key": "data.name"} ],
"transformationTemplate":"",
"actions": []
```
i.e. `value` turned into `filterValue` (we can think about more realistic examples, but it's just to give the idea). The manager can simply store the (unstructured) data into the `processor_definition` column, and return it to every shard as it is. 

The problem is that now the shards must know how they have to deserialize the object. In theory with the endpoint `/api/v1/shard/processors` it might even get processors with different schemas!

If we have a breaking change we have different options
1) we have to migrate the existing data
2) we have to store the new data in a different column (or add a version property in the data) so that the manager can expose the processors depending on their schema

manager/src/test/java/com/redhat/service/bridge/manager/BridgesServiceIT.java
```diff
@@ -14,14 +14,17 @@
...
+import io.quarkus.test.common.QuarkusTestResource;
 import io.quarkus.test.junit.QuarkusTest;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 
 @QuarkusTest
-public class BridgesServiceTest {
+@QuarkusTestResource(PostgresResource.class)
+public class BridgesServiceIT {
```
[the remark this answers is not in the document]
robpblake: @kostola My ask is that we're consistent across everything and that it's clear to everyone how to name things. So for example, `ProcessorDAOTest` is using the DB, but you haven't renamed that one to have the  `IT` suffix. Should we?

My other question is: what is the benefit to adding the `IT` suffix at this point? We may have requirements in the future to create an integration Test suite, but again the "how" of that would be considered in that story.

Does that make sense?

infra/src/main/java/com/redhat/service/bridge/infra/models/processors/ProcessorDefinition.java
robpblake: As per some of the issues raised by @r00ta , I would expect any update to our API structures to be rolled out gradually and by definition we'd have to support N-1 compatibility in the Manager for a period of time. So for example changing the data structure for the ProcessorDefinition, I'd expect something like this as a flow of deployment (this would happen over multiple deployments into prod)

1. Create the new schema and add a new column to the DB. Write a migration to copy all existing data to the new version and update the Manager to dual write into both columns
2. Update the  the API on the Manager that allows retrieval of either old or new version of the data structure
3. Update the shard to be able to work with new version of the data structure and deploy it to a subset of our shards
4. Monitor things for a bit and assuming all good, rollout to all shards
5. Once we're happy, remove the dual write in the Manager and drop the legacy column

manager/src/test/java/com/redhat/service/bridge/manager/BridgesServiceIT.java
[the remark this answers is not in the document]
kostola: yes let's keep this as a separate discussion and revert back to original names for this PR

infra/src/main/java/com/redhat/service/bridge/infra/models/processors/ProcessorDefinition.java
kostola: Following our sync call on this topic, I implemented the `definition` field as `JsonNode` only in the model and kept every other DTO / Request / Response as it is. We'll discuss the plan for versioning and data migration separately.

manager/src/test/resources/application.properties
kostola: Removed :+1:

kostola:
> Can you please update `DEMO.md` with the new processor request structure? Thanks!

No need to do this since the user APIs are not changed by this PR.

manager/src/main/java/com/redhat/service/bridge/manager/models/Bridge.java
```diff
@@ -55,7 +52,7 @@
 
     @Column(name = "status")
     @Enumerated(EnumType.STRING)
-    private BridgeStatus status;
+    public BridgeStatus status;
```
danielezonca, line 55: Why this change?

danielezonca (approved):

r00ta (approved):

manager/src/main/java/com/redhat/service/bridge/manager/models/Bridge.java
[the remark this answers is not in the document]
kostola: Nice catch...this is a leftover of a wrong automatic refactoring by IntelliJ 😅

kostola: It's a leftover of a wrong automatic refactoring by IntelliJ 😅

kostola: Fixed :+1:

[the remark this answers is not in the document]
kostola: Fixed :+1:

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
```diff
@@ -131,6 +127,52 @@ public void deleteProcessor(String bridgeId, String processorId, String customer
...
+        if (processor.getBridge() != null) {
+            processorResponse.setHref(APIConstants.USER_API_BASE_PATH + processor.getBridge().getId() + "/processors/" + processor.getId());
+            processorResponse.setBridge(bridgesService.toResponse(processor.getBridge()));
+        }
+
+        return processorResponse;
+    }
+
+    @Override
+    public JsonNode definitionToJsonNode(ProcessorDefinition definition) {
```
robpblake, line 163: @kostola This and the `jsonNodeToDefinition` methods are inheriting the `@Transactional` class-level annotation. They don't need a Transaction and given how frequently they are going to be called, I think we should remove it. Can you please remove the class level `@Transactional` annotation and move it only to the methods where a transaction is required.

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
```diff
@@ -117,4 +119,28 @@ public Bridge updateBridge(BridgeDTO bridgeDTO) {
         LOGGER.info("[manager] Bridge with id '{}' has been updated for customer '{}'", bridge.getId(), bridge.getCustomerId());
         return bridge;
     }
+
+    @Override
+    public BridgeDTO toDTO(Bridge bridge) {
```
robpblake, line 124: @kostola Same point here. These are inheriting the `@Transactional` from the class-level annotation. There is no need for a transaction on these methods, so could you please remove the class-level annotation and place `@Transactional` only on methods where it is required.

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorService.java
```diff
@@ -23,4 +26,12 @@
     ListResult<Processor> getProcessors(String bridgeId, String customerId, int page, int size);
 
     void deleteProcessor(String bridgeId, String processorId, String customerId);
+
+    ProcessorDTO toDTO(Processor processor);
+
+    ProcessorResponse toResponse(Processor processor);
+
+    JsonNode definitionToJsonNode(ProcessorDefinition definition);
```
robpblake, line 34: Do `definitionToJsonNode` and `jsonNodetoDefinition` need to be declared on the interface? AFAICS you only use them internally to the implementation  of the `ProcessorServiceImpl`.

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
kostola: Done :+1:

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
kostola: Done :+1:

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorService.java
kostola: They were used also in a test but I managed to make it work without the need to add these methods to the interface as well :+1:

robpblake (approved):

PR #95, opened 2021-10-27
MGDOBR-103: deploy BridgeIngress application in BridgeIngressController loop

r00ta:
[MGDOBR-103](https://issues.redhat.com/browse/MGDOBR-103) 

Please make sure that your PR meets the following requirements:

- [x] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/main/kogito-build/kogito-ide-config)
- [x] Your commit messages are clear and reference the JIRA issue e.g: "[MGDOBR-1] - $clear_explanation_of_what_you_did"
- [x] All new functionality is tested and uses org.assertj.core.api.Assertions for Assertions
- [x] Pull Request title is properly formatted: `MGDOBR-XYZ Subject`
- [x] Pull Request contains link to the JIRA issue
- [x] Pull Request contains link to any dependent or related Pull Request
- [x] Pull Request contains description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket

PR #101, opened 2021-11-03
MGDOBR-106: Deploy service for a given BridgeIngress Deployment resource

r00ta:
[MGDOBR-106](https://issues.redhat.com/browse/MGDOBR-106) 

Once the BridgeIngress deployment is ready, a Service has to be deployed for such resource.

Depends on https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/95


shard-operator/src/main/kubernetes/sample.yml
```diff
@@ -1,11 +1,11 @@
 apiVersion: com.redhat.service.bridge/v1alpha1
 kind: BridgeIngress
 metadata:
-  name: 99259d8b-3fdd-48de-a810-a66db8b9914c
+  name: my-bridge-ingress
```
r00ta, line 4: Since we do create a `deployment` and a `service` with the same `BridgeIngress.metadata.name`, it must be a valid name for such resources. `99259d8b..` is not a valid Service name. 

Revisit this logic https://issues.redhat.com/browse/MGDOBR-116

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/BridgeIngressServiceImpl.java
```diff
@@ -79,6 +67,42 @@ public Deployment fetchOrCreateBridgeIngressDeployment(BridgeIngress bridgeIngre
         return kubernetesClient.apps().deployments().inNamespace(bridgeIngress.getMetadata().getNamespace()).create(deployment);
     }
 
+    @Override
+    public Service fetchOrCreateBridgeIngressService(BridgeIngress bridgeIngress, Deployment deployment) {
+        Service service = kubernetesClient.services().inNamespace(bridgeIngress.getMetadata().getNamespace()).withName(bridgeIngress.getMetadata().getName()).get();
+
+        if (service != null) {
+            return service;
+        }
```
danielezonca, line 76: Is this thread safe? is it possible that two concurrent requests will produce two instances of the service?
If the second call is going to fail we might need to add a try/catch at [line 85](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/101/files#diff-93fb913d2b2f15cb9a101f0d12c29a300b6dbe266ba067417c23da2ef69189d2R85) and in that case execute again the call to load the existing service.
Wdyt?
Btw what about at least add a log information if `service == null` to log the fact that a new service instance is going to be created? This can help to debug similar issues

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/controllers/BridgeIngressController.java
```diff
@@ -67,6 +69,21 @@ public void init(EventSourceManager eventSourceManager) {
...
+
+        // Create Service
+        Service service = bridgeIngressService.fetchOrCreateBridgeIngressService(bridgeIngress, deployment);
+        if (service.getStatus() == null) {
+            LOGGER.info("Ingress service BridgeIngress: '{}' in namespace '{}' is NOT ready", bridgeIngress.getMetadata().getName(),
+                    bridgeIngress.getMetadata().getNamespace());
+            bridgeIngress.setStatus(new BridgeIngressStatus(PhaseType.AUGMENTATION));
+            return UpdateControl.updateStatusSubResource(bridgeIngress);
+        }
+        LOGGER.info("Ingress service BridgeIngress: '{}' in namespace '{}' is ready", bridgeIngress.getMetadata().getName(), bridgeIngress.getMetadata().getNamespace());
```
danielezonca, line 82: Are we sure that if `service.getStatus() != null` this implies that it is ready?

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/watchers/ServiceEvent.java
```diff
@@ -0,0 +1,50 @@
...
+import io.fabric8.kubernetes.client.Watcher;
+import io.javaoperatorsdk.operator.processing.event.DefaultEvent;
+
+public class ServiceEvent extends DefaultEvent {
+
+    private final Watcher.Action action;
+    private final Service service;
+
+    public ServiceEvent(Watcher.Action action, Service service, ServiceEventSource serviceEventSource) {
+        // TODO: this mapping is really critical and should be made more explicit by the java operator sdk
```
danielezonca, line 13: Given this comment about how critical is, do we have a test to make sure this is always done?

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/watchers/ServiceEventSource.java
```diff
@@ -0,0 +1,82 @@
...
+            return;
+        }
+
+        eventHandler.handleEvent(new ServiceEvent(action, service, this));
+    }
+
+    @Override
+    public void onClose(WatcherException e) {
+        if (e == null) {
+            return;
```
danielezonca, line 70: Is this happening in case of a graceful shutdown (aka without any error)?
What about adding a log?

shard-operator/src/test/java/com/redhat/service/bridge/shard/operator/BridgeIngressServiceTest.java
```diff
@@ -63,11 +82,29 @@ public void testBridgeIngressCreation() {
...
+        BridgeIngress bridgeIngress = kubernetesClient
+                .resources(BridgeIngress.class)
+                .inNamespace(customerNamespaceProvider.resolveName(dto.getCustomerId()))
+                .withName(KubernetesResourceUtil.sanitizeName(dto.getId()))
+                .get();
+        assertThat(bridgeIngress).isNull();
+    }
+
+    @Test
+    @Disabled("Delete loop in BridgeIngressController does not get called. Bug in the SDK?")
```
danielezonca, line 103: Unrelated to this PR but do we have a ticket to track this problem?

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/controllers/BridgeIngressController.java
r00ta: Yes, I am :) If anybody has a proof that this is not always true just let me know

r00ta: cc @sutaakar ^^ ?

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/BridgeIngressServiceImpl.java
r00ta: Even if this `fetchOrCreateBridgeIngressService` gets created concurrently it would not be a problem: the line 85 can't create two instances of the service because it is not possible to have two services with the same resource name (in the same namespace). 

Instead, I smell there could be a concurrency problem in the BridgeIngressController itself. But honestly I've never seen any concurrency management in any operator controller so far: let me double check with the java operator sdk team

shard-operator/src/test/java/com/redhat/service/bridge/shard/operator/BridgeIngressServiceTest.java
r00ta: Just to clarify: as far as I've seen this is a problem of the testing framework, because this works on a real kubernetes cluster. I'm going to create a ticket for this test

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/BridgeIngressServiceImpl.java
danielezonca: > Even if this `fetchOrCreateBridgeIngressService` gets called concurrently it would not be a problem: the line 85 can't create two instances of the service because it is not possible to have two services with the same resource name (in the same namespace).
> 
> Instead, I smell there could be a concurrency problem in the BridgeIngressController itself. But honestly I've never seen any concurrency management in any operator controller so far: let me double check with the java operator sdk team

Ok it is possible that the operator processes events sequentially so there is no concurrency problem 👍 
I'm asking because the code by itself is not preventing unexpected/undesired scenarios

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/controllers/BridgeIngressController.java
sutaakar: I guess so. I am not aware of any dedicated attribute or annotation which describes Service status.

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/BridgeIngressServiceImpl.java
r00ta: Checked with the java sdk team: the concurrency is managed by the java sdk so we don't have to worry about it.

r00ta: Checked with the java sdk team: the concurrency is managed by the sdk itself so we don't have to worry about it

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/watchers/ServiceEventSource.java
r00ta: In theory this method is called when the watcher closes due to an Exception. The method `close()` is invoked when the watcher is gracefully closed instead. 

I'll add a log message

danielezonca: Ok if this is the case, how it is possible that the watcher failed with an Exception but here there is no exception (aka `e == null`)?

shard-operator/src/test/java/com/redhat/service/bridge/shard/operator/BridgeIngressServiceTest.java
r00ta: added link to ticket

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/watchers/ServiceEvent.java
r00ta: This is already covered by our tests: we do have tests that when the BridgeController deploys resources (Deployments, Services) and such resources get modified, the reconcile loop of the controller gets called -> meaning that those resources are linked properly. i.e. we do have integration tests that cover this scenario, but not unit tests.

danielezonca (approved):
LGTM 👍

PR #107, opened 2021-11-05
[MGDOBR-82] Update all responses from the API to use Error

fjtirado:


manager/src/main/java/com/redhat/service/bridge/manager/api/EventBridgeManagerExceptionMapper.java
```diff
@@ -1,23 +1,41 @@
...
         LOGGER.error("Failure", e);
-        return Response.status(e.getStatusCode()).entity(e.getMessage()).build();
+        Optional<Error> error = errorsService.getError(e);
+        ResponseBuilder builder = Response.status(e.getStatusCode());
+        if (error.isPresent()) {
+            ErrorResponse errorResponse = ErrorResponse.from(error.get());
+            errorResponse.setReason(e.getMessage());
+            builder.entity(errorResponse);
+        } else {
+            builder.entity(e.getMessage());
```
r00ta, line 37: Since this should not happen, should we `log.warn` ?

r00ta (approved):

manager/src/main/java/com/redhat/service/bridge/manager/api/EventBridgeManagerExceptionMapper.java
fjtirado: yes, we should indeed, thanks

danielezonca (approved):

PR #119, opened 2021-11-12
[MGDOBR-143] Adding IN filter

fjtirado:
[https://issues.redhat.com/browse/MGDOBR-143) 
Besides adding ValuesIn class, I did some refactor, trying to keep what I though was the original idea behind BaseFilter hierarchy.  
Some rationale that might or not be shared below ;):
-  getValuesAsString method is not really needed and, if was so, I think building the string represantion belongs to filter generation, not to BaseFilter hierarchy. 
- A generic was added to BaseFilter (which requires casting to subclass, but this was already assumed by checking the class type before) so now there is a cast class but not a value cast (which I believe is part of the logic inherent to BaseFilter hierarchy, to hold the proper value data type in the child filter)   
- ValuesIn (I name it following the type-operator convention of other BaseFilter classes), accepts a list of objects.
- ObjectMapper instance was moved to a static class. This instance is still used for builidng the internal object representation from the value string passed in the constructor (Im not sure we should pass a string, but I agree is useful for testing, so I kept that)

### DIFF
--- a/executor/src/main/java/com/redhat/service/bridge/executor/filters/FilterEvaluatorFactoryFEEL.java
+++ b/executor/src/main/java/com/redhat/service/bridge/executor/filters/FilterEvaluatorFactoryFEEL.java
@@ -5,10 +5,14 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.redhat.service.bridge.infra.models.filters.BaseFilter;
+import com.redhat.service.bridge.infra.models.filters.ObjectMapperFactory;
 import com.redhat.service.bridge.infra.models.filters.StringBeginsWith;
 import com.redhat.service.bridge.infra.models.filters.StringContains;
 import com.redhat.service.bridge.infra.models.filters.StringEquals;
+import com.redhat.service.bridge.infra.models.filters.ValuesIn;
 
 public class FilterEvaluatorFactoryFEEL implements FilterEvaluatorFactory {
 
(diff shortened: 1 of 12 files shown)


r00ta (commented):
Hey @fjtirado , I tried the following new test but it's red 

```java
    @Test
    public void testInFilter1() {
        FilterEvaluator evaluator = TEMPLATE_FACTORY_FEEL.build(Collections.singleton(new ValuesIn("source", "[\"Service\", \"1\", 2]")));

        assertThat(evaluator.evaluateFilters(Collections.singletonMap("source", "Service"))).isTrue();
        assertThat(evaluator.evaluateFilters(Collections.singletonMap("source", "1"))).isTrue();
        
        // It does not FAIL, but I think it should
        assertThat(evaluator.evaluateFilters(Collections.singletonMap("source", "2"))).isTrue();

        // FAIL
        assertThat(evaluator.evaluateFilters(Collections.singletonMap("source", 2))).isTrue();
    }
```
Is my assumption wrong?

fjtirado:
> Is my assumption wrong?

Thats clearly wrong, but Im now wondering if   assertThat(evaluator.evaluateFilters(Collections.singletonMap("source", 2))).isTrue(); is valid then should not be   assertThat(evaluator.evaluateFilters(Collections.singletonMap("source", "2"))).isFalse

r00ta:
> > Is my assumption wrong?
> 
> Thats clearly wrong, but Im now wondering if   assertThat(evaluator.evaluateFilters(Collections.singletonMap("source", 2))).isTrue(); is valid then should not be   assertThat(evaluator.evaluateFilters(Collections.singletonMap("source", "2"))).isFalse
> 
> I mean, since Feel expects a string, we might pass the number (as it is not done now) or the string (as it is done now). I vote for passing the original value, in this case the number
> 

Yes my main point is on the user expectation. Meaning that if the user creates a filter for `"2"`, then  `2` should not match and viceversa. Wdyt?

fjtirado:
> Yes my main point is on the user expectation. Meaning that if the user creates a filter for `"2"`, then `2` should not match and viceversa. Wdyt?
I think that if user put 2, it should be equal to 2 and "2" should be equals to "2" (this is a great sentence to add to a science computer book). I have updated the code and unit unit test accordingly

executor/src/main/java/com/redhat/service/bridge/executor/filters/FilterEvaluatorFactoryFEEL.java
```diff
@@ -22,28 +26,35 @@ public FilterEvaluator build(Set<BaseFilter> filters) {
...
-        for (String value : (List<String>) filter.getValue()) {
-            conditions.add(String.format(singleFormatTemplate, filter.getKey(), value));
+        ObjectMapper objectMapper = ObjectMapperFactory.get();
+        for (T value : filter.getValue()) {
+            try {
+                conditions.add(String.format(singleFormatTemplate, filter.getKey(), objectMapper.writeValueAsString(value)));
+            } catch (JsonProcessingException e) {
+                // TODO Auto-generated catch block
+                e.printStackTrace();
+            }
```
r00ta, line 57: Log? 
Just before I forget about it, I have a more general consideration: in the future we will have to notify the user about runtime issues like this one. If you agree on this point, may I ask you to create an epic like "runtime exception monitoring/notification for user"? I will be complex feature to design and maybe to implement, but it's something we might have to add in the future 
cc @robpblake @danielezonca

executor/src/test/java/com/redhat/service/bridge/executor/filters/FilterEvaluatorFEELTest.java
```diff
@@ -49,6 +50,16 @@ public void testStringContainsListFilter() {
...
     }
 
+    @Test
+    public void testInFilter() {
+        FilterEvaluator evaluator = TEMPLATE_FACTORY_FEEL.build(Collections.singleton(new ValuesIn("source", "[\"Service\", \"Testing\", 2]")));
+
+        assertThat(evaluator.evaluateFilters(Collections.singletonMap("source", "Service"))).isTrue();
+        assertThat(evaluator.evaluateFilters(Collections.singletonMap("source", "Testing"))).isTrue();
+        assertThat(evaluator.evaluateFilters(Collections.singletonMap("source", 2))).isTrue();
+        assertThat(evaluator.evaluateFilters(Collections.singletonMap("source", "notMyApplication"))).isFalse();
```
r00ta, line 60: What about adding also

```java
        assertThat(evaluator.evaluateFilters(Collections.singletonMap("source", "2"))).isFalse();
```
?

r00ta (approved):
Nice! Thank you @fjtirado ! 
We have also a small documentation for filters here https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/FILTERS.md . Feel free to add a small paragraph about this new filter, otherwise we'll do later

executor/src/main/java/com/redhat/service/bridge/executor/filters/FilterEvaluatorFactoryFEEL.java
fjtirado: In this case, this was an omission on my side, this should throw IllegalArgumentException (and I think we probably need an entry in our error catalog for this runtime, but lets discuss that)

fjtirado:
> Nice! Thank you @fjtirado ! We have also a small documentation for filters here https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/FILTERS.md . Feel free to add a small paragraph about this new filter, otherwise we'll do later

Thanks a lot, documentation updated

executor/src/main/java/com/redhat/service/bridge/executor/filters/FilterEvaluatorFactoryFEEL.java
danielezonca: We have a ticket about defining error strategy and this is definitely a good use case for that

r00ta: Hey @fjtirado , the problem is that this error is thrown in the `executor` application, so the user does not have any visibility on this application yet. It's a kind of internal service that the user can not see

robpblake: @fjtirado @r00ta @danielezonca I'm planning on covering this error scenario as part of the work for https://issues.redhat.com/browse/MGDOBR-145 this sprint 👍

PR #147, opened 2021-11-25
MGDOBR-147: Create basic test suite for Shard operator integration tests

sutaakar:

[MGDOBR-147](https://issues.redhat.com/browse/MGDOBR-147) 

Initial implementation of Shard operator integration tests.
The tests focus specifically on operator functionality, the BridgeIngress images were replaced with dummy ones as original images require Kafka - is currently not available for CI tests.

@r00ta  Currently integration tests aren't added to parent module list as the integration tests are executed by default there. Running this module without prepared environment will cause test failures. Would it have more sense to add the module into module list, allowing tests to be executed only when parameter/profile is activated?

### DIFF
--- a/shard-operator-integration-tests/src/test/java/com/redhat/service/bridge/shard/operator/cucumber/common/TimeUtils.java
+++ b/shard-operator-integration-tests/src/test/java/com/redhat/service/bridge/shard/operator/cucumber/common/TimeUtils.java
@@ -1,0 +1,38 @@
+package com.redhat.service.bridge.shard.operator.cucumber.common;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.BooleanSupplier;
+
+public final class TimeUtils {
+
+    private static final Duration DEFAULT_WAIT_STEP = Duration.of(1, ChronoUnit.SECONDS);
+
+    private TimeUtils() {
+    }
+
+    public static void wait(Duration duration) {
+        try {
+            Thread.sleep(duration.toMillis());
+        } catch (InterruptedException e) {
+            throw new RuntimeException("Waiting was interrupted", e);
+        }
+    }
+
+    public static void waitForCondition(Duration maxTime, BooleanSupplier conditionSupplier, String timeoutMessage) throws TimeoutException {
+        waitForCondition(maxTime, DEFAULT_WAIT_STEP, conditionSupplier, timeoutMessage);
+    }
+
+    public static void waitForCondition(Duration maxDuration, Duration waitStep, BooleanSupplier conditionSupplier, String timeoutMessage) throws TimeoutException {
+        Instant startTime = Instant.now();
+        while (startTime.plus(maxDuration).isAfter(Instant.now()) && !conditionSupplier.getAsBoolean()) {
+            wait(waitStep);
+        }
+
+        if (!conditionSupplier.getAsBoolean()) {
+            throw new TimeoutException(timeoutMessage);
+        }
+    }
+}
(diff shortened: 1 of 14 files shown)


shard-operator-integration-tests/src/test/java/com/redhat/service/bridge/shard/operator/cucumber/steps/BridgeIngressSteps.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private Context context;
+
+    public BridgeIngressSteps(Context context) {
+        this.context = context;
+    }
+
+    @When("^deploy BridgeIngress:$")
+    public void deployBridgeIngress(String bridgeIngressYaml) {
+        InputStream resourceStream = new ByteArrayInputStream(bridgeIngressYaml.getBytes(StandardCharsets.UTF_8));
```
ricardozanini, line 29: We must probably have unique `TemplateProvider` factories for those. This way we can easily change how we read these templates. We are looking for using Qute in the near future.

shard-operator-integration-tests/src/test/java/com/redhat/service/bridge/shard/operator/cucumber/common/TimeUtils.java
```diff
@@ -0,0 +1,35 @@
+package com.redhat.service.bridge.shard.operator.cucumber.common;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.BooleanSupplier;
+
+public class TimeUtils {
```
ricardozanini, line 9: Usually `Utils` are final classes with private constructors.

shard-operator-integration-tests/src/test/resources/features/bridgeingress-deploy.feature
```diff
@@ -0,0 +1,50 @@
...
+  Scenario: BridgeIngress is in phase AVAILABLE
+    When deploy BridgeIngress:
+    """
+    apiVersion: com.redhat.service.bridge/v1alpha1
+    kind: BridgeIngress
+    metadata:
+      name: my-bridge-ingress
+      app.kubernetes.io/managed-by: bridge-fleet-shard-operator
+    spec:
+      image: nginx:1.14.2
```
ricardozanini, line 16: That's interesting. I think we might need to think in a way to enforce a "real" image for our custom resources. 🤔

ricardozanini (approved):
Pretty nice! 👍

shard-operator-integration-tests/src/test/java/com/redhat/service/bridge/shard/operator/cucumber/steps/BridgeIngressSteps.java
sutaakar: Could be used to share same image for all BridgeIngress instances for tests using proper image.

shard-operator-integration-tests/src/test/resources/features/bridgeingress-deploy.feature
sutaakar: My plan is to have two levels of tests.
First one is Shard operator tests (integration) checking interactions between operator and Kubernetes environment. These should be ideally quick tests with scope limited to operator only, possibly mocking unnecessary dependencies. Assuming that the dependencies themselves are tested in their dedicated tests.
Second level are e2e tests (user acceptance) testing whole solution. These tests will be significantly slower as they will require full environment with all dependencies. The tests there will be written from user perspective (checking outputs visible to user).

ricardozanini (approved):

r00ta:
Hey @sutaakar 

> @r00ta Currently integration tests aren't added to parent module list as the integration tests are executed by default there. Running this module without prepared environment will cause test failures. Would it have more sense to add the module into module list, allowing tests to be executed only when parameter/profile is activated?

Yes I think it would be good to add it to the parent module and configure the maven surfire plugin to run those tests only if a property/profile is specified.

sutaakar:
@r00ta Done, tests are enabled using profile `cucumber`.
I have switched test execution to Failsafe plugin to align maven phase as the tests are integration tests.

r00ta (approved):
Looks great! Nice work @sutaakar !

PR #141, opened 2021-11-24
MGDOBR-134: implement SendToBridge action

kostola:
[MGDOBR-134](https://issues.redhat.com/browse/MGDOBR-134) 

Please make sure that your PR meets the following requirements:

- [x] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/main/kogito-build/kogito-ide-config)
- [x] Your commit messages are clear and reference the JIRA issue e.g: "[MGDOBR-1] - $clear_explanation_of_what_you_did"
- [x] All new functionality is tested
- [x] Pull Request title is properly formatted: `MGDOBR-XYZ Subject`
- [x] Pull Request contains link to the JIRA issue
- [x] Pull Request contains link to any dependent or related Pull Request
- [x] Pull Request contains description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket

PR #152, opened 2021-11-30
MGDOBR-134: fix bridge webhook in SendToBridge action

kostola:
[MGDOBR-134](https://issues.redhat.com/browse/MGDOBR-134) 

### DIFF
--- a/manager/src/main/java/com/redhat/service/bridge/manager/actions/sendtobridge/SendToBridgeActionTransformer.java
+++ b/manager/src/main/java/com/redhat/service/bridge/manager/actions/sendtobridge/SendToBridgeActionTransformer.java
@@ -8,9 +8,6 @@ import java.util.Map;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import com.redhat.service.bridge.actions.ActionProviderException;
 import com.redhat.service.bridge.actions.ActionTransformer;
 import com.redhat.service.bridge.actions.webhook.WebhookAction;
@@ -20,8 +17,6 @@ import com.redhat.service.bridge.manager.models.Bridge;
 
 @ApplicationScoped
 public class SendToBridgeActionTransformer implements ActionTransformer {
-
-    private static final Logger LOG = LoggerFactory.getLogger(SendToBridgeActionTransformer.class);
 
     @Inject
     BridgesService bridgesService;
@@ -36,7 +31,6 @@ public class SendToBridgeActionTransformer implements ActionTransformer {
         try {
             parameters.put(WebhookAction.ENDPOINT_PARAM, getBridgeWebhookUrl(destinationBridge.getEndpoint()));
         } catch (MalformedURLException e) {
-            LOG.error("Error when translating bridge endpoint to webhook URL", e);
             throw new ActionProviderException("Can't find events webhook for bridge " + destinationBridgeId);
         }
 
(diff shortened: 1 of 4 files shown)


r00ta (approved):

manager/src/main/java/com/redhat/service/bridge/manager/actions/sendtobridge/SendToBridgeActionTransformer.java
```diff
@@ -24,7 +32,13 @@ public BaseAction transform(BaseAction action, String bridgeId, String customerI
         Bridge destinationBridge = bridgesService.getAvailableBridge(destinationBridgeId, customerId);
 
         Map<String, String> parameters = new HashMap<>();
-        parameters.put(WebhookAction.ENDPOINT_PARAM, destinationBridge.getEndpoint());
+
+        try {
+            parameters.put(WebhookAction.ENDPOINT_PARAM, getBridgeWebhookUrl(destinationBridge.getEndpoint()));
+        } catch (MalformedURLException e) {
+            LOG.error("Error when translating bridge endpoint to webhook URL", e);
```
robpblake, line 39: @kostola Does this result in duplicate stacktrace printed in the log? We have `error` level logging in the `EventBridgeManagerExceptionMapper` which should print all failures.

manager/src/main/java/com/redhat/service/bridge/manager/actions/sendtobridge/SendToBridgeActionTransformer.java
```diff
@@ -33,4 +47,9 @@ public BaseAction transform(BaseAction action, String bridgeId, String customerI
 
         return transformedAction;
     }
+
+    static String getBridgeWebhookUrl(String bridgeEndpoint) throws MalformedURLException {
```
robpblake, line 51: Can we please move this logic to the `Bridge` class for now? So something like:

```
getIngressUrl() // returns endpoint + /events
getPlainIngressUrl() // returns endpoint + /events/plain
```

My concern is that we're hiding  information that should be available via our API. By having to do this here, we're basically saying that our API doesn't currently expose enough information to clients for them to be able to interact with OpenBridge without further information e.g. they need to know that `/events` is the correct suffix to use.

I'd like us to consider updating the API to expose both endpoints so that our API is fully navigable. That's beyond the scope of this PR, but I think centralising the logic to get the endpoint onto the `Bridge` is a good first step.
I'll create an issue in our backlog to

robpblake (changes requested):
@kostola Small change requested please. Cheers.

manager/src/main/java/com/redhat/service/bridge/manager/actions/sendtobridge/SendToBridgeActionTransformer.java
kostola: What about creating a dedicated task for these changes? This PR is meant to be a fix for [the initial MGDOBR-134 PR](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/141) and I'd like to fix it according to the current underlying implementation (in which the API returns the base endpoint for the bridge).

manager/src/main/java/com/redhat/service/bridge/manager/actions/sendtobridge/SendToBridgeActionTransformer.java
kostola: You're right. I removed it :+1:

manager/src/main/java/com/redhat/service/bridge/manager/actions/sendtobridge/SendToBridgeActionTransformer.java
robpblake: That's fair @kostola . FYI, issue raised as https://issues.redhat.com/browse/MGDOBR-178

robpblake (approved):

PR #162, opened 2021-12-02
MGDOBR-140: check if expected resources are deployed, redeploy eventually

r00ta:
[MGDOBR-140](https://issues.redhat.com/browse/MGDOBR-140) 

This PR aims to implement a patter like for the resources controlled by the BridgeIngress and BridgeExecutor controllers. 

```java
T expected = buildExpectedResource()
T existing = retrieveExisting()
if (existing == null | !expected.equals(existing)){
  createOrReplace(expected);
}
```


shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/BridgeIngressServiceImpl.java
```diff
@@ -80,26 +84,31 @@ public Deployment fetchOrCreateBridgeIngressDeployment(BridgeIngress bridgeIngre
...
-        }
+        // Specs
+        expected.getSpec().setSelector(new LabelsBuilder().withAppInstance(deployment.getMetadata().getName()).build());
 
-        service = templateProvider.loadBridgeIngressServiceTemplate(bridgeIngress);
+        Service existing = kubernetesClient.services().inNamespace(bridgeIngress.getMetadata().getNamespace()).withName(bridgeIngress.getMetadata().getName()).get();
 
-        // Specs
-        service.getSpec().setSelector(new LabelsBuilder().withAppInstance(deployment.getMetadata().getName()).build());
+        if (existing == null || !expected.getSpec().equals(existing.getSpec())) {
```
sutaakar, line 107: Minor:
Service spec contains clusterIP, which is added by Kubernetes/OpenShift. Therefore equals will always be false here. Though it doesn't seem to cause any issue apart from unnecessary client call.

sutaakar (approved):

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/BridgeIngressServiceImpl.java
```diff
@@ -80,26 +84,31 @@ public Deployment fetchOrCreateBridgeIngressDeployment(BridgeIngress bridgeIngre
...
+        expected.getSpec().getTemplate().getSpec().getContainers().get(0).setEnv(environmentVariables);
+
+        Deployment existing = kubernetesClient.apps().deployments().inNamespace(bridgeIngress.getMetadata().getNamespace()).withName(bridgeIngress.getMetadata().getName()).get();
 
-        return kubernetesClient.apps().deployments().inNamespace(bridgeIngress.getMetadata().getNamespace()).create(deployment);
+        if (existing == null || !DeploymentSpecUtils.isDeploymentEqual(expected, existing)) {
+            return kubernetesClient.apps().deployments().inNamespace(bridgeIngress.getMetadata().getNamespace()).createOrReplace(expected);
+        }
+
+        return existing;
```
ricardozanini, line 95: The code in both services seems pretty much the same for me. Maybe we could refactor to a Kubernetes Resource Builder pattern? Instead of having `BridgeIngressService` and `BridgeExecutorService`, we would have `ShardServiceBuilder`, `ShardDeploymentBuilder`, and so on? wdyt?

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/BridgeIngressServiceImpl.java
ricardozanini: The most incredible thing about Java `equals` is that we can control what to compare. In this case, we are only comparing the attributes that Service cares about. I haven't seen the code yet, but my guess would be that.

ricardozanini (approved):

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/BridgeIngressServiceImpl.java
r00ta: Yeah, when I implemented this I checked the fabric8 override for `equals` and they are checking (recursively) that the object is of course exactly the same. The problems are the properties set by kubernetes... I had to implement a quite specific `equals` check for the `Deployment` few line below for this reason. 
I will move to a more specific check also for the `Service`, thanks for the catch :)

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/BridgeIngressServiceImpl.java
r00ta: Hi @ricardozanini , the BridgeIngress and BridgeExecutor deployments are different. Some specific labels are set in the template itself (for example the `component` property that we use to filter the resources in the controllers), and some others are set by the `BridgeIngressService`/`BridgeExecutorService`, for example the environment variables are different. 
The `ShardServiceBuilder` and the `ShardDeploymentBuilder` can't really be generic, since the labels and some properties like the env variables need to be specific for the executor and the ingress, so I'm not sure it really improves the code (since we need to set them somewhere, probably again in a kind of `BridgeIngressService`). 
If it's fine for you I suggest we move forward with this PR and better discuss if we can improve the code in another ticket, is that ok for you @ricardozanini ?

ricardozanini: That's fine, this comment is not a blocker. We can discuss this further in another thread. But labels and env vars can be injected or used as a source to this new service. They are also pretty generic (literally a map). I think we should consider having the k8s resource creation centralized somehow. They both will share labels, envs, probes, and so on.

r00ta:
Hey @ricardozanini @sutaakar , I've adjusted a couple of things. Could you have a final look?

sutaakar:
@r00ta Can you please rebase?

r00ta:
> @r00ta Can you please rebase?

Hi @sutaakar , it's already rebased

sutaakar:
GitHub complains with `This branch cannot be rebased due to conflicts`, see below

r00ta:
That's strange @sutaakar 
Screenshot from 2021-12-15 16-44-45

This is what I see

sutaakar:
Weird:
image
Anywal, will check the code again.

sutaakar (approved):
Looks good to me

ricardozanini (approved):

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/monitoring/ServiceMonitorServiceImpl.java
```diff
@@ -31,35 +31,31 @@
...
-            serviceMonitor = ServiceMonitorClient
-                    .get(kubernetesClient)
-                    .inNamespace(resource.getMetadata().getNamespace())
-                    .withName(service.getMetadata().getName())
-                    .create(serviceMonitor);
-        } else {
-            this.ensureLabels(serviceMonitor, service, componentName);
-            serviceMonitor = ServiceMonitorClient
+
+        if (existing == null || !expected.getSpec().equals(existing.getSpec())) {
```
r00ta, line 50: Hey @sutaakar , I've just rebased after merging Ricardo's PR for service monitoring. I checked and it seems that the ServiceMonitor spec is not changed by the operator so I can use `!expected.getSpec().equals(existing.getSpec())`. Could you double check? Thank you very much!

sutaakar: Yes, works as expected.

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/monitoring/ServiceMonitorServiceImpl.java
```diff
@@ -31,35 +31,31 @@
     @Override
     public Optional<ServiceMonitor> fetchOrCreateServiceMonitor(final CustomResource resource, final Service service, final String componentName) {
         if (!ServiceMonitorClient.isServiceMonitorAvailable(kubernetesClient)) {
-            LOGGER.debug("Prometheus Operator is not available in this cluster");
+            LOGGER.warn("Prometheus Operator is not available in this cluster");
```
ricardozanini, line 34: Please revert to DEBUG since the controllers are already warning a similar message.

ricardozanini (changes requested):

PR #163, opened 2021-12-02
MGDOBR-141 MC: Let OpenBridge call the Managed Connector to create the Connector

lucamolteni:
Missing: update of the connector status (will be addressed by [MGDOBR-142](https://issues.redhat.com/browse/MGDOBR-142))

Some suggestions on error checking are welcome. 

[MGDOBR-141](https://issues.redhat.com/browse/MGDOBR-141) 

- Created Connectors Action transformation in fleet manager
- Persistence of Connectors in the DB
- Test MCAction to KafkaTopic action transformation
- Added Configuration for kafka in the fleet manager as well
- Added MC cluster id in configuration
- Connectors REST Api is Proxy for easier mocking
- Added dependency to Connectors REST Java SDK
- Use offline Token for Connector auth


manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorAuthImpl.java
```diff
@@ -0,0 +1,94 @@
...
+
+    String bearerToken;
+
+    // TODO-MC this should be cached for five minutes
+    @Override
+    public String bearerToken() {
+        if (bearerToken == null) {
+            bearerToken = bearerTokenFromOfflineToken(offlineToken);
+        }
+        return null;
```
sutaakar, line 41: Shouldn't `bearerToken` be returned here?

manager/src/main/resources/application.properties
```diff
@@ -45,3 +45,22 @@ quarkus.native.resources.includes=exception/exceptionInfo.json
 %dev.quarkus.iam.aws.credentials.type=static
 %dev.quarkus.iam.aws.credentials.static-provider.access-key-id=testing
 %dev.quarkus.iam.aws.credentials.static-provider.secret-access-key=testing
+
+
+# The Kafka broker location (defaults to localhost:9092)
+kafka.bootstrap.servers=${KAFKA_BOOTSTRAP_SERVERS:localhost:9092}
```
danielezonca, line 51: This is wrong in general: we should use MK API (similar to what we do for now with MC) to get Kafka information.
In the future we might have also multiple instances and the manager decide where to deploy
Do we have a similar ticket?

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsServiceImpl.java
```diff
@@ -0,0 +1,176 @@
...
+        Map<String, String> actionParameters = action.getParameters();
+        String connectorType = actionParameters.get("connectorType");
+        String newConnectorName = actionParameters.get("connectorName");
+
+        JsonNode connectorPayload = parseConnectorPayloadString(actionParameters);
+
+        // Assume connectors actions are transformed to send-to-kafka
+        String kafkaTopicName = resolvedAction.getParameters().get(KafkaTopicAction.TOPIC_PARAM);
+        if (kafkaTopicName == null) {
+            throw new RuntimeException("Need a kafka topic to create a connector");
```
danielezonca, line 83: Same here about the exception

danielezonca (commented):
Thanks for the PR, well done. Some comments

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorAuthImpl.java
```diff
@@ -0,0 +1,94 @@
...
+    private String bearerTokenFromOfflineToken(String offlineToken) {
+        try {
+            HttpRequest request =
+                    HttpRequest.newBuilder().uri(URI.create(authServerUrl + "/" + tokenPath))
+                            .header("content-type", "application/x-www-form-urlencoded")
+                            .timeout(Duration.ofMinutes(2)).POST(ofFormData("grant_type", "refresh_token",
+                                    "client_id", clientId, "refresh_token", offlineToken))
+                            .build();
+
+            HttpClient client = HttpClient.newBuilder().build();
```
robpblake, line 60: Just wondering if it makes sense to rely on the VertX HTTP client here? We already use that HTTP Client in the Shard, so it would be good to keep our approach to HTTP requests consistent across the project.

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsApiImpl.java
```diff
@@ -0,0 +1,44 @@
...
+import com.openshift.cloud.api.connector.invoker.ApiClient;
+import com.openshift.cloud.api.connector.invoker.ApiException;
+import com.openshift.cloud.api.connector.invoker.Configuration;
+import com.openshift.cloud.api.connector.invoker.auth.HttpBearerAuth;
+import com.openshift.cloud.api.connector.models.Connector;
+
+@RequestScoped
+public class ConnectorsApiImpl implements ConnectorsApi {
+
+    @ConfigProperty(name = "mc.services.url")
```
robpblake, line 17: Same here for the property name `managed-connectors.service.url`

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsApiImpl.java
```diff
@@ -0,0 +1,44 @@
...
+        try {
+            return connectorsAPI.createConnector(true, connector);
+        } catch (ApiException e) {
+            // TODO-MC error handling
+            throw new RuntimeException(e);
+        }
+
+    }
+
+    private com.openshift.cloud.api.connector.ConnectorsApi createConnectorsAPI() {
```
robpblake, line 35: Can this be `@Inject`? Feels like it would be better to try and create an instance of this bean at start-up and then fail if we cannot.

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsServiceImpl.java
```diff
@@ -0,0 +1,176 @@
...
+        kafka.put("topic", kafkaTopic);
+    }
+
+    private void persistConnector(Processor processor, String newConnectorName, JsonNode connectorPayload) {
+        ConnectorEntity newConnectorEntity = new ConnectorEntity();
+
+        newConnectorEntity.setName(newConnectorName);
+        newConnectorEntity.setStatus(ConnectorStatus.REQUESTED);
+        newConnectorEntity.setSubmittedAt(ZonedDateTime.now());
+        newConnectorEntity.setPublishedAt(ZonedDateTime.now());
```
robpblake, line 111: I assume this should be set once the Connector is fully deployed?

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsServiceImpl.java
```diff
@@ -0,0 +1,176 @@
...
+    String kafkaBootstrapServer;
+
+    @ConfigProperty(name = "kafka.client.id")
+    String serviceAccountId;
+
+    @ConfigProperty(name = "kafka.client.secret")
+    String serviceAccountSecret;
+
+    @Override
+    @Transactional(Transactional.TxType.MANDATORY)
```
robpblake, line 63: I think the default `REQUIRED` TxType is fine here.

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsServiceImpl.java
```diff
@@ -0,0 +1,176 @@
...
+        createConnectorRequest.setConnectorSpec(connectorPayload);
+
+        KafkaConnectionSettings kafka = new KafkaConnectionSettings();
+        kafka.setBootstrapServer(kafkaBootstrapServer);
+        kafka.setClientId(serviceAccountId);
+        kafka.setClientSecret(serviceAccountSecret);
+        createConnectorRequest.setKafka(kafka);
+
+        try {
+            Connector connectorResult = connectorsApi.createConnector(createConnectorRequest);
```
robpblake, line 153: @lucamolteni I assume this call is async and that we have to poll for the status of the Connector after this initial request? The thing we need to be careful of here is that we are invoking a remote system whilst potentially holding DB transaction locks on our tables (you're in a method annotated `@Transactional`). If for whatever reason MC is unresponsive then we need to make sure we don't hold the transaction open for a long time.

manager/src/main/java/com/redhat/service/bridge/manager/dao/ConnectorDAO.java
```diff
@@ -0,0 +1,32 @@
...
+import io.quarkus.hibernate.orm.panache.PanacheQuery;
+import io.quarkus.hibernate.orm.panache.PanacheRepositoryBase;
+import io.quarkus.panache.common.Parameters;
+
+@ApplicationScoped
+@Transactional
+public class ConnectorDAO implements PanacheRepositoryBase<ConnectorEntity, String> {
+
+    private static final String IDS_PARAM = "ids";
+
```
robpblake, line 19: `findByProcessorIdAndName()` ?

manager/src/main/java/com/redhat/service/bridge/manager/models/ConnectorEntity.java
```diff
@@ -0,0 +1,155 @@
...
+
+import io.quarkiverse.hibernate.types.json.JsonBinaryType;
+import io.quarkiverse.hibernate.types.json.JsonTypes;
+
+@NamedQueries({
+        @NamedQuery(name = "CONNECTORENTITY.findByProcessorIdAndName",
+                query = "from ConnectorEntity c where c.name=:name and c.processor.id=:processorId"),
+})
+@Entity
+@Table(name = "CONNECTOR_ENTITY")
```
robpblake, line 34: I think the DB table can be just call `CONNECTOR`. WDYT?

infra/src/main/java/com/redhat/service/bridge/infra/models/dto/ConnectorStatus.java
```diff
@@ -0,0 +1,11 @@
+package com.redhat.service.bridge.infra.models.dto;
+
+// TODO-MC review statuses
```
robpblake, line 3: @lucamolteni Is there a reason why this status is independent from `BridgeStatus`? They look the same to me, so can they be combined (assuming we need to maintain our own status as per comment of @danielezonca)

robpblake (commented):
Some initial comments from me @lucamolteni . This is a good start 👍

manager/src/main/java/com/redhat/service/bridge/manager/dao/ConnectorDAO.java
```diff
@@ -0,0 +1,32 @@
...
+    public ConnectorEntity findByProcessIdName(String processorId, String name) {
```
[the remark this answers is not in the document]
robpblake: Let's be consistent. All our other DAOs currently return `null` if the requested entity doesn't exist.

manager/src/main/java/com/redhat/service/bridge/manager/models/ConnectorEntity.java
lucamolteni: It was renamed `CONNECTORENTITY` to avoid clash with the SDK. Do you mean to change only the table name and and keep the entity name `ConnectorEntity`?

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorAuthImpl.java
```diff
@@ -0,0 +1,94 @@
...
+import java.time.Duration;
+
+import javax.enterprise.context.RequestScoped;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import io.vertx.core.json.JsonObject;
+
+@RequestScoped
+public class ConnectorAuthImpl implements ConnectorAuth {
```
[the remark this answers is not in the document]
lucamolteni: @r00ta I didn't know such client existed, if it works I'm glad to replace it. WDYT @robpblake ?

manager/src/main/java/com/redhat/service/bridge/manager/models/ConnectorEntity.java
robpblake: Yes. That was my suggestion, but feel free to tell me it's a bad one :-)

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorAuthImpl.java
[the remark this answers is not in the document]
robpblake: @lucamolteni  @r00ta Yep. It's a good suggestion. +1 from me

infra/src/main/java/com/redhat/service/bridge/infra/models/dto/ConnectorStatus.java
lucamolteni: @robpblake I assumed MC will eventually have a different life cycles than our bridge instances, anyway when I review those if I discover they're identical we might consider using only one

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsServiceImpl.java
lucamolteni: I used `MANDATORY` to avoid calling the `ConnectorService` without an active transaction from a `Processor`. In fact, we never want a connector to be created without a processor or in a different transaction. 
This is already enforced by the FK but at an API level I find it more clear. 
WDYT? I have no strong feeling about this, to me it's fine to use `REQUIRED`

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsServiceImpl.java
lucamolteni: @robpblake It's async (managed-service style, not HTTP async style), so it should return pretty fast. 
I think that we set a timeout and it hits it, it'll throw an exception and the row won't be committed. 

Alternatively, we could commit the transaction and then call the WS outside it and if it fails we update the processor row with `ERROR` - that will surely be more robust. 

WDYT? Anyway I'm going to tackle those details in https://issues.redhat.com/browse/MGDOBR-142

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorAuthImpl.java
```diff
@@ -0,0 +1,94 @@
...
+
+            HttpClient client = HttpClient.newBuilder().build();
+            HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+            if (response.statusCode() == 200) {
+                String tokens = response.body();
+                JsonObject json = new JsonObject(tokens);
+                return json.getString("access_token");
+            } else {
+                String errorMessage = String.format("Http Error Code %d\n", response.statusCode()) + response.body();
+                throw new RuntimeException(errorMessage);
```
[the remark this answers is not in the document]
lucamolteni: As this authentication mechanism will be replaced by https://issues.redhat.com/browse/MGDOBR-196 I'll close this

manager/src/main/resources/application.properties
lucamolteni: https://issues.redhat.com/browse/MGDOBR-197

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsServiceImpl.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    private Connector callConnectorService(String connectorType,
+            JsonNode connectorPayload,
+            String newConnectorName) {
+        Connector createConnectorRequest = new Connector();
+
+        ConnectorAllOfMetadata metadata = new ConnectorAllOfMetadata();
+        metadata.setName(newConnectorName);
+        metadata.setKafkaId("kafkaId-ignored"); // this is currently ignored in the Connectors API
```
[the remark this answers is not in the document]
lucamolteni: https://issues.redhat.com/browse/MGDOBR-198

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorAuthImpl.java
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @param offlineToken the token from ss.redhat.com
+     * @return a token to be used as a bearer token to authorize the user
+     */
+    private String bearerTokenFromOfflineToken(String offlineToken) {
+        try {
+            HttpRequest request =
+                    HttpRequest.newBuilder().uri(URI.create(authServerUrl + "/" + tokenPath))
+                            .header("content-type", "application/x-www-form-urlencoded")
+                            .timeout(Duration.ofMinutes(2)).POST(ofFormData("grant_type", "refresh_token",
```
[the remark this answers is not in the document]
lucamolteni: Will be removed by https://issues.redhat.com/browse/MGDOBR-196

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorAuthImpl.java
lucamolteni: Will use the Quarkus OID https://issues.redhat.com/browse/MGDOBR-196

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsServiceImpl.java
lucamolteni: https://issues.redhat.com/browse/MGDOBR-142

lucamolteni:
Ready for review. Now we have a SlackAction instead of a ManagedConnector action

manager/src/main/java/com/redhat/service/bridge/manager/exceptions/CannotCreateConnectorException.java
```diff
@@ -0,0 +1,12 @@
+package com.redhat.service.bridge.manager.exceptions;
+
+public class CannotCreateConnectorException extends RuntimeException {
```
robpblake, line 3: I think this should extend `EventBridgeManagerException`

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsServiceImpl.java
robpblake: > avoid calling the ConnectorService without an active transaction from a Processor

I don't see this scenario in the code atm, unless I'm missing something?

My concern here is consistency (and explaining in comments when we deviate from the norm). For other people coming to this code from fresh without your knowledge: why is this method (and no others) annotated `MANDATORY`? Should our other `@Transactional` methods be updated to follow? When should they use `MANDATORY` or just rely on the defaults?

I think consistent and expected behaviour of our code should be one of our goals so that's why I'm +1 to rely on the defaults unless it really doesn't make sense, and when it doesn't make sense, leave a comment in the code as to why it's been done that way.

manager/src/main/java/com/redhat/service/bridge/manager/actions/connectors/SlackActionValidator.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Map;
+
+import javax.enterprise.context.ApplicationScoped;
+
+import com.redhat.service.bridge.actions.ActionParameterValidator;
+import com.redhat.service.bridge.actions.ValidationResult;
+import com.redhat.service.bridge.infra.models.actions.BaseAction;
+
+@ApplicationScoped
+public class SlackActionValidator implements ActionParameterValidator {
```
robpblake, line 12: @lucamolteni Did you create a JIRA issue for us to work on improving the ability to validate Action parameters?

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
```diff
@@ -77,22 +82,32 @@ public Processor createProcessor(String bridgeId, String customerId, ProcessorRe
...
-        p.setName(processorRequest.getName());
-        p.setDefinition(definitionToJsonNode(definition));
-        p.setSubmittedAt(ZonedDateTime.now());
-        p.setStatus(BridgeStatus.REQUESTED);
-        p.setBridge(bridge);
+        processorDAO.persist(newProcessor);
 
-        processorDAO.persist(p);
-        return p;
+        connectorService.createConnectorIfNeeded(resolvedAction, newProcessor, actionProvider);
```
robpblake, line 108: @lucamolteni cc @kostola  I wonder here if we should be using the outbox pattern to create/monitor the deployment of Connectors (and other resources) see: https://debezium.io/blog/2019/02/19/reliable-microservices-data-exchange-with-the-outbox-pattern/ In particular, please see the `Sending Events to the Outbox` part of ^

So in that model we would do something like:

1. Persist the Processor here with the resolved Action
2. Fire an Event that a Processor has been requested
3. Something listens for the Event and persists the need to create a Connector to fulfil the request
4. Another process creates the Connector and then calls back the ProcessorService when the Connector is ready
5. At that point we expose details to the Shard that it can start to provision the Executor

It's out of scope for this PR, but I want to raise the pattern/approach before we get too much further into the implementation so that we can have a conversation about adopting it.

manager/src/main/resources/application.properties
```diff
@@ -45,3 +45,23 @@ quarkus.native.resources.includes=exception/exceptionInfo.json
...
+# This shouldn't be put in the properties see https://issues.redhat.com/browse/MGDOBR-197
+kafka.bootstrap.servers=${KAFKA_BOOTSTRAP_SERVERS:localhost:9092}
+kafka.client.id=${KAFKA_CLIENT_ID:fake_id}
+kafka.client.secret=${KAFKA_CLIENT_SECRET:fake_secret}
+kafka.sasl.mechanism=PLAIN
+kafka.security.protocol=SASL_SSL
+kafka.sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required username="${KAFKA_CLIENT_ID:fake_id}" password="${KAFKA_CLIENT_SECRET:fake_secret}";
+
+# Managed Connectors
+managed-connectors.cluster.id=${MC_CLUSTER_ID:managed-connectors.cluster.id}
```
robpblake, line 60: @lucamolteni Can you please add the values required for our demo environment to the `Kustomize` overlays, otherwise as soon as we deploy this to our demo cluster, the environment will break.

Let me or @r00ta know if you need help doing that.

robpblake (commented):
@lucamolteni Some minor feedback from me at this point and one larger discussion/implementation approach I'd like us to consider as we continue with the implementation of this work. The larger discussion around the "outbox" pattern does not block this PR getting merged.

This is a really solid start to our MC work. Thanks for your effort here.

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsServiceImpl.java
lucamolteni: That's exactly the point, you won't find a case when `ConnectorService` is called without an active transaction because it's forbidden by the API - that's the meaning of `MANDATORY`. It's a way to explicitly said it's forbidden to call this method without an active transaction. 

For consistency, I'd avoid using `REQUIRED` for services that needs to be called from other services (and never alone). This prevents users from using it without an active transaction.

manager/src/main/java/com/redhat/service/bridge/manager/actions/connectors/SlackActionValidator.java
lucamolteni: https://issues.redhat.com/browse/MGDOBR-202

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
lucamolteni: I agree with it @robpblake

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsAuthImpl.java
```diff
@@ -0,0 +1,100 @@
...
+
+            HttpClient client = HttpClient.newBuilder().build();
+            HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+            if (response.statusCode() == 200) {
+                String tokens = response.body();
+                JsonObject json = new JsonObject(tokens);
+                return json.getString("access_token");
+            } else {
+                String errorMessage = String.format("Http Error Code %d\n", response.statusCode()) + response.body();
+                throw new RuntimeException(errorMessage);
```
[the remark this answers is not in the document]
lucamolteni: This code is bound to be replaced anyway so I wound't change it much from the original source code

robpblake (approved):
I'm +1 pending the missing kustomize update. 🥂

use-cases/managed-connectors-integration/src/main/java/com/redhat/service/ManagedConnectorServiceApplication.java
```diff
@@ -42,16 +57,20 @@ public static void main(String[] args) throws Exception {
...
         managedConnectorServiceApplication.serviceAccountSecret = args[4];
         managedConnectorServiceApplication.webhookUrl = args[5];
+        managedConnectorServiceApplication.bearerToken = managedConnectorServiceApplication.bearerTokenFromOfflineToken(managedConnectorServiceApplication.offlineToken);
+
         Connector slackConnector = managedConnectorServiceApplication.createSlackConnector();
         do {
             slackConnector = managedConnectorServiceApplication.pollSlackConnector(slackConnector);
         } while (!"ready".equals(slackConnector.getStatus()));
+
+        System.exit(0);
```
danielezonca, line 73: Is this necessary?

danielezonca (approved):
LGTM 👍

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsAuthImpl.java
```diff
@@ -0,0 +1,100 @@
...
+    public String bearerToken() {
+        if (bearerToken == null) {
+            bearerToken = bearerTokenFromOfflineToken(offlineToken);
+        }
+        return bearerToken;
+    }
+
+    @Scheduled(every = "5m")
+    void resetBearerToken() {
+        bearerToken = null;
```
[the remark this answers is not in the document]
lucamolteni: Yes @r00ta @danielezonca that's the one

manager/src/main/resources/application.properties
```diff
@@ -45,3 +45,23 @@ quarkus.native.resources.includes=exception/exceptionInfo.json
...
+managed-connectors.services.url=${MC_SERVICES_URL:managed-connectors.services.url}
+managed-connectors.topic-name=${MC_TOPIC_NAME:test_topic}
+
+managed-connectors.auth.server-url=${MC_AUTH_SERVER_URL:managed-connectors.auth.server-url}
```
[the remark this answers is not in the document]
lucamolteni: These values are to be provided by the user with a local `application.properties`.

Tests are running

manager/src/test/java/com/redhat/service/bridge/manager/dao/ConnectorsDAOTest.java
```diff
@@ -0,0 +1,109 @@
...
+        databaseManagerUtils.cleanDatabase();
+    }
+
+    private Processor createProcessor(Bridge bridge, String name) {
+        Processor p = new Processor();
+        p.setBridge(bridge);
+        p.setName(name);
+        p.setStatus(BridgeStatus.REQUESTED);
+        p.setSubmittedAt(ZonedDateTime.now());
+        p.setPublishedAt(ZonedDateTime.now());
```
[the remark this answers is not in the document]
lucamolteni: Will be reviewed in https://issues.redhat.com/browse/MGDOBR-142

PR #165, opened 2021-12-02
[MGDOBR-184] Adding validation to filters

fjtirado:
[MGDOBR-184](https://issues.redhat.com/browse/MGDOBR-184)


integration-tests/src/test/java/com/redhat/service/bridge/integration/tests/End2EndTestIT.java
```diff
@@ -190,6 +203,15 @@ public void testAddProcessor() throws Exception {
         assertThat(action.getParameters()).containsEntry(KafkaTopicAction.TOPIC_PARAM, TOPIC_NAME);
     }
 
+    @Test
```
r00ta, line 206: All the tests in this class are disabled. Is it possible to move this test to another module?

infra/src/main/java/com/redhat/service/bridge/infra/models/filters/FilterFactory.java
```diff
@@ -1,18 +1,31 @@
...
+                case StringEquals.FILTER_TYPE_NAME:
+                    return new StringEquals(key, value);
+                case ValuesIn.FILTER_TYPE_NAME:
+                    return new ValuesIn(key, ObjectMapperFactory.get().readValue(value, new TypeReference<List<Object>>() {
+                    }));
+                default:
+                    throw new IllegalArgumentException("Type " + type + " is not valid");
+            }
+        } catch (IOException io) {
+            throw new IllegalArgumentException("Value " + value + " is not valid");
```
robpblake, line 28: @fjtirado Can you please chain the `IOException` in the constructor so that we don't lose the full stracktrace.

infra/src/main/java/com/redhat/service/bridge/infra/models/filters/FilterFactory.java
```diff
@@ -1,18 +1,31 @@
...
+                case StringContains.FILTER_TYPE_NAME:
+                    return new StringContains(key, ObjectMapperFactory.get().readValue(value, new TypeReference<List<String>>() {
+                    }));
+                case StringEquals.FILTER_TYPE_NAME:
+                    return new StringEquals(key, value);
+                case ValuesIn.FILTER_TYPE_NAME:
+                    return new ValuesIn(key, ObjectMapperFactory.get().readValue(value, new TypeReference<List<Object>>() {
+                    }));
+                default:
+                    throw new IllegalArgumentException("Type " + type + " is not valid");
```
robpblake, line 25: @fjtirado I think we should throw a subclass of (or `EventBridgeManagerException`). At this minute, this Exception will not be correctly handled by our Error mapping as that only handles `EventBridgeManagerException` instances.

infra/src/main/java/com/redhat/service/bridge/infra/models/filters/StringBeginsWith.java
```diff
@@ -2,28 +2,24 @@
...
     }
 
-    public StringBeginsWith(String key, String values) {
+    public StringBeginsWith(String key, List<String> values) {
         super(FILTER_TYPE_NAME, key);
-        try {
-            this.values = ObjectMapperFactory.get().readValue(values, new TypeReference<List<String>>() {
-            });
-        } catch (JsonProcessingException e) {
-            throw new IllegalArgumentException("The value is not a list of strings.");
```
robpblake, line 25: @fjtirado Can we preserve these messages? The reason for the failure is much clearer in the current implementation e.g before your changes the error message I get is: `The value is not a list of strings` versus afterwards, `Type 'StringBeginsWith' is not valid`.

Ideally we'd go to something like `The value of filter type 'StringBeginsWith' must be a list of strings`. I think we were closer with the previous implementation.

robpblake (changes requested):

r00ta (changes requested):
Hi @fjtirado , I think the validation on the filters is not working properly because I can create a processor with 
```
{
  "name": "myProcessorsd",
  "action": {
    "name": "myKafkaAction",
    "parameters": {"topic":  "demoTopic"},
    "type": "KafkaTopicAction"
  },
  "filters": [
    {
      "key": "source",
      "type": "ValuesIn"
    }
  ]
}
```
successfully (the `values` are not specified, same for all the other filters). Can anybody double check this? @sutaakar

infra/src/main/java/com/redhat/service/bridge/infra/models/filters/StringBeginsWith.java
fjtirado: Current implementation was including in the filter class a responsibility that does not belong to a POJO (build the internal state from a string json representation)  
I would argue that saying that the value that failed is not a list of strings (given the fact that the class that failed is in the stack trace) is more useful that printing that saying that the name of the filter is not valid (former is at least indicating why the code is failing), although I of course agree that we can provide both.

infra/src/main/java/com/redhat/service/bridge/infra/models/filters/FilterFactory.java
fjtirado: @robpblake After reviewing the usage of FilterFactory, this class is only used by tests, so Im now wondering if we sohuld just move it to test and completely forgot this discussion ;)

fjtirado:
> Hi @fjtirado , I think the validation on the filters is not working properly because I can create a processor with
> 
> ```
> {
>   "name": "myProcessorsd",
>   "action": {
>     "name": "myKafkaAction",
>     "parameters": {"topic":  "demoTopic"},
>     "type": "KafkaTopicAction"
>   },
>   "filters": [
>     {
>       "key": "source",
>       "type": "ValuesIn"
>     }
>   ]
> }
> ```
> 
> successfully (the `values` are not specified, same for all the other filters). Can anybody double check this? @sutaakar

@r00ta you are right, the unit test was ok because the key is not passed, but if key is passed and value is not, the filter pass. Im investigating how to fix that with the current inheritance hierarchy

fjtirado:
@r00ta Validation should work fine now

infra/src/main/java/com/redhat/service/bridge/infra/models/filters/FilterFactory.java
fjtirado: Removing class and associated test

robpblake (approved):
Thanks @fjtirado . Looks good 👍

r00ta (approved):
Thank you @fjtirado

PR #183, opened 2021-12-15
MGDOBR-187: refactoring of error/exceptions

r00ta:
[MGDOBR-187](https://issues.redhat.com/browse/MGDOBR-187) 

This PR aims to refactor the errors/exceptions centralizing them in the `infra` module. The idea is that all the components can re-use the exceptions defined in the `infra` package, in particular the MGDOBR-114 ticket the shard is going to use some `Error`s defined in the `infra` module and report them to the manager. The manager will then report them to the user

A new abstraction for the exceptions/errors is introduced: the `EventBridgeException` goes away and it's replaced by 
1) `UserFaultException` (we might think about a better name, but it gives the idea): this exception and all its subclasses are supposed to be exceptions that are caused by the user's interaction/misconfiguration. For example, the user has submitted wrong kafka credentials. 
2) `PlatformInternalException`: all the exceptions that are not user's fault. For example the shard has full disk, Prometheus is not installed on the cluster. All those exception details are hidden to the user, he/she is going to get the generic `PlatformInternalException` regardless the subclass that raised the exception. The `reason` of `PlatformInternalException` is something like "an internal exception has occurred, please open a bug with the details you have.". => the user is not supposed to be able to fix this exception.  


r00ta (commented):
For reviewers

infra/src/main/resources/exception/exceptionInfo.json
```diff
@@ -0,0 +1,43 @@
+[
```
robpblake, line 1: @r00ta Should we move this into the Manager and put a blank file in the Shard that we will add to as we do the work in https://issues.redhat.com/browse/MGDOBR-114 ?

Most of these errors only really make sense in the context of the Fleet Manager and it would make things clearer IMHO if each component owned it's own `exceptionInfo.json`. WDYT?

infra/src/main/java/com/redhat/service/bridge/infra/exceptions/ErrorsService.java
```diff
@@ -0,0 +1,15 @@
...
+import com.redhat.service.bridge.infra.models.ListResult;
+import com.redhat.service.bridge.infra.models.QueryInfo;
+
+public interface ErrorsService {
+
+    ListResult<Error> getErrors(QueryInfo pageInfo);
+
+    Optional<Error> getError(int errorId);
+
+    Optional<Error> getError(Exception e);
```
robpblake, line 14: Not sure it feels right that we can't always return an Error given an `Exception`. That means that our API is going to be missing error/errors codes for certain failures. It feels like we need a default `Error` that will be returned unless there is an `Error` specifically configured for the passed `Exception`.

It looks like that is the functionality of the current implementation, so happy to do that in another JIRA if you agree?

robpblake (commented):
Thanks @r00ta . Couple of minor comments from me.

infra/src/main/java/com/redhat/service/bridge/infra/exceptions/ErrorsService.java
fjtirado: Returning optional is just a convenience thing so the layer invoking this handle the situation in which there is no error for an exception. Logic that Im not sure belongs to ErrorService.

infra/src/main/resources/exception/exceptionInfo.json
r00ta: @robpblake I thought the shard and the manager would have shared some exceptions. Or should we just duplicate the exception/error definition in every component?

fjtirado: If error API is going to be available for shard and manager, then we need two files. 
If this is going to be exposed only in the manager, then just one file 
I hope you like this rule ;)

r00ta: @fjtirado the `errorAPI` is only in the manager. The user is not supposed to be able to reach directly any shard by design

fjtirado: Then, all errors should be defined in just one file. It does not care that the error is not produced directly by the manager, it is handled by it anyway.

infra/src/main/resources/exception/exceptionInfo.json
```diff
@@ -0,0 +1,80 @@
...
+  "isUserException": true
+},
+{
+  "exception":"com.redhat.service.bridge.infra.exceptions.definitions.platform.PrometheusNotInstalledException",
+  "id": 1001,
+  "code": "OPENBRIDGE-PLATFORM-1",
+  "reason": "Prometheus is not installed on the shard.",
+  "isUserException": false
+},
+{
```
r00ta, line 73: `exceptionInfo.json` now contains the user exceptions and the platform exceptions

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/resources/Condition.java
```diff
@@ -47,6 +47,7 @@
     private Date lastTransitionTime;
     private ConditionReason reason;
     private String message;
+    private String errorCode;
```
r00ta, line 50: `errorCode` is added to the `Condition`

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/controllers/BridgeExecutorController.java
```diff
@@ -94,7 +99,12 @@ public void init(EventSourceManager eventSourceManager) {
...
             LOGGER.warn("Executor service monitor resource BridgeExecutor: '{}' in namespace '{}' is failed to deploy, Prometheus not installed.", bridgeExecutor.getMetadata().getName(),
                     bridgeExecutor.getMetadata().getNamespace());
-            bridgeExecutor.getStatus().markConditionFalse(ConditionType.Ready, ConditionReason.PrometheusUnavailable, ConditionMessages.PROMETHEUS_UNVAILABLE);
+            Error prometheusNotAvailableError = errorsService.getError(PrometheusNotInstalledException.class)
+                    .orElseThrow(() -> new RuntimeException("PrometheusNotInstalledException not found in error catalog"));
+            bridgeExecutor.getStatus().markConditionFalse(ConditionType.Ready,
+                    ConditionReason.PrometheusUnavailable,
+                    prometheusNotAvailableError.getReason(),
+                    prometheusNotAvailableError.getCode());
             notifyManager(bridgeExecutor, BridgeStatus.FAILED);
```
r00ta, line 108: The shard retrieves the error code and store it in the `Condition`. The `errorCode` will be included in the manager update in another PR

infra/src/main/java/com/redhat/service/bridge/infra/exceptions/ErrorInMemoryDAO.java
```diff
@@ -0,0 +1,122 @@
...
+            LOGGER.error("Error closing file with exception errors", io);
+        }
+    }
+
+    private void populate(ErrorInfo errorInfo) {
+        Error error = errorInfo.toError();
+        if (errorInfo.isUserException()) {
+            userErrorList.add(error);
+            userErrorsFromId.put(error.getId(), error);
+        }
```
r00ta, line 96: only user exceptions are present in the user's catalog

r00ta (commented):
For reviewers

fjtirado:
@r00ta Just my 2c. I do not think we should store code string in the definition file. Thats why I do not see any extra value in keeping two sequences of id, one for user exceptions and another for platform ones. I think string code should remain "generable" from the unique id (by using the same prefix both for user and platform(, that should be the same sequence for all exceptions. 
Obviously, this is not the idea followed in this PR and it is ok as far as we have a good reason to differentiate between "user" and "platform" exception in the code. Im kind of reluctant to expose that difference there, but there might be reasons Im not aware. I think we need to elaborate on them.

r00ta:
> @r00ta Just my 2c. I do not think we should store code string in the definition file. Thats why I do not see any extra value in keeping two sequences of id, one for user exceptions and another for platform ones. I think string code should remain "generable" from the unique id (by using the same prefix both for user and platform), that should be the same sequence for all exceptions. Obviously, this is not the idea followed in this PR and it is ok as far as we have a good reason to differentiate between "user" and "platform" exception in the code. Im kind of reluctant to expose that difference there, but there might be reasons Im not aware. I think we need to elaborate on them.

I moved the `code` to the `json` for a specific reason: if the `exceptionInfo.json` is our kind of catalog, I think all the information should be there already and not generated somewhere in the code. 
It's much more immediate to take a look at the json catalog without having to run any application. 
Regarding the `code` separation between the `user` and `platform` exception: I introduced it because from the api guidelines https://gitlab.cee.redhat.com/service/api-guidelines it looks like we can use the code to specify the category of the error. But if we don't like it, we can go for the same prefix for all the exceptions and keep also the unique sequence of numbers. The user's error catalog is going to have some holes here and there. 
Anyway I'm pretty sure the entire catalog thing will be properly reviewed by the productization/documentation team only when we will reach GA. So all the codes/reasons/id will probably change according to their feedbacks. My main point is to agree on the general structure and the exception/error handling/notification between services

r00ta:
@fjtirado @danielezonca see here https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/33efc7822bba1cee7e790a0998b9c728ccae4877/pkg/errors/errors.go#L20 . They have different prefix depending on the category - different catalogs. I think we are in the same situation: we have the user catalog and our internal catalog.

fjtirado:
> > @r00ta Just my 2c. I do not think we should store code string in the definition file. Thats why I do not see any extra value in keeping two sequences of id, one for user exceptions and another for platform ones. I think string code should remain "generable" from the unique id (by using the same prefix both for user and platform), that should be the same sequence for all exceptions. Obviously, this is not the idea followed in this PR and it is ok as far as we have a good reason to differentiate between "user" and "platform" exception in the code. Im kind of reluctant to expose that difference there, but there might be reasons Im not aware. I think we need to elaborate on them.
> 
> I moved the `code` to the `json` for a specific reason: if the `exceptionInfo.json` is our kind of catalog, I think all the information should be there already and not generated somewhere in the code. It's much more immediate to take a look at the json catalog without having to run any application. Regarding the `code` separation between the `user` and `platform` exception: I introduced it because from the api guidelines https://gitlab.cee.redhat.com/service/api-guidelines it looks like we can use the code to specify the category of the error. But if we don't like it, we can go for the same prefix for all the exceptions and keep also the unique sequence of numbers. The user's error catalog is going to have some holes here and there. Anyway I'm pretty sure the entire catalog thing will be properly reviewed by the productization/documentation team only when we will reach GA. So all the codes/reasons/id will probably change according to their feedbacks. My main point is to agree on the general structure and the exception/error handling/notification between services

@r00ta Now I think about it, Im wondering why a unique string code is needed at all if we have a unique integer code, or the other way around, why do we need to have a unique integer code if we have a unique string coce. I mean, having two unique ids sounds like a waste of resources and, as any duplication in life, a potential source of problems. A composite unique ID makes more sense, meaning that the numeric id is unique within a certain "system", so you still has two fields "id" and "system", but they are not strictly unique. Maybe we should discuss this with the "convention" guys.

r00ta:
> > > @r00ta Just my 2c. I do not think we should store code string in the definition file. Thats why I do not see any extra value in keeping two sequences of id, one for user exceptions and another for platform ones. I think string code should remain "generable" from the unique id (by using the same prefix both for user and platform), that should be the same sequence for all exceptions. Obviously, this is not the idea followed in this PR and it is ok as far as we have a good reason to differentiate between "user" and "platform" exception in the code. Im kind of reluctant to expose that difference there, but there might be reasons Im not aware. I think we need to elaborate on them.
> > 
> > 
> > I moved the `code` to the `json` for a specific reason: if the `exceptionInfo.json` is our kind of catalog, I think all the information should be there already and not generated somewhere in the code. It's much more immediate to take a look at the json catalog without having to run any application. Regarding the `code` separation between the `user` and `platform` exception: I introduced it because from the api guidelines https://gitlab.cee.redhat.com/service/api-guidelines it looks like we can use the code to specify the category of the error. But if we don't like it, we can go for the same prefix for all the exceptions and keep also the unique sequence of numbers. The user's error catalog is going to have some holes here and there. Anyway I'm pretty sure the entire catalog thing will be properly reviewed by the productization/documentation team only when we will reach GA. So all the codes/reasons/id will probably change according to their feedbacks. My main point is to agree on the general structure and the exception/error handling/notification between services
> 
> @r00ta Now I think about it, Im wondering why a unique string code is needed at all if we have a unique integer code, or the other way around, why do we need to have a unique integer code if we have a unique string code?. I mean, having two unique ids sounds like a waste of resources and, as any duplication in life, a potential source of problems. A composite unique ID makes more sense, meaning that the numeric id is unique within a certain "system", so you still has two fields "id" and "system", but they are not strictly unique. Maybe we should discuss this with the "convention" guys.

See here
Screenshot from 2021-12-22 11-08-31

I think the `id` is just unique for the catalog, i.e. the user can query our api with that number `id` and retrieve the details. The `code` is just unique across all the projects

r00ta:
And actually we should use the identifier of our API... which is still unknown probably

fjtirado:
@r00ta Thats another reason (prefix might change) why I prefer to generate the code from the id, rather than storing in the file. But I think is kind of clear that the prefix should be one per application, so assuming openbridge as a whole is an application, the prefix should be unique. This way you avoid the id "jump", user exceptions starting with 1 and system exceptions starting with 1000.

r00ta:
> @r00ta Thats another reason (prefix might change) why I prefer to generate the code from the id, rather than storing it in the file. Besides storing the code in the file or generating it, I think is kind of clear that the prefix should be one per application, so assuming openbridge as a whole is an application, the prefix should be unique. One implication of that is that you can avoid the id "jump convention" in the json file, user exceptions starting with 1 and system exceptions starting with 1000, which is not needed anymore.

I've reverted the changes and now the code is generated programmatically. The same prefix is used for the user and platform exceptions. Let me know if you have any additional comments about the PR

infra/src/main/java/com/redhat/service/bridge/infra/exceptions/ErrorInMemoryDAO.java
```diff
@@ -0,0 +1,116 @@
...
+import com.redhat.service.bridge.infra.models.QueryInfo;
+import com.redhat.service.bridge.infra.utils.Constants;
+
+import io.quarkus.runtime.Quarkus;
+
+@ApplicationScoped
+public class ErrorInMemoryDAO implements ErrorDAO {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ErrorInMemoryDAO.class);
+    private static final ObjectMapper MAPPER = new ObjectMapper();
```
fjtirado, line 30: I know this is picky, but MAPPER and LOGGER are not literal constants, but final references, and I do not think they should be in uppercase 
see https://stackoverflow.com/a/12069687/16843807

r00ta: Well, if you really want to discuss such things https://rules.sonarsource.com/java/tag/convention/RSPEC-1312 ;) The brutal truth is that there is no rule on this regard, only conventions. If you really have a strong opinion on this, I'll make them lowercase

fjtirado: I think we need a convention for singleton instance naming (which are private final references, but different from literal constants, than since the 70s has been in uppercase in all programming languages) , but it is outside the scope of this JIRA. 
Im not stronger support of conventions, because
- I do not like to put { when there is just one statement after an if, but there is a convention and I have to add an uneeded bracket  because of conventions
- My whole programmer life I haves used  { in new line, but there is a google convention to put it on the same line. 
and so on. 
Lets discuss on a different forum

fjtirado (approved):

infra/src/main/java/com/redhat/service/bridge/infra/exceptions/Error.java
```diff
@@ -1,16 +1,18 @@
-package com.redhat.service.bridge.manager.models;
+package com.redhat.service.bridge.infra.exceptions;
 
 public class Error {
```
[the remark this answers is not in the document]
r00ta: 100% agree

infra/pom.xml
```diff
@@ -15,6 +15,10 @@
   <name>Event Bridge :: Infra</name>
 
   <dependencies>
+    <dependency>
+      <groupId>org.jboss.spec.javax.ws.rs</groupId>
```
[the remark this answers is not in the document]
r00ta: we need also some `javax.ws.rs.` classes that are not included in those modules :/ Any other suggestion?

infra/src/main/java/com/redhat/service/bridge/infra/exceptions/ErrorDAO.java
```diff
@@ -0,0 +1,15 @@
+package com.redhat.service.bridge.infra.exceptions;
+
+import com.redhat.service.bridge.infra.models.ListResult;
+import com.redhat.service.bridge.infra.models.QueryInfo;
+
+public interface ErrorDAO {
```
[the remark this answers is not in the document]
r00ta: I've refactored the interface into 
```java 

public interface BridgeErrorDAO {

    ListResult<BridgeError> findAllErrorsByType(QueryInfo queryInfo, BridgeErrorType type);

    BridgeError findErrorByIdAndType(int errorId, BridgeErrorType type);

    BridgeError findByException(Exception ex);

    BridgeError findByException(Class clazz);
}
```

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/controllers/BridgeIngressController.java
```diff
@@ -119,7 +124,12 @@ public void init(EventSourceManager eventSourceManager) {
         } else {
             LOGGER.warn("Ingress monitor resource BridgeIngress: '{}' in namespace '{}' is failed to deploy, Prometheus not installed.", bridgeIngress.getMetadata().getName(),
                     bridgeIngress.getMetadata().getNamespace());
-            bridgeIngress.getStatus().markConditionFalse(ConditionType.Ready, ConditionReason.PrometheusUnavailable, ConditionMessages.PROMETHEUS_UNVAILABLE);
+            Error prometheusNotAvailableError = errorsService.getError(PrometheusNotInstalledException.class)
+                    .orElseThrow(() -> new RuntimeException("PrometheusNotInstalledException not found in error catalog"));
```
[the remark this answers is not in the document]
r00ta: We might add a `CatalogErrorNotFoundException extends PlatformInternalException` and use that, wdyt?

r00ta:
@danielezonca I've resolved all the comments are replied to a couple. Could you take a second look? Thanks!

infra/pom.xml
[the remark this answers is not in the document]
danielezonca: 🤔 
It seems also Quarkus is mixing jakarta and jboss.spec modules 
https://github.com/quarkusio/quarkus/blob/main/independent-projects/resteasy-reactive/server/runtime/pom.xml#L22-L29
So let's do the same 🤔

danielezonca (approved):

PR #184, opened 2021-12-15
[MGDOBR-163] Adding Pormetheus ServiceMonitor to BridgeExecutor

ricardozanini:
Signed-off-by: Ricardo Zanini <zanini@redhat.com>

[MGDOBR-163](https://issues.redhat.com/browse/MGDOBR-163) 

Small change to `BridgeExecutor` to include `ServiceMonitor`. If Prometheus is not available, the CR will change its status to `Ready: false` and will notify the manager, so the component status will be `FAILED`.

I tried to do a small refactoring of the Controllers since they have almost the same algorithm. It's fairly easy to do, but the change would be huge. I'd rather have it in another PR. @r00ta @robpblake wdyt? I can open a JIRA to make the `ShardController` and `ShardResource` to manage the common bits between these two.

### DIFF
--- a/shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/controllers/BridgeExecutorController.java
+++ b/shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/controllers/BridgeExecutorController.java
@@ -87,11 +87,13 @@ public class BridgeExecutorController implements ResourceController<BridgeExecut
             return UpdateControl.updateStatusSubResource(bridgeExecutor);
         }
 
-        Optional<ServiceMonitor> serviceMonitor = monitorService.fetchOrCreateServiceMonitor(bridgeExecutor, service);
+        Optional<ServiceMonitor> serviceMonitor = monitorService.fetchOrCreateServiceMonitor(bridgeExecutor, service, BridgeExecutor.COMPONENT_NAME);
         if (serviceMonitor.isPresent()) {
             // this is an optional resource
             LOGGER.debug("Executor service monitor resource BridgeExecutor: '{}' in namespace '{}' is ready", bridgeExecutor.getMetadata().getName(), bridgeExecutor.getMetadata().getNamespace());
         } else {
+            LOGGER.warn("Executor service monitor resource BridgeExecutor: '{}' in namespace '{}' is failed to deploy, Prometheus not installed.", bridgeExecutor.getMetadata().getName(),
+                    bridgeExecutor.getMetadata().getNamespace());
             bridgeExecutor.getStatus().markConditionFalse(ConditionType.Ready, ConditionReason.PrometheusUnavailable, ConditionMessages.PROMETHEUS_UNVAILABLE);
             notifyManager(bridgeExecutor, BridgeStatus.FAILED);
             return UpdateControl.updateStatusSubResource(bridgeExecutor);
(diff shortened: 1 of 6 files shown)


shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/controllers/BridgeExecutorController.java
```diff
@@ -75,6 +86,17 @@ public void init(EventSourceManager eventSourceManager) {
             bridgeExecutor.getStatus().markConditionTrue(ConditionType.Augmentation, ConditionReason.ServiceNotReady);
             return UpdateControl.updateStatusSubResource(bridgeExecutor);
         }
+
+        Optional<ServiceMonitor> serviceMonitor = monitorService.fetchOrCreateServiceMonitor(bridgeExecutor, service);
+        if (serviceMonitor.isPresent()) {
+            // this is an optional resource
+            LOGGER.debug("Executor service monitor resource BridgeExecutor: '{}' in namespace '{}' is ready", bridgeExecutor.getMetadata().getName(), bridgeExecutor.getMetadata().getNamespace());
+        } else {
+            bridgeExecutor.getStatus().markConditionFalse(ConditionType.Ready, ConditionReason.PrometheusUnavailable, ConditionMessages.PROMETHEUS_UNVAILABLE);
```
r00ta, line 95: I'd also add a `LOGGER.warn` since the deployment failed due to our internal misconfiguration

r00ta (changes requested):
Hey @ricardozanini , you are missing to set the `app.kubernetes.io/component: ` label on the ServiceMonitor CRD. If you don't set it, the controller does not get called when the resource is updated

sutaakar (approved):

ricardozanini:
> Hey @ricardozanini , you are missing to set the `app.kubernetes.io/component: ` label on the ServiceMonitor CRD. If you don't set it, the controllers do not get called when the resource is updated

Indeed!

I've updated the PR. Please I need the reply about the refactoring of these controllers.

r00ta (approved):
Thanks @ricardozanini . 

Regarding the refactoring of the Controllers, I have some doubts honestly. They are _similar_ because they both provision a Deployment and a Service. But
1) We still need to register 2 controllers
2) The Deployment and the Service are configured differently
3) The BridgeController deploys a Route then

I don't know if the code is going to improve just to remove the apparent duplication of the management of the Deployment/Service (literally 10 lines of code). At the moment the two controllers are pretty separated, and I think it's a good thing for different reasons at the end.

But maybe I'm not getting what you have in mind, so if you have a clear picture of the refactoring of the components a diagram or a schema would help me.

r00ta:
> Thanks @ricardozanini .
> 
> Regarding the refactoring of the Controllers, I have some doubts honestly. They are _similar_ because they both provision a Deployment and a Service. But
> 
> 1. We still need to register 2 controllers
> 2. The Deployment and the Service are configured differently
> 3. The BridgeController deploys a Route then
> 
> I don't know if the code is going to improve just to remove the apparent duplication of the management of the Deployment/Service (literally 10 lines of code). At the moment the two controllers are pretty separated, and I think it's a good thing for different reasons at the end.
> 
> But maybe I'm not getting what you have in mind, so if you have a clear picture of the refactoring of the components a diagram or a schema would help me.

Thinking about it again.. maybe you are thinking about something like a `public abstract class BridgeIngressController<T extends CustomResource> implements ResourceController<T>` and put there some utility methods like `isDeploymentReady` and some others to be overwritten by the concrete Controllers?

ricardozanini:
> Thinking about it again.. maybe you are thinking about something like a public abstract class BridgeIngressController<T extends CustomResource> implements ResourceController<T> and put there some utility methods like isDeploymentReady and some others to be overwritten by the concrete Controllers?

Something like that. They will share the use cases, but we will keep them separatedely. There's a lot of opportunities to use some patterns to improve reuse.

PR #194, opened 2021-12-22
MGDOBR-207: fix refresh token expiration and add safe buffer to refresh the token

r00ta:
[MGDOBR-207](https://issues.redhat.com/browse/MGDOBR-207) 

This PR aims to add 30s buffer to the token expiration check. In addition to that, in case the refresh token has expired, a new fresh token is retrieved. 

### DIFF
--- a/shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/ManagerSyncServiceImpl.java
+++ b/shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/ManagerSyncServiceImpl.java
@@ -9,6 +9,7 @@ import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
+import org.eclipse.microprofile.context.ThreadContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -22,6 +23,7 @@ import com.redhat.service.bridge.infra.models.dto.ProcessorDTO;
 import com.redhat.service.bridge.shard.operator.exceptions.DeserializationException;
 
 import io.quarkus.oidc.client.OidcClient;
+import io.quarkus.oidc.client.OidcClientException;
 import io.quarkus.oidc.client.Tokens;
 import io.quarkus.runtime.Quarkus;
 import io.quarkus.scheduler.Scheduled;
@@ -51,6 +53,9 @@ public class ManagerSyncServiceImpl implements ManagerSyncService {
 
     @Inject
     OidcClient client;
+
+    @Inject
+    ThreadContext threadContext;
 
     Tokens currentTokens;
 
@@ -176,6 +181,10 @@ public class ManagerSyncServiceImpl implements ManagerSyncService {
      * @throws IllegalStateException in case of an invalid HTTP response
      */
     private void validateHttpResponseBeforeParsing(HttpResponse<Buffer> httpResponse) {
+        if (httpResponse.statusCode() == 401) {
+            LOGGER.debug("Request to the manager was unauthorized.");
+            threadContext.contextualRunnable(this::refreshTokens);
+        }
         if (httpResponse.statusCode() < 200 || httpResponse.statusCode() >= 400) {
             throw new DeserializationException(String.format("Got %d HTTP status code response, skipping deserialization process", httpResponse.statusCode()));
         }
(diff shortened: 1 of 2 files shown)


shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/ManagerSyncServiceImpl.java
```diff
@@ -176,24 +181,34 @@ private void processingComplete(Class<?> entity) {
      * @throws IllegalStateException in case of an invalid HTTP response
      */
     private void validateHttpResponseBeforeParsing(HttpResponse<Buffer> httpResponse) {
+        if (httpResponse.statusCode() == 401) {
```
fjtirado, line 184: do we really need this? is it not the check when we create the request enough?

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/ManagerSyncServiceImpl.java
```diff
@@ -176,24 +181,34 @@ private void processingComplete(Class<?> entity) {
...
+            LOGGER.debug("Request to the manager was unauthorized.");
+            threadContext.contextualRunnable(this::refreshTokens);
```
kostola, line 186: what is the purpose of this?

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/ManagerSyncServiceImpl.java
r00ta: Nope, the token can expire in the meanwhile

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/ManagerSyncServiceImpl.java
r00ta: @kostola you mean `threadContext.contextualRunnable` or the refresh of the token when we get unauthorized?

kostola: the use of `threadContext.contextualRunnable`

r00ta: Because `validateHttpResponseBeforeParsing` is running on a thread that can not block

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/ManagerSyncServiceImpl.java
fjtirado: yes, but which is the effect? I understood that if the token expired just in that small interval (between the check to validate the token is active and the actual call) then we fire a token refreshal here, but the failed call, is it performed again?
If not performed again, it will return a 401 to the caller and the next call will try to refresh the token anyway (because the isTokenExpired will return true in that case), so thats Im wodering is this is really needed.

r00ta: I'm implementing the retry mechanism in another pull request

r00ta: I'll rebase this one once I'm ready with the other one

shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/ManagerSyncServiceImpl.java
```diff
@@ -176,24 +181,34 @@ private void processingComplete(Class<?> entity) {
...
         return executor.apply(request);
     }
+
+    private void refreshTokens() {
+        LOGGER.info("Refreshing tokens..");
+        Tokens tokens = currentTokens;
+        try {
+            tokens = client.refreshTokens(tokens.getRefreshToken()).await().atMost(SSO_CONNECTION_TIMEOUT);
+        } catch (OidcClientException e) {
+            LOGGER.warn("Shard could not use refresh token. Trying to get a new fresh token.", e);
```
danielezonca, line 208: Do we know when this can happen? Are there expected scenarios where this is supposed to happen?

r00ta: The refresh token has a lifespan: when it is not valid anymore we have to retrieve a brand new token with its refresh token

r00ta:
This fix is included in https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/198 . Closing this one

PR #196, opened 2021-12-22
MGDOBR-168 MGDOBR-169: create kafka topic and service account in RHOAS (with app-services-sdk)

kostola:
* [MGDOBR-168](https://issues.redhat.com/browse/MGDOBR-168)
* [MGDOBR-169](https://issues.redhat.com/browse/MGDOBR-169)

**IMPORTANT:** replaces #190


rhoas-client/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+    <dependency>
+      <groupId>com.redhat.cloud</groupId>
+      <artifactId>kafka-management-sdk</artifactId>
+      <version>0.11.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.redhat.cloud</groupId>
+      <artifactId>kafka-instance-sdk</artifactId>
+      <version>0.11.0</version>
+    </dependency>
```
danielezonca, line 31: Move to depMap in the `build-parent`

rhoas-client/src/test/java/com/redhat/service/bridge/rhoas/RhoasServiceTest.java
```diff
@@ -0,0 +1,107 @@
...
+        wireMockServer.verify(0, WireMock.deleteRequestedFor(WireMock.urlEqualTo(kafkaMgmtConfigurator.pathOf("/service_accounts/" + TEST_SERVICE_ACCOUNT_ID))));
+        wireMockServer.verify(0, WireMock.deleteRequestedFor(WireMock.urlMatching(kafkaInstanceConfigurator.pathOf("/acls?.*"))));
+    }
+
+    @Test
+    void testCreateTopicAndConsumerServiceAccountWithBrokenTopicCreation() {
+        configureMockAPIWithBrokenTopicCreation();
+
+        rhoasService.createTopicAndConsumerServiceAccount(new TopicAndServiceAccountRequest(TEST_TOPIC_NAME, TEST_SERVICE_ACCOUNT_NAME))
+                .onFailure().invoke(Throwable::printStackTrace)
```
danielezonca, line 58: Why just `printStackTrace` instead of proper test `fail()`? (same for similar pattern in this class)

rhoas-client/src/test/java/com/redhat/service/bridge/rhoas/RhoasAPI.java
```diff
@@ -0,0 +1,29 @@
...
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+import com.redhat.service.bridge.rhoas.dto.TopicAndServiceAccountRequest;
+import com.redhat.service.bridge.rhoas.dto.TopicAndServiceAccountResponse;
+
+@Path("/rhoas/")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public class RhoasAPI {
```
danielezonca, line 18: Add a comment to explain the reason of this test class. As far as I remember you told me it is necessary to test that there is no blocking code in the I/O Thread of Vertx. Am I right?

rhoas-client/src/main/resources/application.properties
```diff
@@ -0,0 +1,21 @@
+event-bridge.rhoas.mgmt-api.host=https://api.openshift.com
+event-bridge.rhoas.instance-api.host=http://localhost
+event-bridge.rhoas.sso.red-hat.auth-server-url=https://sso.redhat.com/auth/realms/redhat-external
+event-bridge.rhoas.sso.red-hat.refresh-token=dummy-refresh-token
+event-bridge.rhoas.sso.mas.auth-server-url=https://identity.api.openshift.com/auth/realms/rhoas
```
danielezonca, line 5: Please use variables (with default value) for each of them

rhoas-client/src/main/resources/application.properties
```diff
@@ -0,0 +1,21 @@
+event-bridge.rhoas.mgmt-api.host=https://api.openshift.com
+event-bridge.rhoas.instance-api.host=http://localhost
+event-bridge.rhoas.sso.red-hat.auth-server-url=https://sso.redhat.com/auth/realms/redhat-external
+event-bridge.rhoas.sso.red-hat.refresh-token=dummy-refresh-token
```
danielezonca, line 4: What does `dummy-refresh-token` mean?

rhoas-client/src/main/resources/application.properties
```diff
@@ -0,0 +1,21 @@
...
+event-bridge.rhoas.sso.mas.auth-server-url=https://identity.api.openshift.com/auth/realms/rhoas
+event-bridge.rhoas.sso.mas.client-id=dummy-client-id
+event-bridge.rhoas.sso.mas.client-secret=dummy-client-id
```
danielezonca, line 7: Same here

rhoas-client/src/main/resources/application.properties
```diff
@@ -0,0 +1,21 @@
...
+quarkus.oidc-client.red-hat-sso.auth-server-url=${event-bridge.rhoas.sso.red-hat.auth-server-url}
+quarkus.oidc-client.red-hat-sso.client-id=cloud-services
+quarkus.oidc-client.red-hat-sso.credentials.secret=secret
+quarkus.oidc-client.red-hat-sso.grant.type=refresh
+# MAS SSO
+quarkus.oidc-client.mas-sso.auth-server-url=${event-bridge.rhoas.sso.mas.auth-server-url}
+quarkus.oidc-client.mas-sso.client-id=${event-bridge.rhoas.sso.mas.client-id}
+quarkus.oidc-client.mas-sso.credentials.secret=${event-bridge.rhoas.sso.mas.client-secret}
+quarkus.oidc-client.mas-sso.grant.type=client
+quarkus.oidc-client.mas-sso.scopes=email
```
danielezonca, line 21: Are these fixed values?

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/dto/TopicAndServiceAccountResponse.java
```diff
@@ -0,0 +1,75 @@
+package com.redhat.service.bridge.rhoas.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class TopicAndServiceAccountResponse {
```
danielezonca, line 6: I would expect in the future that this bean (or a similar one) can be provided by the SDK 🤔

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/dto/TopicAndServiceAccountRequest.java
```diff
@@ -0,0 +1,42 @@
+package com.redhat.service.bridge.rhoas.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class TopicAndServiceAccountRequest {
```
danielezonca, line 6: Same here

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/KafkaInstanceAdminClientImpl.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @ConfigProperty(name = "event-bridge.rhoas.instance-api.host")
+    String basePath;
+
+    @Inject
+    Vertx vertx;
+
+    @Inject
+    @NamedOidcClient("mas-sso")
+    Tokens tokens;
```
danielezonca, line 33: We probably need to review token refresh mechanism. I think it is better to do that with another PR

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/KafkasMgmtV1ClientImpl.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public Uni<Void> deleteServiceAccount(String id) {
+        return securityVoidCall(api -> api.deleteServiceAccountById(id));
+    }
+
+    @Override
+    protected String getAccessToken() {
+        if (tokens == null || tokens.isAccessTokenExpired()) {
+            tokens = client.refreshTokens(refreshToken).await().indefinitely();
```
danielezonca, line 57: Same here

rhoas-client/src/test/java/com/redhat/service/bridge/rhoas/RhoasAPI.java
kostola: I wrote this dummy test API to simulate a real usage of the `RhoasClient` in a real JAX-RS bean.
This allows to spot errors caused by the client mistakenly blocking the Vert.x event loop thread, that can't be spotted testing directly the client in a `@QuarkusTest` (because blocking the test main thread doesn't cause any error: it simply blocks).

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/KafkaInstanceAdminClientImpl.java
kostola: I was checking what @r00ta is doing in #198 , but probably we can move on with this and then refactor the token refresh mechanism later once we finalize it.

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/KafkasMgmtV1ClientImpl.java
kostola: I was checking what @r00ta is doing in #198 , but probably we can move on with this and then refactor the token refresh mechanism later once we finalize it.

rhoas-client/src/main/resources/application.properties
kostola: It's just a placeholder to make it clear where to put the refresh token for Red Hat SSO (aka offline token in this case)

rhoas-client/src/main/resources/application.properties
kostola: Yes (based on the current MAS SSO auth flow)

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/exceptions/RhoasClientException.java
```diff
@@ -0,0 +1,13 @@
+package com.redhat.service.bridge.rhoas.exceptions;
+
+import java.util.List;
+
+import io.smallrye.mutiny.CompositeException;
+
+public class RhoasClientException extends RuntimeException {
```
robpblake, line 7: @kostola Any reason why these shouldn't extend our own `EventBridgeManagerException` hierarchy?

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/exceptions/AppServicesException.java
```diff
@@ -0,0 +1,13 @@
+package com.redhat.service.bridge.rhoas.exceptions;
+
+public class AppServicesException extends RuntimeException {
```
robpblake, line 3: @kostola Same here. Should this extend the `EventBridgeManagerException`?

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/KafkasMgmtV1ClientImpl.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Inject
+    @NamedOidcClient("red-hat-sso")
+    OidcClient client;
+
+    Tokens tokens;
+
+    @PostConstruct
+    void onPostConstruct() {
+        init(vertx);
```
robpblake, line 41: @kostola Can the `@Inject` be on the `Vertx` field in the super-class? Doc of `javax.inject.Inject` says:

> Fields and methods in superclasses are injected before those in subclasses

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/KafkasMgmtV1ClientImpl.java
robpblake: We also shouldn't `await().indefinitely()`. A sensible timeout should be applied with a clear error message if we cannot obtain a token within a period of time.

@kostola How do you want to address this? In this PR or wait for @r00ta work?

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/RhoasClient.java
```diff
@@ -0,0 +1,12 @@
+package com.redhat.service.bridge.rhoas;
+
+import com.redhat.service.bridge.rhoas.dto.TopicAndServiceAccountRequest;
+import com.redhat.service.bridge.rhoas.dto.TopicAndServiceAccountResponse;
+
+import io.smallrye.mutiny.Uni;
+
+public interface RhoasClient {
+
+    Uni<TopicAndServiceAccountResponse> createTopicAndConsumerServiceAccount(TopicAndServiceAccountRequest request);
```
robpblake, line 10: @kostola I don't see this integrated into any of the code that is called when the user requests an OpenBridge Instance? How/When is this method actually being called?

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/dto/TopicAndServiceAccountRequest.java
```diff
@@ -0,0 +1,42 @@
+package com.redhat.service.bridge.rhoas.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
robpblake, line 5: @kostola I think I need to see where the `RhoasClient` is used in response to creating an OpenBridge Instance to offer feedback on this. My initial reaction is that the `RhoasClient` should take care of generating the topic name and service account name and not the thing calling it.

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/RhoasClientImpl.java
```diff
@@ -0,0 +1,274 @@
...
+                    LOG.error(reason, failure);
+                    return Uni.createFrom().item(ctx.withFailure(failure, reason));
+                });
+    }
+
+    private Uni<Context> revertTopicCreation(final Context ctx) {
+        LOG.info("Deleting topic {} to restore original state after failure", ctx.getTopicName());
+        return instanceClient.deleteTopic(ctx.getTopicName())
+                .onItem().transform(v -> {
+                    LOG.info("Deleted topic {}", ctx.getTopicName());
```
robpblake, line 68: @kostola Can you please ensure all placeholders in log statements are surrounded by `'`.

rhoas-client/src/main/resources/application.properties
robpblake: @kostola Can you make this accept an ENV variable as that is how we will configure the offline token in our deployment

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/AbstractAppServicesClientImpl.java
```diff
@@ -0,0 +1,100 @@
...
+
+import com.openshift.cloud.api.kas.SecurityApi;
+import com.openshift.cloud.api.kas.auth.AclsApi;
+import com.openshift.cloud.api.kas.auth.TopicsApi;
+import com.redhat.service.bridge.rhoas.exceptions.AppServicesException;
+
+import io.smallrye.mutiny.Uni;
+import io.vertx.mutiny.core.Vertx;
+
+abstract class AbstractAppServicesClientImpl {
```
robpblake, line 13: @kostola I personally find it quite hard to follow the intention of this class and would feel nervous about breaking it if I needed to refactor it. 

Some comments in the code about your general approach here would be useful IMHO.

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/AbstractAppServicesClientImpl.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    private <A> Object executeAndHandleException(A apiInstance, ApiVoidExecutor<A> executor) {
+        return executeAndHandleException(apiInstance, api -> {
+            executor.execute(apiInstance);
+            return null;
+        });
+    }
+
+    protected interface ApiValueExecutor<A, T> {
```
robpblake, line 92: Can you provide a comment as to what these interfaces do?

rhoas-client/pom.xml
kostola: Done :+1:

rhoas-client/src/test/java/com/redhat/service/bridge/rhoas/RhoasServiceTest.java
kostola: These `printStackTrace` are not related to test failure / success. I removed them anyway because they were only cluttering the test output.

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/KafkaInstanceAdminClientImpl.java
kostola: Let's refactor this in a dedicated PR once we agree on a final refresh mechanism to overcome Quarkus issues.

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/KafkasMgmtV1ClientImpl.java
kostola: @robpblake I added a timeout of 30 seconds similar to the one in [ManagerSyncServiceImpl](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/9e6430b2a0493a209e1a1789eee30dbc8c2a4185/shard-operator/src/main/java/com/redhat/service/bridge/shard/operator/ManagerSyncServiceImpl.java#L38).

As for the [other token instance](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/196#discussion_r778630573), let's refactor this in a dedicated PR once we agree on a final refresh mechanism to overcome Quarkus issues.

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/RhoasClientImpl.java
kostola: Done :+1:

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/KafkasMgmtV1ClientImpl.java
kostola: Good catch! Fixed :+1:

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/AbstractAppServicesClientImpl.java
kostola: I recognize it's a bit complex. I commented the whole class now. :+1:

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/AbstractAppServicesClientImpl.java
kostola: Done :+1:

rhoas-client/src/test/java/com/redhat/service/bridge/rhoas/RhoasAPI.java
kostola: Added comment also in the code :+1:

rhoas-client/src/main/resources/application.properties
kostola: I removed the default values and listed all the configuration variables in the `README` file :+1:

rhoas-client/src/main/resources/application.properties
kostola: @robpblake Quarkus already translates env variables to properties if formatted correctly. I put all the property / env variables pair in the README file now, removing the default values.

rhoas-client/src/main/resources/application.properties
kostola: I removed the default values and listed all the configuration variables in the `README` file :+1:

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/dto/TopicAndServiceAccountResponse.java
kostola: This is part of `RhoasClient`, which is an aggregated client for our specific needs, so I don't expect this to be part of the SDK. However, I refactored this to use the SDK DTOs instead of plain fields, as we discussed in a call.

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/dto/TopicAndServiceAccountRequest.java
kostola: This is part of RhoasClient, which is an aggregated client for our specific needs, so I don't expect this to be part of the SDK.

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/exceptions/RhoasClientException.java
kostola: @robpblake I can't find the exception you mentioned. It seems like it was removed in some refactoring. Am I wrong?

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/exceptions/AppServicesException.java
kostola: @robpblake I can't find the exception you mentioned. It seems like it was removed in some refactoring. Am I wrong?

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/RhoasClient.java
kostola: I integrated it in the manager inside `BridgesServiceImpl`. Take a look :nerd_face:

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/dto/TopicAndServiceAccountRequest.java
kostola: I integrated `RhoasClient` in the manager inside `BridgesServiceImpl`. Take a look :nerd_face:

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/KafkasMgmtV1ClientImpl.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    protected String getAccessToken() {
+        if (tokens == null || tokens.isAccessTokenExpired()) {
+            try {
+                tokens = client.refreshTokens(refreshToken).await().atMost(SSO_CONNECTION_TIMEOUT);
+            } catch (RuntimeException e) {
+                LOG.warn("Could not fetch a new authentication token from red-hat-sso server");
+                throw e;
```
robpblake, line 54: @kostola Please wrap the `RuntimeException` as the `cause` of a `InternalPlatformException` and set the `message` to the log statement and re-throw 👍

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/exceptions/AppServicesException.java
robpblake: @kostola You're right. It was refactored as part of #183 by @r00ta . `InternalPlatformException` is the one to use now 👍

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/exceptions/RhoasClientException.java
robpblake: @kostola It's now called `InternalPlatformException` after refactor in #183

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
```diff
@@ -54,6 +70,22 @@ public Bridge createBridge(String customerId, BridgeRequest bridgeRequest) {
         bridge.setSubmittedAt(ZonedDateTime.now(ZoneOffset.UTC));
         bridge.setCustomerId(customerId);
         bridgeDAO.persist(bridge);
+
+        if (rhoasEnabled) {
```
robpblake, line 74: @kostola Under what circumstances would we want to disable the creation of the Topic for an OpenBridge Instance?

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
```diff
@@ -54,6 +70,22 @@ public Bridge createBridge(String customerId, BridgeRequest bridgeRequest) {
...
+            String topicName = String.format("ob-%s", bridge.getId());
+            String serviceAccountName = String.format("ob-%s-consumer", bridge.getId());
+            TopicAndServiceAccountRequest request = new TopicAndServiceAccountRequest(topicName, serviceAccountName);
+            try {
+                rhoasClient.createTopicAndConsumerServiceAccount(request).await().atMost(Duration.ofSeconds(rhoasTimeout));
```
robpblake, line 79: @kostola I don't think the user of the `RhoasClient` should have to think about the timeout, otherwise we have to duplicate this logic wherever we use it.

I think it makes more sense that the default timeout is a concern of the `RhoasClient` and it does the waiting on behalf of the user. This means we have centralised control over timeouts when working with other services.

The user then expects the `RhoasClient` to return either the successful `TopicAndServiceAccountResponse` or throw an `InternalPlatformException` if something went wrong. That would mean a signature on the RhoasClient of something like:

```
TopicAndServiceAccountResponse createTopicAndConsumerServiceAccount(TopicAndServiceAccountRequest r);
```

rhoas-client/src/main/resources/application.properties
```diff
@@ -0,0 +1,13 @@
+### OIDC Clients for RHOAS
+quarkus.oidc-client.client-enabled=false
+# Red Hat SSO (refresh token read from "event-bridge.rhoas.sso.red-hat.refresh-token" property)
+quarkus.oidc-client.red-hat-sso.auth-server-url=${event-bridge.rhoas.sso.red-hat.auth-server-url}
+quarkus.oidc-client.red-hat-sso.client-id=cloud-services
```
robpblake, line 5: @kostola These should be injected as environment parameters and not hardcoded

rhoas-client/src/main/resources/application.properties
```diff
@@ -0,0 +1,13 @@
...
+quarkus.oidc-client.red-hat-sso.credentials.secret=secret
+quarkus.oidc-client.red-hat-sso.grant.type=refresh
+# MAS SSO
+quarkus.oidc-client.mas-sso.auth-server-url=${event-bridge.rhoas.sso.mas.auth-server-url}
```
robpblake, line 9: @kostola Can you make the ENV variable naming consistent with what we already have please. Current approach is UPPERCASE with words separated by `_` e.g `RHOAS_SSO_MAS_AUTH_SERVER_URL`

Take a look at `manager/src/main/resources/application.properties` for examples 👍

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
```diff
@@ -24,14 +27,24 @@
...
 @ApplicationScoped
 public class BridgesServiceImpl implements BridgesService {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(BridgesServiceImpl.class);
+    private static final String DEFAULT_RHOAS_TIMEOUT = "60";
+
+    @ConfigProperty(name = "event-bridge.manager.rhoas.enabled", defaultValue = "false")
+    boolean rhoasEnabled;
+
+    @ConfigProperty(name = "event-bridge.manager.rhoas.timeout", defaultValue = DEFAULT_RHOAS_TIMEOUT)
```
robpblake, line 46: @kostola Can we make the unit of the timeout explicit in the config variable e.g `event-bridge.manager.rhoas.timeout-seconds`. I think we can also put the default into the `application.properties`, so we either set it via ENV variable, or default to something sensible e.g:

```
event-bridge.manager.rhoas.timeout-seconds=${RHOAS_TIMEOUT_SECONDS:60}

**Review by robpblake — COMMENTED** [review]
@kostola Thank you. This is looking good.

Some comments from me. We also need to add the required configuration to our `kustomize` overlays so that the Fleet Manager continues to work when we deploy this to our demo environment. Let me know if you need any help with that.

File: manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java (hunk 21 above)
Comment (kostola) [reply manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java:74]: In my idea, I made it opt-in via a boolean flag property because, since the other tasks regarding interaction with Managed Services are not yet fully implemented (e.g. the deletion of the topic and service account when a bridge is deleted), I assumed that it's better for who handles the deployment to be aware that this (currently incomplete) feature must be explicitly enabled.
Once the whole integration will be completed, we can remove the opt-in parameter.
@robpblake what do you think?

File: manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java (hunk 22 above)
Comment (kostola) [reply manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java:79]: In this way the `RhoasClient` would become blocking, losing the advantage of the async implementation. Are we going to stick with thread-per-request model for handling HTTP request?

File: manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java (hunk 21 above)
Comment (robpblake) [reply manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java:74]: @kostola OK, thanks for the explanation. I'm fine with that, but can you please add another JIRA issue to the epic to ensure we remove the feature flag when this work is complete. Worst case is that we leave the property hanging around.

Perhaps also a naming strategy to indicate it's a feature flag e.g `event-bridge.feature-flags.rhoas-enabled=true` and a brief addition to the README so that everyone knows? Might be worth discussing this briefly with the team so that everyone can be aware of this is how we feature flag things on/off.

File: rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/exceptions/RhoasClientException.java (hunk 12 above)
Comment (kostola) [reply rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/exceptions/RhoasClientException.java]: Done :+1:

File: rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/exceptions/AppServicesException.java (hunk 13 above)
Comment (kostola) [reply rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/exceptions/AppServicesException.java]: Done :+1:

File: rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/KafkasMgmtV1ClientImpl.java (hunk 20 above)
Comment (kostola) [reply rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/KafkasMgmtV1ClientImpl.java]: Done :+1:

File: rhoas-client/src/main/resources/application.properties (hunk 24 above)
Comment (kostola) [reply rhoas-client/src/main/resources/application.properties:9]: This is already working thanks to [MicroProfile Config mapping between env vars and properties](https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/configsources.asciidoc#environment-variables-mapping-rules).
Since the mapping is only from env vars to properties and not backwards, I preferred this syntax because it allows both methods to be used, while explicitly using env vars name prevents properties to be used.
@robpblake what do you think?

File: rhoas-client/src/main/resources/application.properties (hunk 23 above)
Comment (kostola) [reply rhoas-client/src/main/resources/application.properties:5]: I set this as a fixed value because it's what the production Red Hat SSO auth flow expects, but it can be overridden either with the property or the env var.
Should I define a dedicated `event-bridge.rhoas.sso.red-hat.client-id` property for this? Is the property name OK to you?

File: manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java (hunk 25 above)
Comment (kostola) [reply manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java:46]: Ok but I'd like to discuss the naming first (see [my other comment](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/196#discussion_r778971192)).

File: manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java (hunk 22 above)
Comment (robpblake) [reply manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java:79]: @kostola OK, let's leave it as it is for now, but can you please set the `CompletionException` and `TimeoutException` as the `cause` on an  `InternalServiceException` with the log statement as the message. At the minute we don't have a non-blocking use-case on the client (we need the topic + service account to be created before the Shard starts to try and deploy the OpenBridge instance), but I think that may change as we progress the implementation.

What we need to think about is "what happens if the Fleet Manager crashes halfway through creating the topic". How would we recover that scenario and continue with the topic creation in an idempotent manner? Lets get this PR resolved and merged, and then we can start to consider how we approach that problem at a later date.

Comment (kostola) [reply manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java:79]: Done :+1:

File: rhoas-client/src/main/resources/application.properties (hunk 24 above)
Comment (robpblake) [reply rhoas-client/src/main/resources/application.properties:9]: @kostola Thank you for the link. That's useful to know and your suggestion is the right way IMHO.

When we make decisions like this, it's important that we're consistent across the project. Making the change in a PR only for a small subset of the properties and then not communicating it as an implementation choice to the rest of the team means we end up in a position where we have multiple approaches to the same thing and people won't follow the same approach as you in the future. I've put a proposal in our Google Chat room about how we capture these decisions.

In the meantime, could you please create another JIRA task to make all properties follow this approach.

File: manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java (hunk 25 above)
Comment (robpblake) [reply manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java:46]: Replied.

File: manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java (hunk 21 above)
Comment (robpblake) [reply manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java:74]: @kostola Feature flagging like this would also be part of the proposal that I've made in our Google chat room.

File: rhoas-client/src/main/resources/application.properties (hunk 23 above)
Comment (robpblake) [reply rhoas-client/src/main/resources/application.properties:5]: Variable name is fine for me. Given that this is sensitive information, it should be encrypted at rest in our Git repo. Take a look at https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/tree/main/kustomize#sealed-secrets on how to do that.

If you need any help, just give me a shout.

File: manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java (hunk 21 above)
Comment (kostola) [reply manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java:74]: I renamed the property and added a paragraph regarding feature flags to the main `README.md` :+1:

File: rhoas-client/src/main/resources/application.properties (hunk 24 above)
Comment (kostola) [reply rhoas-client/src/main/resources/application.properties:9]: I created the JIRA task: https://issues.redhat.com/browse/MGDOBR-215

File: rhoas-client/src/main/resources/application.properties (hunk 23 above)
Comment (kostola) [reply rhoas-client/src/main/resources/application.properties]: I defined the variable and added the secrets :+1:

File: manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java (hunk 25 above)
Comment (kostola) [reply manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java]: Done :+1:

**Review by danielezonca — APPROVED** [review]
Well done :)

File: build-parent/pom.xml [hunk 26]
```diff
@@ -50,6 +50,8 @@
     <version.io.quarkiverse.operatorsdk>2.0.1</version.io.quarkiverse.operatorsdk>
     <version.io.github.classgraph>4.8.129</version.io.github.classgraph>
     <version.com.redhat.connector-management-sdk>0.11.0</version.com.redhat.connector-management-sdk>
+    <version.com.redhat.cloud.kafka-instance-sdk>0.11.0</version.com.redhat.cloud.kafka-instance-sdk>
+    <version.com.redhat.cloud.kafka-management-sdk>0.11.0</version.com.redhat.cloud.kafka-management-sdk>
```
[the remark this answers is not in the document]
robpblake: Yep. Good catch 👍

robpblake (approved):
@kostola Nice work. Great determination to see this one over the line. One small comment from @danielezonca about unifying dependency versions.

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/KafkasMgmtV1ClientImpl.java
```diff
@@ -0,0 +1,66 @@
...
+import io.quarkus.oidc.client.NamedOidcClient;
+import io.quarkus.oidc.client.OidcClient;
+import io.quarkus.oidc.client.Tokens;
+import io.smallrye.mutiny.Uni;
+
+@ApplicationScoped
+public class KafkasMgmtV1ClientImpl extends AbstractAppServicesClientImpl implements KafkasMgmtV1Client {
+
+    private static final Logger LOG = LoggerFactory.getLogger(KafkasMgmtV1ClientImpl.class);
+    private static final Duration SSO_CONNECTION_TIMEOUT = Duration.ofSeconds(30);
```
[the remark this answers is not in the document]
robpblake: @danielezonca @kostola Might be easier to wait for the work that @r00ta is proposing in #198 as I suspect that will change things.

build-parent/pom.xml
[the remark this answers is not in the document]
kostola: Done :+1:

rhoas-client/src/test/resources/application.properties
```diff
@@ -0,0 +1,6 @@
+quarkus.log.category."io.quarkus.oidc.client.runtime.OidcClientImpl".level=TRACE
+quarkus.log.category."io.quarkus.oidc.client.runtime.OidcClientImpl".min-level=TRACE
+quarkus.log.category."io.quarkus.oidc.client.runtime.OidcClientRecorder".level=TRACE
+quarkus.log.category."io.quarkus.oidc.client.runtime.OidcClientRecorder".min-level=TRACE
+quarkus.log.category."com.redhat.service.bridge".level=TRACE
+quarkus.log.category."com.redhat.service.bridge".min-level=TRACE
```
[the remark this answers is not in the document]
kostola: Good catch. I removed it :+1:

rhoas-client/src/main/java/com/redhat/service/bridge/rhoas/KafkasMgmtV1ClientImpl.java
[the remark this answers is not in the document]
kostola: yes let's make everything uniform once #198 is finalized :+1:

PR #239, opened 2022-01-19
MGDOBR-112: add shard and webhook-robot role and allow only shards to access internal api of the manager

r00ta:
[MGDOBR-112](https://issues.redhat.com/browse/MGDOBR-112) 


infra/src/main/java/com/redhat/service/bridge/infra/api/APIConstants.java
```diff
@@ -20,6 +20,11 @@
      */
     public static final String SHARD_API_BASE_PATH = "/api/v1/shard/bridges/";
 
+    /**
+     * Shard realm role.
+     */
+    public static final String SHARD_REALM_ROLE = "shard";
```
robpblake, line 26: @r00ta Can I ask you to raise this at the security workstream tomorrow. IIRC the pattern of defining roles for a Shard is going to be dropped in favour of another approach of managing which Service Accounts are Shards and should have access to the API.

robpblake (commented):
@r00ta All looks good, apart from the open question of whether or not we should implement a Role for the Shard.

sutaakar (approved):
Looks good to me

manager/src/main/java/com/redhat/service/bridge/manager/api/internal/ShardBridgesSyncAPI.java
```diff
@@ -70,8 +85,16 @@ public Response getBridges() {
 
     @PUT
     public Response updateBridge(BridgeDTO dto) {
+        failIfNotAuthorized(identity);
         LOGGER.info("Shard wants to update the Bridge with id '{}' with the status '{}'", dto.getId(), dto.getStatus());
         bridgesService.updateBridge(dto);
         return Response.ok().build();
     }
+
+    private void failIfNotAuthorized(SecurityIdentity identity) {
```
robpblake, line 94: @r00ta On the next iteration of this work, perhaps we can consider a `Filter` and a CDI annotation to bind the `Filter` to this resource: https://quarkus.io/specs/jaxrs/2.1/index.html#Name_Binding

That would mean that we could annotate this Resource (or other resources) that need a valid Shard identity with something like `@ValidShard` and then we simply encapsulate the logic of checking the shard is valid in the `Filter` we implement. That would align our implementation with other conventions e.g `@Authenticated`

We don't need to do that in this PR, but might be a nice approach if we need to extend the ways in which we check a Shard is valid.

WDYT?

robpblake (approved):
LGTM! 🚀

manager/src/main/java/com/redhat/service/bridge/manager/api/internal/ShardBridgesSyncAPI.java
r00ta: https://issues.redhat.com/browse/MGDOBR-245

sutaakar (approved):

PR #242, opened 2022-01-19
MGDOBR-173 MGDOBR-234: use only two service accounts for MK integration in Fleet Manager

kostola:
[MGDOBR-173](https://issues.redhat.com/browse/MGDOBR-173) 
[MGDOBR-234](https://issues.redhat.com/browse/MGDOBR-234) 


lucamolteni (approved):

rhoas-client/src/test/java/com/redhat/service/bridge/rhoas/RhoasAPI.java
```diff
@@ -29,8 +30,25 @@
 
     @POST
     @Path("topic")
-    public TopicAndServiceAccountResponse createTopicAndConsumerServiceAccount(TopicAndServiceAccountRequest request) {
-        return rhoasClient.get().createTopicAndConsumerServiceAccount(request).await().atMost(Duration.ofSeconds(60));
+    public Topic createTopicAndConsumerServiceAccount(Request request) {
```
r00ta, line 33: Can you clarify why we need this api in the `test` project? I see 
```
/**
 * This dummy test API simulates a real usage of the {@link RhoasClient} in a real JAX-RS bean.
 * It allows {@link RhoasAPITest} to spot errors caused by the client mistakenly blocking the
 * Vert.x event loop thread, which can't be spotted testing directly the client in a QuarkusTest
 * (because blocking the test's main thread doesn't cause any error: it simply blocks).
 */
 ```
 but I do not see any test using such exception

Comment (kostola) [reply rhoas-client/src/test/java/com/redhat/service/bridge/rhoas/RhoasAPI.java:33]: This API class is used by `RhoasAPITest` and was introduced in #196. Its purpose is to spot if the `RhoasClient` blocks the Vert.x event loop thread. There's no explicit assert on any explicit exception, the test would simply fail if a block occurs.

Does this answer your question?

File: manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java [hunk 2]
```diff
@@ -181,23 +187,47 @@ public BridgeResponse toResponse(Bridge bridge) {
...
+            throw new InternalPlatformException(msg, e);
+        } catch (TimeoutException e) {
+            String msg = "Timeout deleting topic and revoking access for bridge " + bridgeId;
+            LOGGER.warn(msg, e);
+            throw new InternalPlatformException(msg, e);
+        }
+    }
+
+    private String topicNameFor(String bridgeId) {
+        return String.format("ob-%s", bridgeId);
```
robpblake, line 231: @kostola Does this need to be stored on the `Bridge` entity when we persist it to the DB? We need to be able to communicate the details of the Topic chosen to the Shard as part of https://issues.redhat.com/browse/MGDOBR-171

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
```diff
@@ -181,23 +187,47 @@ public BridgeResponse toResponse(Bridge bridge) {
...
+        try {
+            rhoasClient.deleteTopicAndRevokeAccess(topicNameFor(bridgeId), rhoasOpsAccountClientId, RhoasTopicAccessType.CONSUMER_AND_PRODUCER)
+                    .await().atMost(Duration.ofSeconds(rhoasTimeout));
+        } catch (CompletionException e) {
+            String msg = "Failed deleting topic and revoking access for bridge " + bridgeId;
+            LOGGER.warn(msg, e);
+            throw new InternalPlatformException(msg, e);
+        } catch (TimeoutException e) {
+            String msg = "Timeout deleting topic and revoking access for bridge " + bridgeId;
+            LOGGER.warn(msg, e);
```
robpblake, line 225: @kostola Let's just throw the `Exception` and not `LOGGER.warn` the same message. We'll end up with duplicate stack traces in the logs as `InternalPlatformExceptionMapper` additionally logs the error.

robpblake (commented):
@kostola Thanks. This is looking good.

We need to generate the service accounts to be used in our demo environment and ensure that they are included (encrypted of course) in our `kustomize` templates.

rhoas-client/src/test/java/com/redhat/service/bridge/rhoas/RhoasAPI.java
r00ta: Uhm I'm not sure I get the point. Is it related to https://github.com/smallrye/smallrye-mutiny/issues/821 ?

kostola: You're actually right: this API and the corresponding test are useless implemented this way.

My original idea was to implement the API using `quarkus-resteasy-reactive` and return `Uni<Topic>`, so that if, by mistake, a blocking piece of code went into the `RhoasClient` method, the test would fail with the following exception:
```
java.lang.IllegalStateException: The current thread cannot be blocked: vert.x-eventloop-thread-XYZ
```
These exceptions don't happen in normal tests because the main test thread can be blocked without issues.

Probably I changed the implementation when I realized that currently we never use `quarkus-resteasy-reactive` in our APIs and I didn't realize the test now lost its original purpose.

@danielezonca @r00ta @robpblake should I remove this? Or reimplement it with `quarkus-resteasy-reactive`?

r00ta: @kostola I'm fine to remove them if we are already covering the methods in other tests

robpblake: @kostola Fine for me to remove it, particularly when you say

> this API and the corresponding test are useless implemented this way.

I would hope that as we start to expand our integration tests, we would pick up issues in our clients that interact with external services rather than having to test this sort of functionality each time we write something reactive.

How do you feel about that?

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
kostola: I discussed this with @r00ta, who is going to implement MGDOBR-171, and we agreed that it's not needed to store this yet because it can be retrieved from the bridge id.

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
kostola: Done :+1:

rhoas-client/src/test/java/com/redhat/service/bridge/rhoas/RhoasAPI.java
kostola: Removed :+1:

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
robpblake: @r00ta @kostola Just to be sure: The Manager will still expose the topic name to the Shard via our `ShardBridgesSyncAPI`, but because the topic name can be inferred from `ob- + $bridge.getId()`, we're saying we don't need to persist it in the DB.

Am I correct?

kostola: afaik you're correct @robpblake :+1:

robpblake (approved):

r00ta (approved):

PR #243, opened 2022-01-19
MGDOBR-240: update webhook action

r00ta:

### DIFF
--- a/actions/src/main/java/com/redhat/service/bridge/actions/webhook/WebhookAction.java
+++ b/actions/src/main/java/com/redhat/service/bridge/actions/webhook/WebhookAction.java
@@ -7,7 +7,9 @@ import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
 import com.redhat.service.bridge.actions.ActionInvoker;
+import com.redhat.service.bridge.actions.GlobalConfig;
 import com.redhat.service.bridge.actions.InvokableActionProvider;
+import com.redhat.service.bridge.infra.exceptions.definitions.platform.TechnicalBearerTokenNotConfiguredException;
 import com.redhat.service.bridge.infra.exceptions.definitions.user.ActionProviderException;
 import com.redhat.service.bridge.infra.models.actions.BaseAction;
 import com.redhat.service.bridge.infra.models.dto.ProcessorDTO;
@@ -20,6 +22,7 @@ public class WebhookAction implements InvokableActionProvider {
 
     public static final String TYPE = "Webhook";
     public static final String ENDPOINT_PARAM = "endpoint";
+    public static final String USE_TECHNICAL_BEARER_TOKEN = "useTechincalBearerToken";
 
     private WebClient client;
 
@@ -45,9 +48,15 @@ public class WebhookAction implements InvokableActionProvider {
     }
 
     @Override
-    public ActionInvoker getActionInvoker(ProcessorDTO processor, BaseAction baseAction) {
+    public ActionInvoker getActionInvoker(ProcessorDTO processor, BaseAction baseAction, GlobalConfig globalConfig) {
         String endpoint = Optional.ofNullable(baseAction.getParameters().get(ENDPOINT_PARAM))
                 .orElseThrow(() -> buildNoEndpointException(processor));
+        if (baseAction.getParameters().containsKey(USE_TECHNICAL_BEARER_TOKEN) && baseAction.getParameters().get(USE_TECHNICAL_BEARER_TOKEN).equals("true")) {
+            if (globalConfig.getWebhookTechnicalBearerToken() == null) {
+                throw new TechnicalBearerTokenNotConfiguredException("A webhook action needed the technical bearer token but it was not configured.");
+            }
+            return new WebhookInvoker(endpoint, client, globalConfig.getWebhookTechnicalBearerToken());
+        }
         return new WebhookInvoker(endpoint, client);
     }
 
--- a/executor/src/main/java/com/redhat/service/bridge/executor/GlobalConfigProvider.java
+++ b/executor/src/main/java/com/redhat/service/bridge/executor/GlobalConfigProvider.java
@@ -1,0 +1,24 @@
+package com.redhat.service.bridge.executor;
+
+import java.util.Optional;
+
+import javax.enterprise.context.Dependent;
+import javax.ws.rs.Produces;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import com.redhat.service.bridge.actions.GlobalConfig;
+
+@Dependent
+public class GlobalConfigProvider {
+
+    @ConfigProperty(name = "event-bridge.webhook.technical-bearer-token")
+    Optional<String> webhookTechnicalBearerToken;
+
+    @Produces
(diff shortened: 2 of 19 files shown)


manager/src/test/java/com/redhat/service/bridge/manager/actions/sendtobridge/SendToBridgeActionTransformerTest.java
```diff
@@ -129,7 +129,9 @@ private void assertValid(BaseAction transformedAction, String expectedName, Stri
         assertThat(transformedAction.getType()).isEqualTo(WebhookAction.TYPE);
         assertThat(transformedAction.getName()).isEqualTo(expectedName);
         assertThat(transformedAction.getParameters()).containsEntry(WebhookAction.ENDPOINT_PARAM, expectedEndpoint);
-        assertThat(actionProviderFactory.getActionProvider(transformedAction.getType()).getParameterValidator().isValid(transformedAction).isValid()).isTrue();
+
+        // TODO: Need to disable this, unless we go for a bigger refactoring
```
danielezonca, line 133: Can you clarify?

executor/src/test/java/com/redhat/service/bridge/executor/ExecutorServiceTest.java
```diff
@@ -9,6 +9,7 @@
 import org.junit.jupiter.api.Test;
 import org.mockito.ArgumentCaptor;
 
+import com.redhat.service.bridge.actions.GlobalConfig;
```
danielezonca, line 12: Unused import?

executor/src/main/java/com/redhat/service/bridge/executor/GlobalConfigProvider.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.Optional;
+
+import javax.enterprise.context.Dependent;
+import javax.ws.rs.Produces;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import com.redhat.service.bridge.actions.GlobalConfig;
+
+@Dependent
```
danielezonca, line 12: Why dependent? Isn't this a single instance for the whole application?

actions/src/main/java/com/redhat/service/bridge/actions/webhook/WebhookAction.java
```diff
@@ -45,9 +48,15 @@ public WebhookActionValidator getParameterValidator() {
     }
 
     @Override
-    public ActionInvoker getActionInvoker(ProcessorDTO processor, BaseAction baseAction) {
+    public ActionInvoker getActionInvoker(ProcessorDTO processor, BaseAction baseAction, GlobalConfig globalConfig) {
```
robpblake, line 51: @r00ta Seeing as this `GlobalConfig` only applies to the WebHook Action, can't we inject it into this class as a bean and then pass it as a constructor param to the `WebHookInvoker`.

That would avoid us updating the method signatures for all our interfaces.

executor/src/main/java/com/redhat/service/bridge/executor/GlobalConfigProvider.java
r00ta: I'm following the reference implementation from quarkus guide https://quarkus.io/guides/cdi-reference#default_beans

actions/src/main/java/com/redhat/service/bridge/actions/webhook/WebhookAction.java
r00ta: @robpblake that would work if we refactor the `actions` module to properly separate the actions definition and the implementations. The `GlobalConfig` bean is supposed to be produced by the `executor` application, and the `actions` module is being used also by the `manager` that does not provide such `bean`. 
The design problem I see here is that the manager should not include the implementation of the actions, because for example since day 0 the manager is bringing in the kafka client from the `sendToKafka` action..

kostola:
Given that this task is specific to the Webhook action, I think it could be simplified by injecting the technical token property:
```
@ConfigProperty(name = "event-bridge.webhook.technical-bearer-token")
Optional<String> webhookTechnicalBearerToken;
```
directly inside the `WebhooAction` bean, which then would pass it to the `WebhookInvoker` instances via constructor (as suggested also by @robpblake [here](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/243#discussion_r787904766).

I'm not saying the current implementation is wrong, but this would allow to get the fix out for the forthcoming release and give us more time afterwards to properly design a generic configuration logic for all actions.

Wdyt @r00ta @danielezonca ?

r00ta:
> Given that this task is specific to the Webhook action, I think it could be simplified by injecting the technical token property:
> 
> ```
> @ConfigProperty(name = "event-bridge.webhook.technical-bearer-token")
> Optional<String> webhookTechnicalBearerToken;
> ```
> 
> directly inside the `WebhooAction` bean, which then would pass it to the `WebhookInvoker` instances via constructor (as suggested also by @robpblake [here](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/243#discussion_r787904766).
> 
> I'm not saying the current implementation is wrong, but this would allow to get the fix out for the forthcoming release and give us more time afterwards to properly design a generic configuration logic for all actions.
> 
> Wdyt @r00ta @danielezonca ?


> Given that this task is specific to the Webhook action, I think it could be simplified by injecting the technical token property:
> 
> ```
> @ConfigProperty(name = "event-bridge.webhook.technical-bearer-token")
> Optional<String> webhookTechnicalBearerToken;
> ```
> 
> directly inside the `WebhooAction` bean, which then would pass it to the `WebhookInvoker` instances via constructor (as suggested also by @robpblake [here](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/243#discussion_r787904766).
> 
> I'm not saying the current implementation is wrong, but this would allow to get the fix out for the forthcoming release and give us more time afterwards to properly design a generic configuration logic for all actions.
> 
> Wdyt @r00ta @danielezonca ?

see https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/243#discussion_r788531480

actions/src/main/java/com/redhat/service/bridge/actions/webhook/WebhookAction.java
r00ta: If we move away from `GlobalConfig` and we inject directly only the optional property it would work I think. Let me try

manager/src/test/java/com/redhat/service/bridge/manager/actions/sendtobridge/SendToBridgeActionTransformerTest.java
danielezonca: Ok thanks for the clarification, can you please create a follow-up ticket?

actions/src/main/java/com/redhat/service/bridge/actions/webhook/WebhookAction.java
r00ta: Done

danielezonca (approved):

robpblake (approved):

PR #246, opened 2022-01-20
MGDOBR-199 MC: Avoid keeping the DB transaction open when calling the create connectors API

lucamolteni:
[MGDOBR-199](https://issues.redhat.com/browse/MGDOBR-199) 

This PR is just to avoid keeping the DB transaction open and doesn't offer a full reconcile loop for stale resources

### DIFF
--- a/infra/src/main/java/com/redhat/service/bridge/infra/models/dto/ConnectorStatus.java
+++ b/infra/src/main/java/com/redhat/service/bridge/infra/models/dto/ConnectorStatus.java
@@ -3,10 +3,16 @@ package com.redhat.service.bridge.infra.models.dto;
 // TODO-MC review statuses
 // https://issues.redhat.com/browse/MGDOBR-195
 public enum ConnectorStatus {
-    REQUESTED,
-    PROVISIONING,
-    AVAILABLE,
-    DELETION_REQUESTED,
-    DELETED,
-    FAILED
+    ACCEPTED("accepted"),
+    TOPIC_CREATED("topic_created"),
+    READY("ready"), // connector and topic have been created
+    TOPIC_DELETED("topic_deleted"),
+    DELETED("deleted"), // connector and topic have been created, used only as Desired status as it's physically deleted eventually
+    FAILED("failed");
+
+    String status;
+
+    ConnectorStatus(String status) {
+        this.status = status;
+    }
 }
--- a/manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
+++ b/manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
@@ -3,6 +3,7 @@ package com.redhat.service.bridge.manager;
 import java.time.ZonedDateTime;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 
 import javax.enterprise.context.ApplicationScoped;
@@ -31,13 +32,18 @@ import com.redhat.service.bridge.infra.models.processors.ProcessorDefinition;
 import com.redhat.service.bridge.manager.api.models.requests.ProcessorRequest;
 import com.redhat.service.bridge.manager.api.models.responses.ProcessorResponse;
 import com.redhat.service.bridge.manager.connectors.ConnectorsService;
+import com.redhat.service.bridge.manager.connectors.Events;
 import com.redhat.service.bridge.manager.dao.ProcessorDAO;
 import com.redhat.service.bridge.manager.models.Bridge;
+import com.redhat.service.bridge.manager.models.ConnectorEntity;
 import com.redhat.service.bridge.manager.models.Processor;
 import com.redhat.service.bridge.manager.providers.InternalKafkaConfigurationProvider;
 
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
+import io.vertx.mutiny.core.eventbus.EventBus;
+
+import static com.redhat.service.bridge.manager.connectors.Events.CONNECTOR_CREATED_EVENT;
 
 @ApplicationScoped
 public class ProcessorServiceImpl implements ProcessorService {
@@ -65,6 +71,9 @@ public class ProcessorServiceImpl implements ProcessorService {
     InternalKafkaConfigurationProvider internalKafkaConfigurationProvider;
 
     @Inject
+    EventBus eventBus;
+
+    @Inject
     ShardService shardService;
 
     @Transactional
@@ -80,7 +89,6 @@ public class ProcessorServiceImpl implements ProcessorService {
         return processor;
     }
 
-    @Transactional
     @Override
     public Processor createProcessor(String bridgeId, String customerId, ProcessorRequest processorRequest) {
         /* We cannot deploy Processors to a Bridge that is not Available */
@@ -112,17 +120,23 @@ public class ProcessorServiceImpl implements ProcessorService {
         ProcessorDefinition definition = new ProcessorDefinition(requestedFilters, requestedTransformationTemplate, requestedAction, resolvedAction);
         newProcessor.setDefinition(definitionToJsonNode(definition));
 
-        processorDAO.persist(newProcessor);
+        Optional<ConnectorEntity> persist = createProcessorConnectorEntity(newProcessor, actionProvider, resolvedAction);
 
-        connectorService.createConnectorIfNeeded(resolvedAction, newProcessor, actionProvider);
+        persist.ifPresent(c -> eventBus.requestAndForget(CONNECTOR_CREATED_EVENT, c));
 
         return newProcessor;
     }
 
     @Transactional
+    public Optional<ConnectorEntity> createProcessorConnectorEntity(Processor newProcessor, ActionProvider actionProvider, BaseAction resolvedAction) {
+        processorDAO.persist(newProcessor);
+        return connectorService.createConnectorEntity(resolvedAction, newProcessor, actionProvider);
+    }
+
+    @Transactional
     @Override
-    public List<Processor> getProcessorByStatusesAndShardId(List<BridgeStatus> statuses, String shardId) {
-        return processorDAO.findByStatusesAndShardId(statuses, shardId);
+    public List<Processor> getProcessorByStatusesAndShardIdWithReadyDependencies(List<BridgeStatus> statuses, String shardId) {
+        return processorDAO.findByStatusesAndShardIdWithReadyDependencies(statuses, shardId);
     }
 
     @Transactional
@@ -160,19 +174,24 @@ public class ProcessorServiceImpl implements ProcessorService {
         return processorDAO.findByBridgeIdAndCustomerId(bridge.getId(), bridge.getCustomerId(), queryInfo);
     }
 
-    @Transactional
     @Override
     public void deleteProcessor(String bridgeId, String processorId, String customerId) {
+        List<ConnectorEntity> connectorEntitiesToBeDeleted = updateProcessorConnectorForDeletionOnDB(bridgeId, processorId, customerId);
+        connectorEntitiesToBeDeleted.forEach(c -> {
+            LOGGER.info("Firing deletion event for entity: " + c);
+            eventBus.requestAndForget(Events.CONNECTOR_DELETED_EVENT, c);
+        });
+    }
+
+    @Transactional
+    public List<ConnectorEntity> updateProcessorConnectorForDeletionOnDB(String bridgeId, String processorId, String customerId) {
         Processor processor = processorDAO.findByIdBridgeIdAndCustomerId(processorId, bridgeId, customerId);
-
-        ProcessorDefinition processorDefinition = jsonNodeToDefinition(processor.getDefinition());
-        ActionProvider actionProvider = actionProviderFactory.getActionProvider(processorDefinition.getRequestedAction().getType());
-
-        connectorService.deleteConnectorIfNeeded(processorDefinition.getResolvedAction(), processor, actionProvider);
-
         processor.setStatus(BridgeStatus.DELETION_REQUESTED);
+
         LOGGER.info("Processor with id '{}' for customer '{}' on bridge '{}' has been marked for deletion", processor.getId(), processor.getBridge().getCustomerId(),
                 processor.getBridge().getId());
+
+        return connectorService.deleteConnectorIfNeeded(processor);
     }
 
     @Override
--- a/manager/src/main/java/com/redhat/service/bridge/manager/api/internal/ShardBridgesSyncAPI.java
+++ b/manager/src/main/java/com/redhat/service/bridge/manager/api/internal/ShardBridgesSyncAPI.java
@@ -29,6 +29,8 @@ import com.redhat.service.bridge.infra.models.dto.ProcessorDTO;
 import com.redhat.service.bridge.manager.BridgesService;
 import com.redhat.service.bridge.manager.ProcessorService;
 import com.redhat.service.bridge.manager.ShardService;
+import com.redhat.service.bridge.manager.models.Bridge;
+import com.redhat.service.bridge.manager.models.Processor;
 
 import io.quarkus.security.Authenticated;
 
@@ -69,8 +71,11 @@ public class ShardBridgesSyncAPI {
     public Response updateProcessorStatus(ProcessorDTO processorDTO) {
         String shardId = identityResolver.resolve(jwt);
         failIfNotAuthorized(shardId);
-        LOGGER.info("Processing update from shard for Processor with id '{}' for bridge '{}' for customer '{}'", processorDTO.getId(), processorDTO.getBridgeId(),
-                processorDTO.getCustomerId());
+        LOGGER.info("Processing update from shard for Processor with id '{}' for bridge '{}' for customer '{}' with status '{}'",
+                processorDTO.getId(),
+                processorDTO.getBridgeId(),
+                processorDTO.getCustomerId(),
+                processorDTO.getStatus());
         processorService.updateProcessorStatus(processorDTO);
         return Response.ok().build();
     }
@@ -81,7 +86,9 @@ public class ShardBridgesSyncAPI {
         String shardId = identityResolver.resolve(jwt);
         failIfNotAuthorized(shardId);
         LOGGER.info("Request from Shard for Processors to deploy or delete.");
-        return Response.ok(processorService.getProcessorByStatusesAndShardId(statuses, shardId)
+        List<Processor> processorToDeployOrDelete = processorService.getProcessorByStatusesAndShardIdWithReadyDependencies(statuses, shardId);
+        LOGGER.info("Found {} processor(s) to deploy or delete", processorToDeployOrDelete.size());
+        return Response.ok(processorToDeployOrDelete
                 .stream()
                 .map(processorService::toDTO)
                 .collect(toList()))
@@ -93,7 +100,9 @@ public class ShardBridgesSyncAPI {
         String shardId = identityResolver.resolve(jwt);
         failIfNotAuthorized(shardId);
         LOGGER.info("Shard asks for Bridges to deploy or delete");
-        return Response.ok(bridgesService.getBridgesByStatusesAndShardId(statuses, shardId)
+        List<Bridge> bridgesToDeployOrDelete = bridgesService.getBridgesByStatusesAndShardId(statuses, shardId);
+        LOGGER.info("Found {} bridge(s) to deploy or delete", bridgesToDeployOrDelete.size());
+        return Response.ok(bridgesToDeployOrDelete
                 .stream()
                 .map(bridgesService::toDTO)
                 .collect(toList()))
--- a/manager/src/main/java/com/redhat/service/bridge/manager/connectors/AbstractConnectorWorker.java
+++ b/manager/src/main/java/com/redhat/service/bridge/manager/connectors/AbstractConnectorWorker.java
@@ -1,0 +1,127 @@
+package com.redhat.service.bridge.manager.connectors;
+
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.util.UUID;
+
+import javax.inject.Inject;
+import javax.persistence.EntityManager;
+import javax.transaction.Transactional;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.redhat.service.bridge.manager.models.ConnectorEntity;
+
+import io.vertx.mutiny.core.eventbus.EventBus;
+
+// R is the type returned by the external service
+public abstract class AbstractConnectorWorker<R> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractConnectorWorker.class);
+
+    @Inject
+    protected EntityManager entityManager;
+
+    @Inject
+    protected EventBus eventBus;
+
+    protected abstract R callExternalService(ConnectorEntity connectorEntity);
+
+    // Don't need to implement this if this is a deletion worker
+    protected ConnectorEntity updateEntityForSuccess(ConnectorEntity connectorEntity, R serviceResponse) {
+        return connectorEntity;
+    }
+
+    protected abstract ConnectorEntity updateEntityForError(ConnectorEntity connectorEntity, Throwable error);
+
+    protected void afterSuccessfullyUpdated(ConnectorEntity c) {
+    }
+
+    // Used for workers that should delete the entity after successfully calling the service.
+    // By default, after the service call updateEntityForSuccess is called, with this physical deletion will occur
+    protected boolean shouldDeleteAfterSuccess() {
+        return false;
+    }
+
+    // Could be the k8s pod name https://issues.redhat.com/browse/MGDOBR-336
+    String workerId = UUID.randomUUID().toString();
+
+    // It's important that at the beginning there is no transaction active
+    // in this way we're sure each step in the following method is executed in its own transaction
+    @Transactional(Transactional.TxType.NEVER)
+    public void execute(ConnectorEntity currentEntity) {
+
+        ConnectorEntity connectorEntity;
+        try {
+            connectorEntity = claimConnector(currentEntity);
+            LOGGER.info("Running worker: {} on Entity: {} with workerId: {} ", workerName(), connectorEntity, workerId);
+        } catch (Exception e) {
+            // Cannot claim, return the unmodified entity
+            LOGGER.trace("Worker {} cannot claim Entity: {} with error: {} ", workerId, workerName(), e);
+            return;
+        }
+
+        try {
+            R serviceResponse = callExternalService(connectorEntity);
+
+            if (shouldDeleteAfterSuccess()) { // Used only for workers that physically delete entities
+                deleteEntity(connectorEntity);
+            } else { // Success path, most common
+                ConnectorEntity successConnectorEntity = successfullyCalledService(connectorEntity, serviceResponse);
+                afterSuccessfullyUpdated(successConnectorEntity);
+            }
+        } catch (Exception e) {
+
+            errorWhileCalling(e, connectorEntity);
+        }
+    }
+
+    @Transactional(Transactional.TxType.REQUIRES_NEW)
+    public ConnectorEntity claimConnector(ConnectorEntity currentEntity) {
+        // As this is a new transaction from a non-transactional state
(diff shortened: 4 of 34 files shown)


manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsOrchestrateImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    @Inject
+    ConnectorsDAO connectorsDAO;
+
+    @Inject
+    EventBus bus;
+
+    @Override
+    @Transactional(Transactional.TxType.NEVER)
+    public Uni<List<Message<ConnectorEntity>>> updatePendingConnectors() {
+        // assume this class is not a singleton as we could have multiple pods running this
```
robpblake, line 31: @lucamolteni I assume this is where the conversation around "locking" came in e.g we have two pods, both pick up the need to deploy the connector at the same time and therefore we need to ensure only 1 does the work.

The alternative here is that we use a Kafka Topic rather than our DB. So the need to deploy a Connector is an Event that we place onto a Kafka Topic. All our consumers of that Topic have the same consumer group id so only one of them ever consumes the Event. That would mean that we could have many pods processing work to deploy connectors, and only one of them ever doing the actual DB update when things have completed.

manager/src/main/java/com/redhat/service/bridge/manager/connectors/AbstractConnectorWorker.java
```diff
@@ -0,0 +1,89 @@
...
+
+public abstract class AbstractConnectorWorker<R> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractConnectorWorker.class);
+
+    @Inject
+    EntityManager entityManager;
+
+    // Could be the POD name
+    String workerId = UUID.randomUUID().toString();
```
robpblake, line 25: @lucamolteni Not sure this is going to work. I suspect it will just mean that we end up with orphaned items in the work queue. Imagine the following workflow:

- workerId that is random selects claims item
- workerId does first part of provisioning of connector
- workerId crashes

A new workerId is spawned by K8S. What happens to the item in the queue that was claimed by the previous worker? 

Using the Pod name as the `workerId` doesn't help as they are generated randomly via K8S controller when it creates them for the `Deployment` of the Fleet Manager.

infra/src/main/java/com/redhat/service/bridge/infra/models/dto/ConnectorStatus.java
```diff
@@ -4,9 +4,9 @@
 // https://issues.redhat.com/browse/MGDOBR-195
 public enum ConnectorStatus {
     REQUESTED,
```
r00ta, line 6: You might want to align to the new statuses (from MK) https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/333

manager/src/main/java/com/redhat/service/bridge/manager/api/internal/ShardBridgesSyncAPI.java
```diff
@@ -81,7 +85,9 @@ public Response getProcessors() {
         String shardId = identityResolver.resolve(jwt);
         failIfNotAuthorized(shardId);
         LOGGER.info("Request from Shard for Processors to deploy or delete.");
-        return Response.ok(processorService.getProcessorByStatusesAndShardId(statuses, shardId)
+        List<Processor> processorToDeployOrDelete = processorService.getProcessorByStatusesAndShardId(statuses, shardId);
+        LOGGER.info("Found {} processor(s) to deploy or delete", processorToDeployOrDelete.size());
```
r00ta, line 89: If you log this for Processor you might add it also for the Bridge ingresses

manager/src/main/java/com/redhat/service/bridge/manager/connectors/creation/CreateConnectorWorker.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @ConsumeEvent(value = KAFKA_TOPIC_CREATED_EVENT, blocking = true)
+    public void consume(ConnectorEntity connectorEntity) {
+        execute(connectorEntity);
+    }
+
+    @Override
+    protected Connector callExternalService(ConnectorEntity connectorEntity) {
+        JsonNode payload = connectorEntity.getDefinition();
+        String newConnectorName = connectorEntity.getName();
```
r00ta, line 53: Sorry for the stylish nitpick (feel free to ignore it): you define those variables here but you use them only once and after few lines. Maybe it would be more readable to do 
```java
createConnectorRequest.setName(connectorEntity.getName());
```

instead of 

```java
String newConnectorName = connectorEntity.getName();
String connectorType = connectorEntity.getConnectorType();
ConnectorRequest createConnectorRequest = new ConnectorRequest();
createConnectorRequest.setName(newConnectorName);
```

manager/src/main/java/com/redhat/service/bridge/manager/connectors/creation/CreateConnectorWorker.java
```diff
@@ -0,0 +1,99 @@
...
+    @ConfigProperty(name = "managed-connectors.kafka.client.id")
+    String serviceAccountId;
+
+    @ConfigProperty(name = "managed-connectors.kafka.client.secret")
+    String serviceAccountSecret;
+
+    @Inject
+    ConnectorsApiClient connectorsApiClient;
+
+    @ConsumeEvent(value = KAFKA_TOPIC_CREATED_EVENT, blocking = true)
```
r00ta, line 45: In the other workers you use `Events.CONSTANT` and here you use static import (also in other workers), maybe one single approach would improve readability

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsOrchestrate.java
```diff
@@ -0,0 +1,14 @@
+package com.redhat.service.bridge.manager.connectors;
+
+import java.util.List;
+
+import com.redhat.service.bridge.manager.models.ConnectorEntity;
+
+import io.smallrye.mutiny.Uni;
+import io.vertx.mutiny.core.eventbus.Message;
+
+public interface ConnectorsOrchestrate {
```
r00ta, line 10: `ConnectorsOrchestrator`?

manager/src/main/java/com/redhat/service/bridge/manager/dao/ConnectorsDAO.java
```diff
@@ -20,9 +20,13 @@ public ConnectorEntity findByProcessorIdAndName(String processorId, String name)
         return singleResultFromList(find("#CONNECTORENTITY.findByProcessorIdAndName", p));
     }
 
-    public ConnectorEntity findByProcessorId(String processorId) {
+    public List<ConnectorEntity> findUnprocessed() {
+        return find("#CONNECTORENTITY.findUnprocessed").list();
+    }
+
+    public List<ConnectorEntity> findByProcessorId(String processorId) {
```
r00ta, line 27: Is this going to return always a list with just one item?

manager/src/main/java/com/redhat/service/bridge/manager/connectors/AbstractConnectorWorker.java
```diff
@@ -0,0 +1,126 @@
...
+                deleteEntity(connectorEntity);
+            }
+        } catch (Exception e) {
+
+            errorWhileCalling(e, connectorEntity);
+        }
+    }
+
+    @Transactional(Transactional.TxType.REQUIRES_NEW)
+    public ConnectorEntity claimConnector(ConnectorEntity currentEntity) {
```
r00ta, line 80: I think there is a concurrency here. Think about the following scenario: 
1) Pod A runs `ConnectorsOrchestrator.updatePendingConnectors` and with `ConnectorDAO.findUnprocessed` finds the connector X. It sends to its internal bus the request. 
2) Pod B does the same and finds the connector X. The request is sent to its internal bus as well. 
3) Pod A starts processing and calls this `claimConnector` which sets `mergedCE.setWorkerId(workerId);`
4) Pod B does the same and overrides with its `mergedCE.setWorkerId(workerId);`

In short, I think this `claimConnector` is not really preventing multiple pods to take ownership of the resource since it is not checking that the `workerId` is still `null`.  Or am I missing something?

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsOrchestrateImpl.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Inject
+    ConnectorsDAO connectorsDAO;
+
+    @Inject
+    EventBus eventBus;
+
+    @Override
+    @Transactional(Transactional.TxType.NEVER)
+    public Uni<List<Message<ConnectorEntity>>> updatePendingConnectors() {
```
r00ta, line 37: I think a pending connector might be also a connector that has the `workerId` set but for any reason that worker did not release the lock (it crashed, it was restarted, any reason). Currently `findUprocessed` is not returning such processors

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
```diff
@@ -112,13 +120,19 @@ public Processor createProcessor(String bridgeId, String customerId, ProcessorRe
         ProcessorDefinition definition = new ProcessorDefinition(requestedFilters, requestedTransformationTemplate, requestedAction, resolvedAction);
         newProcessor.setDefinition(definitionToJsonNode(definition));
 
-        processorDAO.persist(newProcessor);
+        Optional<ConnectorEntity> persist = persistProcessorCreateConnector(newProcessor, actionProvider, resolvedAction);
 
-        connectorService.createConnectorIfNeeded(resolvedAction, newProcessor, actionProvider);
+        persist.ifPresent(c -> eventBus.requestAndForget(CONNECTOR_CREATED_EVENT, c));
```
r00ta, line 125: I'm wondering if it's responsability of the `connectorService` to fire the event

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
```diff
@@ -160,19 +174,24 @@ public Long getProcessorsCount(String bridgeId, String customerId) {
...
-        Processor processor = processorDAO.findByIdBridgeIdAndCustomerId(processorId, bridgeId, customerId);
-
-        ProcessorDefinition processorDefinition = jsonNodeToDefinition(processor.getDefinition());
-        ActionProvider actionProvider = actionProviderFactory.getActionProvider(processorDefinition.getRequestedAction().getType());
-
-        connectorService.deleteConnectorIfNeeded(processorDefinition.getResolvedAction(), processor, actionProvider);
+        List<ConnectorEntity> connectorEntitiesToBeDeleted = updateProcessorConnectorForDeletionOnDB(bridgeId, processorId, customerId);
+        connectorEntitiesToBeDeleted.forEach(c -> {
+            LOGGER.info("Firing deletion event for entity: " + c);
+            eventBus.requestAndForget(Events.CONNECTOR_DELETED_EVENT, c);
```
r00ta, line 182: same here

manager/src/main/java/com/redhat/service/bridge/manager/models/Processor.java
```diff
@@ -29,7 +32,14 @@
         @NamedQuery(name = "PROCESSOR.findByBridgeIdAndName",
                 query = "from Processor p where p.name=:name and p.bridge.id=:bridgeId"),
         @NamedQuery(name = "PROCESSOR.findByStatusesAndShardId",
```
r00ta, line 34: I'd say now the signature is more something like `findByStatusesAndShardIdWithReadyDependencies`

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
```diff
@@ -112,13 +120,19 @@ public Processor createProcessor(String bridgeId, String customerId, ProcessorRe
...
-        processorDAO.persist(newProcessor);
+        Optional<ConnectorEntity> persist = persistProcessorCreateConnector(newProcessor, actionProvider, resolvedAction);
 
-        connectorService.createConnectorIfNeeded(resolvedAction, newProcessor, actionProvider);
+        persist.ifPresent(c -> eventBus.requestAndForget(CONNECTOR_CREATED_EVENT, c));
 
         return newProcessor;
     }
 
+    @Transactional
```
robpblake, line 130: I don't think the `@Transactional` annotation works on private methods. Is there any reason why you removed it from the public `createProcessor` method?

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
```diff
@@ -160,19 +174,24 @@ public Long getProcessorsCount(String bridgeId, String customerId) {
...
-        connectorService.deleteConnectorIfNeeded(processorDefinition.getResolvedAction(), processor, actionProvider);
+        List<ConnectorEntity> connectorEntitiesToBeDeleted = updateProcessorConnectorForDeletionOnDB(bridgeId, processorId, customerId);
+        connectorEntitiesToBeDeleted.forEach(c -> {
+            LOGGER.info("Firing deletion event for entity: " + c);
+            eventBus.requestAndForget(Events.CONNECTOR_DELETED_EVENT, c);
+        });
+    }
 
+    @Transactional
+    public List<ConnectorEntity> updateProcessorConnectorForDeletionOnDB(String bridgeId, String processorId, String customerId) {
```
robpblake, line 187: Is there any reason why this needs to be a separate method to `deleteProcessor`?

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
robpblake: I agree with @r00ta comment.

 I feel it would be cleaner for our first implementation to have a `createConnector()` and `deleteConnector` methods on the `ConnectorService` to align with the `ProcessorService` and `BridgeService` interfaces. 
 
 Within the `createConnector` method, if we don't need to create a Connector, it's essentially a null-op. Otherwise we do what's required and fire the event from within the `createConnector` method.

manager/src/main/java/com/redhat/service/bridge/manager/api/internal/ShardBridgesSyncAPI.java
robpblake: If we decide to keep this can we also include the `shardId` in the log statement. At the minute it doesn't tell us for which Shard e.g:

`Found '4' processor(s) to deploy or delete for shard 'foo'`

manager/src/main/java/com/redhat/service/bridge/manager/connectors/AbstractConnectorWorker.java
```diff
@@ -0,0 +1,126 @@
...
+        } catch (Exception e) {
+            // Cannot claim, return the unmodified entity
+            LOGGER.trace("Worker {} cannot claim Entity: {} with error: {} ", workerId, workerName(), e);
+            return;
+        }
+
+        try {
+            R serviceResponse = callExternalService(connectorEntity);
+
+            if (!shouldDeleteAfterSuccess()) {
```
robpblake, line 66: Can we reverse the logic here to make it easier to follow e.g.:

```
if(shouldDeleteAfterSuccess()) {
  deleteEntity();
} else {
...
}

File: manager/src/main/java/com/redhat/service/bridge/manager/connectors/AbstractConnectorWorker.java (hunk 9 above)
Comment (robpblake) [reply manager/src/main/java/com/redhat/service/bridge/manager/connectors/AbstractConnectorWorker.java:80]: We have a `@Version` annotation on the `ConnectorEntity` to enforce optimistic locking. With that only one transaction will be able to successfully commit the claim on the entity. The other will fail because the other transaction committed first and updated the version of the Entity. 

Essentially the `@Version` annotation makes the logic of the db write something like:

```
int updatedRows = persistEntity(entity);
if(updatedRows == 0){
throw OptimisticLockingException()
}
```

The SQL generated for the update looks something like `update table connector set id=1 where version=1`

If the `version` has changed then the number of rows updated is 0 and the commit fails.

File: manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsService.java [hunk 17]
```diff
@@ -9,11 +10,9 @@
 
 public interface ConnectorsService {
 
-    Optional<ConnectorEntity> createConnectorIfNeeded(BaseAction resolvedAction,
+    Optional<ConnectorEntity> persistConnectorIfNeeded(BaseAction resolvedAction,
```
robpblake, line 13: @lucamolteni Can we call this `createConnector` to align us with our other Service interfaces. The fact that it is returning an `Optional` to me means we don't need the `IfNeeded` suffix.

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsService.java
```diff
@@ -9,11 +10,9 @@
...
             Processor processor,
             ActionProvider actionProvider);
 
-    void deleteConnectorIfNeeded(BaseAction resolvedAction,
-            Processor processor,
-            ActionProvider actionProvider);
+    List<ConnectorEntity> deleteConnectorIfNeeded(Processor processor);
```
robpblake, line 17: @lucamolteni How comes a create returns a single `ConnectorEntity`, but a delete returns a list of `ConnectorEntity` ? Is there a scenario where the there is a one-to-many relationship between `Processor` and `Connector`?

manager/src/main/java/com/redhat/service/bridge/manager/models/Processor.java
```diff
@@ -29,7 +32,14 @@
...
-                query = "from Processor p join fetch p.bridge where p.status in (:statuses) and p.bridge.status='AVAILABLE' and p.shardId=:shardId"),
+                query = "select p " +
```
robpblake, line 35: @lucamolteni I just want to check my memory of our discussions: We said that we would make this more generic in the future so that we can more easily determine if all resources for either a BridgeInstance or a Processor are ready. This is the first iteration of this.

manager/src/main/java/com/redhat/service/bridge/manager/connectors/AbstractConnectorWorker.java
```diff
@@ -0,0 +1,126 @@
...
+    // Used for workers that should delete the entity after successfully calling the service.
+    // By default, after the service call updateEntityForSuccess is called, with this physical deletion will occur
+    protected boolean shouldDeleteAfterSuccess() {
+        return false;
+    }
+
+    // Could be the k8s pod name
+    String workerId = UUID.randomUUID().toString();
+
+    @Transactional(Transactional.TxType.NEVER)
```
robpblake, line 50: @lucamolteni I'd suggest adding a comment here as to why `TxType.NEVER` is important on this method for other people coming to the code who are perhaps less experienced with JPA/Hibernate than you.

robpblake (commented):
Thanks @lucamolteni . Some feedback for you.

This is definitely moving us in the right direction, so thank you for getting it to this point 👍 

cheers,

Rob

manager/src/main/java/com/redhat/service/bridge/manager/dao/ConnectorsDAO.java
lucamolteni: It's a `OneToOne` mapped with a `OneToMany/ManyToOne` so theoretically yes, even though in practice it's impossible. The reason for which I haven't used `OneToOne` is that in Hibernate they're usually complicated and it should be avoided (please excuse me why I don't write the exact reasons , I need to check again but it's something I've memorized) :)

I'd rather not assume in the code that we might not have by error two connectors per processor, so I think it's better to map it with a list anyway.

manager/src/main/java/com/redhat/service/bridge/manager/connectors/AbstractConnectorWorker.java
lucamolteni: `CONNECTORENTITY.findUnprocessed` checks for the unprocessed entities (those who don't have the workerId) and the claim is a lock that prevents two worker from working on the same entity

lucamolteni: Closing this as it should be cleared

manager/src/main/java/com/redhat/service/bridge/manager/models/Processor.java
lucamolteni: As this method is getting more and more specific, I think we should consider it in the future. I'm closing it for now

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsOrchestrateImpl.java
lucamolteni: Yes this will be adressed in https://issues.redhat.com/browse/MGDOBR-155

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
lucamolteni: Good catch, it was the ConnectorService opening a transaction because we changed from `MANDATORY` to `REQUIRES_NEW` in the previous PR. I'm setting Connector's back to `MANDATORY` with a comment

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
lucamolteni: We don't want to fire the event on the service bus while being in a transaction, so the method that writes on the DB is separated with its own `@Transactional` annotation, when it returns the event is fired

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
lucamolteni: Doing do will need to call yet another method on that service. A separation of service mapping 1:1 the name with the entities won't scale when we modify one than more entity.

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
lucamolteni: I think we should revisit the separation of Services - I don't think that having a separate service per entity is useful at all, especially in features dealing with multiple entity types.

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsService.java
lucamolteni: It's a `OneToOne` mapped as a `OneToMany/ManyToOne`, we cannot exclude that for a bug multiple connectors where created, so it's safer to delete them all anyway

manager/src/main/java/com/redhat/service/bridge/manager/connectors/ConnectorsServiceImpl.java
```diff
@@ -70,91 +44,38 @@
...
-        // https://issues.redhat.com/browse/MGDOBR-198
-        // this is currently ignored in the Connectors API
-        kafka.setId(KAFKA_ID_IGNORED);
+        List<ConnectorEntity> optionalConnector = connectorsDAO.findByProcessorId(processor.getId());
 
-        createConnectorRequest.setKafka(kafka);
+        return optionalConnector.stream().peek(c -> {
+            String connectorId = c.getId();
+            c.setDesiredStatus(ConnectorStatus.DELETED);
+            LOGGER.info("connector with id '{}' has been market for deletion", connectorId);
```
jomarko, line 74: small typo here

manager/src/main/java/com/redhat/service/bridge/manager/models/Processor.java
```diff
@@ -28,8 +31,15 @@
...
-                query = "from Processor p join fetch p.bridge where p.status in (:statuses) and p.bridge.status='AVAILABLE' and p.shardId=:shardId"),
+        @NamedQuery(name = "PROCESSOR.findByStatusesAndShardIdWithReadyDependencies",
+                query = "select p " +
+                        "from Processor p " +
+                        "join fetch p.bridge " +
+                        "left join p.connectorEntities as c " +
+                        "where p.status in (:statuses) " +
+                        "and p.bridge.status='AVAILABLE' " +
+                        "and p.shardId=:shardId " +
+                        "and (c is null or c.status = 'READY')"),
```
jomarko, line 42: Do not we have some enums for AVAILABLE and READY? just asking as saw some PR where some enums were renamed. So maybe using enums here could help if some status is renamed in future?

jomarko (commented):
thank you for a PR two newbie comments from me, probably not blocking from merging

manager/src/main/java/com/redhat/service/bridge/manager/connectors/AbstractConnectorWorker.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    // Used for workers that should delete the entity after successfully calling the service.
+    // By default, after the service call updateEntityForSuccess is called, with this physical deletion will occur
+    protected boolean shouldDeleteAfterSuccess() {
+        return false;
+    }
+
+    // Could be the k8s pod name
+    String workerId = UUID.randomUUID().toString();
```
danielezonca, line 48: Do you have a followup ticket to address/review this?

danielezonca (approved):
LGTM

manager/src/main/java/com/redhat/service/bridge/manager/models/Processor.java
lucamolteni: We have a generic `BridgeStatus` enum for the entities having the same lifecycle

infra/src/main/java/com/redhat/service/bridge/infra/models/dto/ConnectorStatus.java
```diff
@@ -3,10 +3,16 @@
 // TODO-MC review statuses
 // https://issues.redhat.com/browse/MGDOBR-195
 public enum ConnectorStatus {
-    REQUESTED,
-    PROVISIONING,
-    AVAILABLE,
-    DELETION_REQUESTED,
-    DELETED,
-    FAILED
+    ACCEPTED("accepted"),
```
r00ta, line 6: Unfortunately you have to migrate the existing rows in the database otherwise the demo cluster is broken. 
In short you have to do something like https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/fb7d556e761207bd649b1c20ba5d50a5f16c1226/manager/src/main/resources/db/migration/0.1/V0.1.6__migrate_statuses.sql

r00ta (approved):
LGTM thanks for the clarifications

manager/src/main/java/com/redhat/service/bridge/manager/connectors/AbstractConnectorWorker.java
```diff
@@ -0,0 +1,127 @@
...
+        return mergedCE;
+    }
+
+    @Transactional(Transactional.TxType.REQUIRES_NEW)
+    public ConnectorEntity successfullyCalledService(ConnectorEntity connectorEntity, R serviceResponse) {
+        ConnectorEntity updatedEntity = updateEntityForSuccess(connectorEntity, serviceResponse);
+        connectorEntity.setWorkerId(null);
+        connectorEntity.setModifiedAt(ZonedDateTime.now(ZoneOffset.UTC));
+
+        LOGGER.info("Worker {} successfully called service with response : {} on entity: {} ", workerName(), serviceResponse, updatedEntity);
```
sutaakar, line 97: Is it safe to log complete response and entity? It may have sensitive informations.

sutaakar (approved):
Looks good to me, though I wasn't able to test it properly due to MK issues.

manager/src/main/java/com/redhat/service/bridge/manager/connectors/AbstractConnectorWorker.java
lucamolteni: I don't think it's much more sensitive than the actual OpenBridge data such as Processor ID, so I'll leave it now as it's useful for debugging

PR #255, opened 2022-01-24
MGDOBR-241: inject webhook tech bearer token inside executor and setup local env dynamically

r00ta:
[MGDOBR-241](https://issues.redhat.com/browse/MGDOBR-241) 


r00ta:
Hi @sutaakar @robpblake @kostola @lucamolteni . Since in our local env we all have a different keycloak instance, we have to retrieve the offline token for the webhook tech bearer token dynamically at startup. 
For this reason, I've created two scripts `run_manager.sh` and `run_shard.sh` that will take care of retrieving the token and injecting it in the config. 

For the minikube overlay, the user has to retrieve it offline and inject it into the configmap. 
For the "prod" overlay, I've already retrieved it and encrypted. No need for additional actions. 

If any of you could double test this it would be great (maybe also on Mac, but I'm pretty confident it is already fine). Thanks

sutaakar (approved):
Seems to work as expected

dev/run_shard.sh
```diff
@@ -0,0 +1,55 @@
...
+    -Dquarkus.http.port=1337 \
+    -Dminikubeip=$minikube_ip \
+    -Devent-bridge.webhook.technical-bearer-token=$webhook_tech_token \
+    -Pminikube \
+    quarkus:dev
+}
+
+function retrieve_webhook_tech_token(){
+  local minikube_ip=$1
+  KEYCLOAK_RESPONSE=$(curl --insecure -X POST http://$minikube_ip:30007/auth/realms/event-bridge-fm/protocol/openid-connect/token --user event-bridge:secret -H 'content-type: application/x-www-form-urlencoded' -d 'username=webhook-robot-1&password=therobot&grant_type=password&scope=offline_access')
```
lucamolteni, line 22: Print the curl response here, it might be useful

lucamolteni (approved):

shard-operator/src/main/resources/templates/bridge-executor-deployment.yaml
```diff
@@ -36,7 +36,9 @@ spec:
               value: ""
             - name: KAFKA_GROUP_ID
               value: ""
-            - name: PROCESSOR_CONFIGURATION
+            - name: PROCESSOR_DEFINITION
```
robpblake, line 39: @r00ta Is the `PROCESSOR_CONFIGURATION` --> `PROCESSOR_DEFINITION` change intended?

r00ta: @robpblake it was a leftover. The template is just defining some empty env variables that are then replaced programmatically. The correct env var name is `PROCESSOR_DEFINITION`. Everything was running fine, it's just something I noticed and quickly fixed

dev/run_shard.sh
r00ta: Printed in case of failures, thanks

robpblake:
> Hi @sutaakar @robpblake @kostola @lucamolteni . Since in our local env we all have a different keycloak instance, we have to retrieve the offline token for the webhook tech bearer token dynamically at startup. For this reason, I've created two scripts `run_manager.sh` and `run_shard.sh` that will take care of retrieving the token and injecting it in the config.
> 
> For the minikube overlay, the user has to retrieve it offline and inject it into the configmap. For the "prod" overlay, I've already retrieved it and encrypted. No need for additional actions.
> 
> If any of you could double test this it would be great (maybe also on Mac, but I'm pretty confident it is already fine). Thanks

Thanks for the clear explanation @r00ta . This looks good.

I've created https://issues.redhat.com/browse/MGDOBR-252 for us to revisit our design/implementation of "Send To Bridge". It's adding a lot of complexity to our setup and as we discussed last week, we need to revisit the design at some point.

robpblake (approved):

PR #284, opened 2022-02-01
MGDOBR-281 - Demo readme improvements

jiripetrlik:
When I was going through the "DEMO.md" readme, I've found out two possible problems which can be improved:

1. There are wrong quotes around the authentication header in the send event request
2. It would be good to save the endpoint URL to environment variable and use it in later requests. It will follow the same principle as we do with the bridge and the processor ids.

### DIFF
--- a/DEMO.md
+++ b/DEMO.md
@@ -104,6 +104,10 @@ the response should look like
 
 Keep track of the `endpoint`, it will be used later when pushing an event to this ingress application.
 
+```bash
+export BRIDGE_ENDPOINT="$(curl -H "Authorization: $OB_TOKEN" -X GET $MANAGER_URL/api/v1/bridges/$BRIDGE_ID | jq --raw-output .endpoint)"
+```
+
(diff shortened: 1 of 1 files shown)


DEMO.md
```diff
@@ -104,6 +104,10 @@ the response should look like
 
 Keep track of the `endpoint`, it will be used later when pushing an event to this ingress application.
 
+```bash
+export BRIDGE_ENDPOINT=http://ob-87508471-ee0f-4f53-b574-da8a61285986-ob-kekkobar.apps.openbridge-dev.fdvn.p1.openshiftapps.com
```
sutaakar, line 108: @jiripetrlik Can you use jq to parse curl response and store the endpoint? That way users don't have to edit your command, just execute it.

jiripetrlik: @sutaakar Thank you for good tip. Done.

sutaakar (approved):

kostola (approved):

PR #305, opened 2022-02-08
MGDOBR-263 - Refactor integration tests to be self contained

jiripetrlik:
[MGDOBR-263](https://issues.redhat.com/browse/MGDOBR-263) 
Currently implemented tests ([MGDOBR-205](https://issues.redhat.com/browse/MGDOBR-205)) depend on each other to be executed in specific order.
This makes them fragile - if a test fails then all subsequent tests will most likely fail too. Also it is not possible to execute a specific test, user needs to execute whole test suite even when interested in one test result.

Purpose of this JIRA is to convert tests to be self contained - every test should be executed separately regardless of other tests or execution order. Every test will be responsible to setup the environment needed.

Acceptance Criteria
- Every test can be executed separately.
- !!!CHANGE: "Every test should clean its resources in case of a successful execution (cleanup during a test failure will be addressed as a separate JIRA)" - From what I've heard we plan to delete all bridges after every test in MGDOBR-264. Therefore I consider deleting of the bridges etc. as unnecessary. Cleanup in the test logic may in my opinion also hide the logic of the test. @Srihari1192 Please correct me if I'm wrong.

- [* ] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/main/kogito-build/kogito-ide-config)
- [* ] Your commit messages are clear and reference the JIRA issue e.g: "[MGDOBR-1] - $clear_explanation_of_what_you_did"
- [* ] All new functionality is tested
- [* ] Pull Request title is properly formatted: `MGDOBR-XYZ Subject`
- [* ] Pull Request contains link to the JIRA issue
- [* ] Pull Request contains link to any dependent or related Pull Request
- [* ] Pull Request contains description of the issue
- [* ] Pull Request does not include fixes for issues other than the main ticket
- [* ] Annotate your PR with the label `safe to test` in order to run the end to end CI pipeline

### DIFF
--- a/integration-tests/src/test/java/com/redhat/service/bridge/integration/tests/steps/End2EndTestITStep.java
+++ b/integration-tests/src/test/java/com/redhat/service/bridge/integration/tests/steps/End2EndTestITStep.java
@@ -4,6 +4,7 @@ import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
+import java.util.UUID;
 
 import org.awaitility.Awaitility;
 import org.hamcrest.Matchers;
@@ -40,6 +41,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 public class End2EndTestITStep {
 
+    private static String randomBridgeName;
     private static String bridgeId;
     private static String processorId;
     private static String endPoint;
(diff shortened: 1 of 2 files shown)


r00ta (approved):

integration-tests/src/test/java/com/redhat/service/bridge/integration/tests/steps/End2EndTestITStep.java
```diff
@@ -56,18 +58,18 @@ public void authenticationIsEnabled(int responseCode) {
                 .statusCode(responseCode);
     }
 
-    @Given("^get list of Bridge instances with access token doesn't contain Bridge \"([^\"]*)\"$")
-    public void getEmptyBridges(String bridgeName) {
+    @Given("bridge with randomly generated name doesn't exist")
```
sutaakar, line 61: What about leaving the old step text there, as it contains context info about containing access token?

sutaakar:
@jiripetrlik To reduce repetitiveness we can restructure scenarios - move testing of deleted resources to the creation scenarios:
- one scenario will test creation and deletion of Bridges and Ingress
- one scenario will provision bridge and test creation and deletion of Processors.

What do you think?

jiripetrlik:
@sutaakar 
Thank you for your review. I've updated the PR.

integration-tests/src/test/java/com/redhat/service/bridge/integration/tests/steps/End2EndTestITStep.java
```diff
@@ -56,18 +58,18 @@ public void authenticationIsEnabled(int responseCode) {
                 .statusCode(responseCode);
     }
 
-    @Given("^get list of Bridge instances with access token doesn't contain Bridge \"([^\"]*)\"$")
-    public void getEmptyBridges(String bridgeName) {
+    @Given("get list of Bridge instances with access token doesn't contain Bridge")
```
sutaakar, line 61: Maybe
```suggestion
    @Given("get list of Bridge instances with access token doesn't contain generated Bridge")
```
?
To make clear what Bridge we reference to.

jiripetrlik: @sutaakar Fixed. Thank you.

sutaakar (approved):

PR #369, opened 2022-02-23
[experiment] Add async processing for ManagedResources

r00ta:

<details>
<summary>
</summary>

  Comment with: `/rebase`.

</details>


r00ta:
@manstis I had a quick sync with @lucamolteni about it this morning. I know the approach here is slightly different from the one implemented for connectors, I just tried to explore some alternatives. It's still a WIP

manager/src/main/java/com/redhat/service/bridge/manager/BridgeWorker.java
```diff
@@ -101,17 +124,45 @@ public void step1_createKafkaTopic(Bridge bridge) {
...
+            transact(preparingWorker);
+            return;
+        }
+
+        PreparingWorker preparingWorker = preparingWorkerDAO.findByEntityId(bridge.getId()).get(0);
+        preparingWorkerDAO.deleteById(preparingWorker.getId());
+
+        // The bridge is ready to be provisioned by the shard.
+        bridge.setStatus(ManagedEntityStatus.DEPROVISION);
+        bridge.setDesiredStatus(ManagedEntityStatus.DELETING);
```
manstis, line 146: Isn't the desired status `DELETED`? 

Or does each step set the desired state to be what it should be for when the step completes successfully.

manager/src/main/java/com/redhat/service/bridge/manager/providers/WorkerIdProviderImpl.java
```diff
@@ -0,0 +1,16 @@
...
+import javax.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped
+public class WorkerIdProviderImpl implements WorkerIdProvider {
+
+    private static final String WORKER_ID = UUID.randomUUID().toString();
+
+    @Override
+    public String getWorkerId() {
+        return WORKER_ID;
```
manstis, line 14: Since this class is `@ApplicationScoped` isn't this likely to return the same ID for all `ManagedResource`'?

manager/src/main/java/com/redhat/service/bridge/manager/BridgeWorker.java
```diff
@@ -0,0 +1,206 @@
...
+    @Inject
+    WorkerIdProvider workerIdProvider;
+
+    @Override
+    public void accept(Bridge b) {
+        LOGGER.info("Accept entity " + b.getId());
+        if (b.getStatus() != null) {
+            LOGGER.warn("The status of the entity is not null. It will be overwritten.");
+        }
+        b.setStatus(ManagedEntityStatus.ACCEPTED);
```
manstis, line 64: I think you may need to review the status's used.

See my comment below about the Shard processing `ACCEPTED` Bridges.

manager/src/main/java/com/redhat/service/bridge/manager/BridgeWorker.java
```diff
@@ -0,0 +1,206 @@
...
+            preparingWorker.setStatus(BridgeWorkerStatus.FAILURE.toString()); // where you are. The rescheduler will fire the event according to the desiredStatus
+            transact(preparingWorker);
+            return;
+        }
+
+        PreparingWorker preparingWorker = preparingWorkerDAO.findByEntityId(bridge.getId()).get(0);
+        preparingWorkerDAO.deleteById(preparingWorker.getId());
+
+        // The bridge is ready to be provisioned by the shard.
+        bridge.setStatus(ManagedEntityStatus.PREPARING);
```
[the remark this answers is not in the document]
r00ta: @manstis I still have to update the shard api: it should return to the shard all the instances with `desiredStatus` equal to `PROVISIONING`

manager/src/main/java/com/redhat/service/bridge/manager/BridgeWorker.java
r00ta: See previous reply

manager/src/main/java/com/redhat/service/bridge/manager/providers/WorkerIdProviderImpl.java
r00ta: Yes, this is what the workerId is supposed to do

r00ta: The workerId identifies a application instance. Since the application can have multiple replica, every instance has its own workerId

manstis: Ok. Understood. I thought it was meant to be per `ManagedResource`.

manager/src/main/java/com/redhat/service/bridge/manager/BridgeWorker.java
r00ta: I still have to think about this. I'll get back to you

manstis: Investigating an issue I have here with my MC changes; I think having each step setting the desired status to what it should be at the end of the step is better (than what I wrote above suggesting it should be the terminal/ultimate status). The problem I have here with using the terminal/ultimate status is that the re-try mechanism is not able to ascertain what the next step should be - when you have multiple steps in a chain as we do for MC provisioning.

manstis: tl; dr; I think you are correct.

manager/src/main/java/com/redhat/service/bridge/manager/BridgeWorker.java
```diff
@@ -0,0 +1,208 @@
...
+        PreparingWorker preparingWorker = preparingWorkerDAO.findByEntityId(bridge.getId()).get(0); // there should be no parallel workers for this resource.
+        preparingWorkerDAO.deleteById(preparingWorker.getId());
+
+        // The bridge is ready to be provisioned by the shard.
+        bridge.setStatus(ManagedEntityStatus.PREPARING);
+        bridge.setDesiredStatus(ManagedEntityStatus.PROVISIONING);
+        transact(bridge);
+    }
+
+    @Transactional
```
robpblake, line 127: @r00ta We need to be careful here. The transaction boundary on this method involves the call to the remote service. Part of the reason to introduce the async model is to avoid calling a remote service within the boundaries of a DB transaction.

manager/src/main/java/com/redhat/service/bridge/manager/BridgeWorker.java
```diff
@@ -0,0 +1,208 @@
...
+        transact(preparingWorker);
+
+        routePreparingWorker(preparingWorker, bridge);
+    }
+
+    @Transactional
+    @ConsumeEvent(value = CREATE_KAFKA_TOPIC, blocking = true)
+    public void step1_createKafkaTopic(Bridge bridge) {
+        try {
+            rhoasService.createTopicAndGrantAccessFor(internalKafkaConfigurationProvider.getBridgeTopicName(bridge), RhoasTopicAccessType.CONSUMER_AND_PRODUCER);
```
robpblake, line 111: We need to ensure that `RhoasService` is idempotent, otherwise we will always fail if we have to retry. Imagine the following scenario:

1. First time we create the Topic, but fail to create the ACL (what we're seeing at the minute a lot)
2. Second time, we fail because the Topic already exists
3. Repeat

`RhoasService` needs to be able to create the ACL if the requested topic already exists.

manager/src/main/java/com/redhat/service/bridge/manager/BridgeWorker.java
```diff
@@ -0,0 +1,208 @@
...
+
+    private PreparingWorker updatePreparingWorker(String entityId, BridgeWorkerStatus status) {
+        PreparingWorker preparingWorker = preparingWorkerDAO.findByEntityId(entityId).get(0);
+        preparingWorker.setModifiedAt(ZonedDateTime.now(ZoneOffset.UTC));
+        preparingWorker.setStatus(status.toString()); // where you are. The rescheduler will fire the event according to the desiredStatus
+        transact(preparingWorker);
+        return preparingWorker;
+    }
+
+    private void routePreparingWorker(PreparingWorker preparingWorker, Bridge bridge) {
```
robpblake, line 196: My feeling is that what the worker needs to do should be encapsulated on the Bridge entity somehow. It may be that the Worker has the ability to add metadata/status to the `Bridge` somehow to track what it is doing.

That way the worker can simply be passed a `Bridge` Entity and then work out what it needs to do from the status of the `Bridge` and making calls to the other remote services to determine if all resources have been provisioned. _How_ the worker is initiated then becomes a separate concern

manager/src/main/java/com/redhat/service/bridge/manager/BridgeWorker.java
```diff
@@ -0,0 +1,208 @@
...
+        bridge.setStatus(ManagedEntityStatus.PREPARING);
+        bridge.setDesiredStatus(ManagedEntityStatus.PROVISIONING);
+        transact(bridge);
+    }
+
+    @Transactional
+    @ConsumeEvent(value = DELETE_KAFKA_TOPIC, blocking = true)
+    public void step1_deleteKafkaTopic(Bridge bridge) {
+        try {
+            rhoasService.deleteTopicAndRevokeAccessFor(internalKafkaConfigurationProvider.getBridgeTopicName(bridge), RhoasTopicAccessType.CONSUMER_AND_PRODUCER);
```
robpblake, line 131: I think we should let the Shard delete first before we delete resources on which it depends. It gives the Shard the chance to safely clean-up the environment before all resources are destroyed.

manager/src/main/java/com/redhat/service/bridge/manager/BridgeWorker.java
```diff
@@ -0,0 +1,208 @@
...
+            routePreparingWorker(preparingWorker, bridge);
+        }
+    }
+
+    /**
+     * Every 5 minutes it will retrieve all the instances that have not been updated for more than 30 minutes.
+     */
+    @Override
+    @Scheduled(every = "5m", concurrentExecution = Scheduled.ConcurrentExecution.SKIP)
+    public void discoverOrphanWorkers() {
```
robpblake, line 163: If I step back, this feels like we're trying to achieve what Kafka could give us: the ability to have multiple workers process a single job queue. 

We have a hard dependency on Kafka for OpenBridge. If we were to consider a work queue for each customer account and then the workers simply become kafka consumers, does that make our life simpler?

At the minute we have an artificial 5 min wait period should a worker get rescheduled (we can of course reduce this through configuration). If instead we didn't commit that the offset for the event on Kafka has been processed, we have the ability to simply bring up another consumer straight away and continue processing.

manstis: I tend to agree; we are inventing our own durable and resilient state machine.

This is in essence the same pattern that we are achieving with our use of the DAO and VertX Event Bus.

It's definitely more things to consider on Monday.

manager/src/main/java/com/redhat/service/bridge/manager/BridgeWorker.java
r00ta: it is idempotent.

r00ta:
handed over to @manstis

PR #370, opened 2022-02-23
[28bfc] Update kustomization images

r00ta:


manstis:
I am intrigued.

You've made a few of these PRs recently and I'd like to understand why; should I make a change somewhere that means I'd need to make similar changes in the future.

r00ta:
> I am intrigued.
> 
> You've made a few of these PRs recently and I'd like to understand why; should I make a change somewhere that means I'd need to make similar changes in the future.

@manstis Every time a pull request is merged some new docker images are deployed on quay. Those PRs aims to update the kustomize project that will deploy automatically the new docker images on the demo cluster

PR #375, opened 2022-02-24
MGDOBR-352: Unify Actions type naming - Remove Action suffix

jomarko:
[MGDOBR-352](https://issues.redhat.com/browse/MGDOBR-352) 

<details>
<summary>
</summary>

  Comment with: `/rebase`.

</details>

### DIFF
--- a/ACTIONS.md
+++ b/ACTIONS.md
@@ -20,7 +20,7 @@ Each `Action` has 3 parameters to specify:
 
 The following `Actions` are currently supported by Event Bridge:
 
-### SendToKafka
+### KafkaTopic
 
 Allows you to send an `Event` to a Kafka Topic on a hard-coded Kafka Cluster made available via the Event Bridge deployment
 
@@ -35,7 +35,7 @@ To send an Event to the topic `myRequestedTopic`:
 ```json
 {
   "action": {
-    "type": "SendToKafka",
+    "type": "KafkaTopic",
     "name": "Send to Kafka Topic",
     "parameters": {
       "topic": "myRequestedTopic"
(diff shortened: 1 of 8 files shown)


r00ta (approved):
Please notify Heiko when you merge this since it is a breaking change

danielezonca (commented):
Overall the PR is good but do we want to create a script to migrate existing processor instances or not for now?
@robpblake @r00ta 
Wdyt?

r00ta:
> Overall the PR is good but do we want to create a script to migrate existing processor instances or not for now? @robpblake @r00ta Wdyt?

By heart it might still work but in the responses we would have the `Action` suffix for the processors that were created before this PR. I'm not sure we should put effort to migrate the existing processors/actions since we would have to write some sql/json magic to do the job. We can erase the db and start from scratch if it's ok for Heiko

robpblake:
I think @r00ta is correct. If the Action is already deployed, then I think it will continue to work. The `type` is just a string that we only use at the point of resolving the Action when the user first submits their request.

It's only new Actions will need to use the new type name however.

manstis:
What if you update a processor? (Not in scope atm, but there's a JIRA for it).

If you don't also update the Action type the re-deployment will fail.

However presumably our test/staging environment would/could be torn down before then?

robpblake:
> What if you update a processor? (Not in scope atm, but there's a JIRA for it).
> 
> If you don't also update the Action type the re-deployment will fail.
> 
> However presumably our test/staging environment would/could be torn down before then?

danielezonca (approved):

danielezonca:
@jomarko 
Can you please fix conflicts?

PR #422, opened 2022-03-09
[MGDOBR-29] - Support Users Fetching Bridge and Processors by Name from the API

vaibhavjainwiz:

MGDOBR-29 - https://issues.redhat.com/browse/MGDOBR-29

Below endpoint are modified to receive bridge id or name as input parameter :
```
GET /api/v1/bridges/{bridgeIdentifier}
DELETE /api/v1/bridges/{bridgeIdentifier}
GET /api/v1/bridges/{bridgeIdentifier}/processors/{processorId}
GET /api/v1/bridges/{bridgeIdentifier}/processors
POST /api/v1/bridges/{bridgeIdentifier}/processors
DELETE {bridgeIdentifier}/processors/{processorId} 
```
`bridgeIdentifier` could be bridge id or bridge name.

<details>
<summary>
</summary>

  Comment with: `/rebase`.

</details>


sutaakar:
@vaibhavjainwiz Can you please check the test failures?

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
```diff
@@ -164,31 +165,33 @@ public Processor updateProcessorStatus(ProcessorDTO processorDTO) {
 
     @Transactional
     @Override
-    public Long getProcessorsCount(String bridgeId, String customerId) {
-        return processorDAO.countByBridgeIdAndCustomerId(bridgeId, customerId);
+    public Long getProcessorsCount(Bridge bridge) {
```
lucamolteni, line 168: I can see why of this change, but I think that having the methods not to depend on the full class but rather on the primitive key instance was intended. It lighten the API and makes it easier to test it for example.

manager/src/main/java/com/redhat/service/bridge/manager/BridgesService.java
```diff
@@ -16,9 +17,31 @@
...
-
-    Bridge getReadyBridge(String bridgeId, String customerId);
+    /**
+     * Provides Bridge instance for given id/name. It throws `ItemNotFoundException` if bridge instance not found.
+     * 
+     * @param bridgeIdentifier It could be a bridge Id or bridge name.
+     * @param customerId Customer id.
+     * @return Bridge instance.
+     */
+    Bridge getBridgeByBridgeIdentifier(String bridgeIdentifier, String customerId);
```
lucamolteni, line 27: `getBridgeByBridgeIdentifier` sounds like a tongue-twister. What's wrong with `getBridgeById`?

manager/src/main/java/com/redhat/service/bridge/manager/BridgesService.java
```diff
@@ -16,9 +17,31 @@
...
+
+    /**
+     * Provide Bridge instance for given bridge id.
+     * 
+     * @param id Bridge id.
+     * @param customerId Customer id.
+     * @return Optional Bridge instance.
+     */
+    Optional<Bridge> getBridgeByIdAndCustomerId(String id, String customerId);
```
lucamolteni, line 36: Let's user either `Id` or `Identifier` as they mean the same thing. I'd go for `Id` as it's shorter and well-known.

lucamolteni (changes requested):
Thank you for the work. 

I added a few stylish comment, can you please take a look at it?

Thanks

manager/src/main/java/com/redhat/service/bridge/manager/BridgesService.java
vaibhavjainwiz: `BridgeIdentifier` could be an `Bridge Id` or a `Bridge Name` that why I change `getBridgeById` to `getBridgeByBridgeIdentifier` but if we suggest I could change this method name to `getBridgeByIdentifier`.

manager/src/main/java/com/redhat/service/bridge/manager/BridgesService.java
vaibhavjainwiz: `Id` & `Identifier` are not same. `Id` corresponds to only `Bridge ID` but Identifier could be an `Bridge Id` or a `Bridge Name`.

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
vaibhavjainwiz: @lucamolteni Sorry I didn't get How it could makes easier to test it for example . Processors are not depends on customer id at all so passing customer id to processor methods are more confusing. 

There are multiple improvement that we could make here as following :
- Ideally we should not pass `Bridge` entity to other services at all. We should fetch `Bridge` entity inside `BridgeDAO` and then convert it to some `BridgeDTO` to pass bridge information internally.

- Inside `processorDAO.countByBridgeIdAndCustomerId` we are making a Query like `p.bridge.id=:bridgeId and p.bridge.customerId=:customerId`. With my changes as we already fetch Bridge instance using `bridgeId` & `customerId`, now we could make this call simple by just filter processors where `p.bridge.id=:bridgeId`.

manager/src/main/java/com/redhat/service/bridge/manager/BridgesService.java
kostola: Using `Id` and `Identifier` with two different logical meanings is very confusing, since in plain english the former is an abbreviation of the latter.

We should find another name (like `key` for example).

manager/src/main/java/com/redhat/service/bridge/manager/BridgesService.java
kostola: Same consideration as [here](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/422#discussion_r823698003)

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
```diff
@@ -85,40 +86,46 @@ public Bridge getBridge(String id) {
...
+    @Override
+    public Bridge getBridgeByBridgeIdentifier(String bridgeIdentifier, String customerId) {
+        Optional<Bridge> bridgeOptional = getBridgeByIdAndCustomerId(bridgeIdentifier, customerId);
+        if (bridgeOptional.isPresent()) {
+            return bridgeOptional.get();
+        }
+
+        bridgeOptional = getBridgeByNameAndCustomerId(bridgeIdentifier, customerId);
+        if (bridgeOptional.isPresent()) {
+            return bridgeOptional.get();
```
kostola, line 102: I think we should create a unified query with a condition like `WHERE customer_id = <customerId> AND (id = <bridgeIdentifier> OR name = <bridgeIdentifier>)`. What do you think?

manager/src/main/java/com/redhat/service/bridge/manager/api/user/ProcessorsAPI.java
```diff
@@ -47,39 +50,49 @@
...
     IdentityResolver identityResolver;
 
     @Inject
     JsonWebToken jwt;
 
     @GET
-    @Path("{bridgeId}/processors/{processorId}")
-    public Response getProcessor(@NotEmpty @PathParam("bridgeId") String bridgeId, @NotEmpty @PathParam("processorId") String processorId) {
+    @Path("{bridgeIdentifier}/processors/{processorId}")
+    public Response getProcessor(@NotEmpty @PathParam("bridgeIdentifier") String bridgeIdentifier, @NotEmpty @PathParam("processorId") String processorId) {
```
r00ta, line 64: `bridgeIdentifier` is a bit misleading, `idOrName` would be better I guess

manager/src/main/java/com/redhat/service/bridge/manager/BridgesService.java
lucamolteni: @vaibhavjainwiz "Identifier" means something that identifies unequivocally a row in the database. 

As the name is not in the primary key, name is not an identifier. I'd change it, it's confusing

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
lucamolteni: it's easier to test because you can call the API without having to have a fully constructed entity

manager/src/main/java/com/redhat/service/bridge/manager/api/user/BridgesAPI.java
```diff
@@ -66,16 +66,16 @@ public Response createBridge(BridgeRequest bridgeRequest) {
     }
 
     @GET
-    @Path("{bridgeId}")
-    public Response getBridge(@PathParam("bridgeId") @NotEmpty String bridgeId) {
-        Bridge bridge = bridgesService.getBridge(bridgeId, identityResolver.resolve(jwt));
+    @Path("{bridgeIdentifier}")
```
robpblake, line 69: @vaibhavjainwiz I think it's clearer for a user on how to use the API if this parameter is `idOrName`. The parameter is self explanatory.

manager/src/main/java/com/redhat/service/bridge/manager/api/user/ProcessorsAPI.java
robpblake: +1 to `idOrName` as it is self-describing.

manager/src/main/java/com/redhat/service/bridge/manager/dao/BridgeDAO.java
```diff
@@ -24,16 +25,16 @@
         return find("#BRIDGE.findByStatusesAndShardId", params).list();
     }
 
-    public Bridge findByNameAndCustomerId(String name, String customerId) {
+    public Optional<Bridge> findByNameAndCustomerId(String name, String customerId) {
```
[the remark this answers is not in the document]
robpblake: Agreed. @vaibhavjainwiz Lets go for something like `from Bridge b where (b.name=:idOrName or id=:idOrName) and customerId=:customerId` rather than having two queries. That just gives us a single entry point for locating a bridge.

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
```diff
@@ -79,24 +80,25 @@
...
 
         return processor;
     }
 
     @Override
-    public Processor createProcessor(String bridgeId, String customerId, ProcessorRequest processorRequest) {
+    public Processor createProcessor(Bridge bridge, ProcessorRequest processorRequest) {
         /* We cannot deploy Processors to a Bridge that is not Available */
-        Bridge bridge = bridgesService.getReadyBridge(bridgeId, customerId);
+        if (!bridgesService.isBridgeReady(bridge)) {
```
robpblake, line 95: This doesn't feel right. Determining whether or not a Bridge is ready should be a concern of the `BridgesService`. The previous implementation was correct IMHO with the `BridgesService.getReadyBridge` returning either the ready bridge instance or throwing the `BridgeLifecycleException`.

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
robpblake: I think I'm +1 on this change. It feels more aligned with our other service interfaces where we are working with sub-resources. For example `ConnectorsService` takes in a `Processor` parameter to store the requested `Connector`.

I think this change, combined with the following, gives us a cleaner overall approach

- `ProcessorsAPI` is marked with `@Transactional` at the class-level to ensure we have a single DB transaction for the read of both the `Bridge` and the associated `Processors`
- The `BridgeServices.isReadyBridge(Bridge bridge)` call throws the `BridgeLifecycleException` if the bridge is not `READY`

I don't agree with the idea of using a DTO internally. I see no reason why we can't use our `Bridge` entity for Services within the Fleet Manager.

I would also prefer to keep the `and p.bridge.customerId=:customerId` on our queries for Processors out of paranoia.

@lucamolteni @vaibhavjainwiz WDYT?

robpblake (changes requested):
Thank you for your first PR @vaibhavjainwiz 🚀 

There's some basic stylistic feedback and then a couple of other considerations.

cheers,

Rob

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
```diff
@@ -85,40 +86,46 @@ public Bridge getBridge(String id) {
     }
 
     @Transactional
```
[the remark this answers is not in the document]
vaibhavjainwiz: We are calling `bridgeDAO.persist(bridge);` in this function. which would fail without transaction.

manager/src/main/java/com/redhat/service/bridge/manager/BridgesService.java
vaibhavjainwiz: As suggested by @robpblake, I changed it to `IdOrName`. Now function is `getBridgeByIdOrName`

manager/src/main/java/com/redhat/service/bridge/manager/BridgesService.java
vaibhavjainwiz: I changed it to `getBridgeByIdOrNameAndCustomerId`.

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
vaibhavjainwiz: done

manager/src/main/java/com/redhat/service/bridge/manager/dao/BridgeDAO.java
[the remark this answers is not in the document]
vaibhavjainwiz: done

manager/src/test/java/com/redhat/service/bridge/manager/actions/sendtobridge/SendToBridgeActionTransformerTest.java
```diff
@@ -70,12 +66,13 @@ static void beforeAll() {
...
-        when(bridgesServiceMock.getReadyBridge(not(or(eq(UNAVAILABLE_BRIDGE_ID), or(eq(BRIDGE_ID), eq(OTHER_BRIDGE_ID)))), eq(TEST_CUSTOMER_ID)))
-                .thenThrow(new ItemNotFoundException("Bridge not found"));
-        when(bridgesServiceMock.getReadyBridge(any(), not(eq(TEST_CUSTOMER_ID)))).thenThrow(new ItemNotFoundException("Customer not found"));
+        when(bridgesServiceMock.isBridgeReady(bridge)).thenReturn(true);
+        when(bridgesServiceMock.isBridgeReady(otherBridge)).thenReturn(true);
+        //
+        //        when(bridgesServiceMock.isBridgeReady(UNAVAILABLE_BRIDGE_ID, TEST_CUSTOMER_ID)).thenThrow(new BridgeLifecycleException("Unavailable bridge"));
+        //        when(bridgesServiceMock.isBridgeReady(not(or(eq(UNAVAILABLE_BRIDGE_ID), or(eq(BRIDGE_ID), eq(OTHER_BRIDGE_ID)))), eq(TEST_CUSTOMER_ID)))
+        //                .thenThrow(new ItemNotFoundException("Bridge not found"));
+        //        when(bridgesServiceMock.isBridgeReady(any(), not(eq(TEST_CUSTOMER_ID)))).thenThrow(new ItemNotFoundException("Customer not found"));
```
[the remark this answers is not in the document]
vaibhavjainwiz: reverted all my changes in this file.

manager/src/main/java/com/redhat/service/bridge/manager/api/user/ProcessorsAPI.java
vaibhavjainwiz: done

manager/src/main/java/com/redhat/service/bridge/manager/api/user/BridgesAPI.java
vaibhavjainwiz: done

manager/src/main/java/com/redhat/service/bridge/manager/dao/BridgeDAO.java
```diff
@@ -24,16 +25,16 @@
...
         Parameters params = Parameters
                 .with("name", name).and("customerId", customerId);
-        return find("#BRIDGE.findByNameAndCustomerId", params).firstResult();
+        return find("#BRIDGE.findByNameAndCustomerId", params).firstResultOptional();
     }
 
-    public Bridge findByIdAndCustomerId(String id, String customerId) {
+    public Optional<Bridge> findByIdAndCustomerId(String id, String customerId) {
```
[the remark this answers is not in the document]
vaibhavjainwiz: done

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
vaibhavjainwiz: done

manager/src/main/java/com/redhat/service/bridge/manager/ProcessorServiceImpl.java
vaibhavjainwiz: @robpblake As you suggested, I had made changes into my PR. But I am not sure with marking `@Transactional` over `ProcessorsAPI`.  Is there be any race condition scenerio if both entities loaded in different transaction?

vaibhavjainwiz:
> @vaibhavjainwiz Can you please check the test failures?

done

r00ta (commented):
I think the code looks good. 
I'm still not 100% convinced about the feature itself, since also the other managed service do not offer it (see https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/tree/main/openapi)

manager/src/test/java/com/redhat/service/bridge/manager/ProcessorServiceTest.java
```diff
@@ -126,40 +125,27 @@ private BaseAction createKafkaAction() {
         return a;
     }
 
-    @Test
```
sutaakar, line 129: Why did you delete these tests?

vaibhavjainwiz: `bridgeNotActive` and `bridgeDoesNotExist` checks are moved to controllers. `processorService.createProcessor` functions would call only when bridge is active/exists.

sutaakar: ok

kostola: > `bridgeNotActive` and `bridgeDoesNotExist` checks are moved to controllers.

I think we should consider each service as independent and not always assume a certain behavior by the caller.
I think these tests must be restored.
Also...I don't see where these checks have been moved in this PR.

vaibhavjainwiz: @kostola 
This change had already been discussed with Luca and Rob in above review comment.
https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/422#discussion_r823957472

These changes are now moved to API's.
https://github.com/vaibhavjainwiz/sandbox/blob/91ef48b69884cc3532ca9f9e06cf0ac79f3f43ac/manager/src/main/java/com/redhat/service/bridge/manager/api/user/ProcessorsAPI.java#L125

r00ta (commented):
Unfortunately with this new feature the `id` and the `names` must be unique. 
Assume that the user `X` creates the Bridge with id `xyz`. Then the user should not be able to create a new Bridge with name `xyz`. We should add this check.

robpblake (changes requested):
@vaibhavjainwiz I'm going to change my mind on this one. I had originally thought this was a good idea and now coming back to it, I'm not sure it was.

I think the more RESTful approach here is for us to support fetching by name as a query parameter instead. So that would leave us with the following endpoints:

- `/api/v1/bridges?name=foo` <-- find a bridge with name `foo`
- `/api/v1/bridges/{id}` <-- find a bridge by id
- `/api/v1/bridges/{id}/processors?name=foo` <-- Find a processor with name `foo` on bridge with `{id}`
- `/api/v1/bridges/{id}/processors/{id}` <-- Find a processor with specific id

I'm not sure my suggestion to overload the `idOrName` parameter was a great one tbh.

I'd suggest we close this PR and perhaps re-work it to use the query parameter approach instead. That would also give us a pattern to use for https://issues.redhat.com/browse/MGDOBR-355 as well.

Apologies for the late change of mind.

cc @r00ta @lucamolteni @danielezonca @kostola @manstis 

cheers,

Rob

PR #449, opened 2022-03-15
MGDOBR-395: add instructions to configure Managed Connector operator update

kostola:
[MGDOBR-395](https://issues.redhat.com/browse/MGDOBR-395)


robpblake (changes requested):
@kostola The `Cronjob` definition should be included as part of our kustomize overlay for every environment where this is required so that it is applied by either ArgoCD or app-interface pipelines.

cheers,

Rob

dev/ManagedConnectors.md
```diff
@@ -144,6 +144,89 @@ ocm login --token <offline_token>
 ./cos-tools/bin/create-cluster-secret <cluster_id>
 ```
 
+### Step 6: Update Job
+
+There's no automatic CI/CD mechanism in Managed Connectors to update the operators with the newest version.
+
+The only way at the moment is to configure a Kubernetes [CronJob](https://v1-21.docs.kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/)
+that kills the existing operator pods, so that the deployment triggers the pull of the newest image (if existing) before recreating it.
```
lucamolteni, line 152: "reschedule" instead of "kill"

lucamolteni (approved):

dev/ManagedConnectors.md
```diff
@@ -144,6 +144,89 @@ ocm login --token <offline_token>
...
+    spec:
+      backoffLimit: 2
+      activeDeadlineSeconds: 60
+      template:
+        spec:
+          serviceAccountName: cos-updater
+          restartPolicy: Never
+          containers:
+            - name: kubectl
+              image: bitnami/kubectl:1.21.6
```
robpblake, line 213: @kostola Is there an official `oc` cli image we can use instead?

robpblake:
@kostola I can see two problems here that we can't forget:

1. What happens if the `CronJob` does not execute for an extended period of time. Is that something that we would need to be alerted on and fix? For example, imagine for whatever reason, we're failing to pull `bitnami/kubectl:1.21.6` container image and we don't restart the pods for days. What's the problem that causes us (if any) and how do we know it needs to be fixed

2. How are we sure that the `cos` services are back up and running after we have killed the pods? As an example: we `kubectl delete` the `cos` pods, K8S tries to reschedule the pods onto another node that doesn't have the image, quay.io is down and therefore we can't pull the image. At that point, we're in outage mode for our Managed Connectors.

r00ta:
> 

Regarding the point 2 I think it should be responsibility of the MC operator to report that the connector is not available to their fleet manager. 
We would then check on MC fleet manager what are the connectors in a degraded status. 

In general we do not have any periodic check on MC fleet manager to find out if any connector changed the status, I think this is something we will need to improve the reliability/resiliency of the service.

kostola:
Honestly I think we're overthinking this too much. This is just a "hack" to temporary overcome the problem that there's no official and well tested mechanism to deliver updates to the managed connector shard operators.
In my humble opinion, we should keep this as it is (it's already working with no major issues in both clusters) and let the managed connector team provide us the instruction for the production-ready configuration once they define one.

robpblake:
> Honestly I think we're overthinking this too much. This is just a "hack" to temporary overcome the problem that there's no official and well tested mechanism to deliver updates to the managed connector shard operators. In my humble opinion, we should keep this as it is (it's already working with no major issues in both clusters) and let the managed connector team provide us the instruction for the production-ready configuration once they define one.

@kostola My point is that we need to understand if the production-ready configuration is a priority for Managed Connectors before we get to Service Preview.

If it's a priority for them and they will deliver it before our own service preview, that's great and we need to schedule some work to remove this hack.

If it's not a priority for them and we have to go to our own Service Preview with this hack in place, then there is work for us to do to make sure this hack doesn't take our own Service offline. 

AFAICS we're not recording the need to address these points ^ beyond this PR. I don't want us to just merge this and then forget about it for a few months.

Does that make sense?

robpblake:
> Honestly I think we're overthinking this too much. This is just a "hack" to temporary overcome the problem that there's no official and well tested mechanism to deliver updates to the managed connector shard operators. In my humble opinion, we should keep this as it is (it's already working with no major issues in both clusters) and let the managed connector team provide us the instruction for the production-ready configuration once they define one.

@kostola My point is that we need to understand if the production-ready configuration is a priority for Managed Connectors before we get to Service Preview.

If it's a priority for them and they will deliver it before our own service preview, that's great and we need to schedule some work to remove this hack.

If it's not a priority for them and we have to go to our own Service Preview with this hack in place, then there is work for us to do to make sure this hack doesn't take our own Service offline. 

AFAICS we're not recording the need to address these points ^ beyond this PR. I don't want us to just merge this and then forget about it for a few months.

Does that make sense?

dev/ManagedConnectors.md
```diff
@@ -144,6 +144,89 @@ ocm login --token <offline_token>
...
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: cos-update-by-delete
+  namespace: cos
+spec:
+  concurrencyPolicy: Forbid
+  schedule: "@hourly"
```
robpblake, line 200: @kostola What about we have this run once per day, Mon-Fri, before everyone gets into the office, so perhaps 7 a.m UTC. That means we can start the day with the latest version of MC and then verify it's still running before we start the working day.

WDYT?

kostola: Good idea...I will update this :+1:

kostola: updated :+1:

dev/ManagedConnectors.md
kostola: Updated (I used "delete" which is the exact wording)

kostola:
> Does that make sense?
@robpblake yes I see your point but I think we could merge this PR and move the discussion about how this fits in our Service Preview architecture to be part of the global discussion about the whole Service Preview setup.

robpblake (approved):

PR #475, opened 2022-03-17
MGDOBR-454: Work is never removed from the queue if the entity has been deleted

manstis:
[https://issues.redhat.com/browse/MGDOBR-454](https://issues.redhat.com/browse/MGDOBR-454)

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.

</details>

### DIFF
--- a/manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
+++ b/manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
@@ -55,13 +55,8 @@ public class BridgesServiceImpl implements BridgesService {
     @Inject
     WorkManager workManager;
 
-    @Transactional
     @Override
     public Bridge createBridge(String customerId, BridgeRequest bridgeRequest) {
-        if (bridgeDAO.findByNameAndCustomerId(bridgeRequest.getName(), customerId) != null) {
-            throw new AlreadyExistingItemException(String.format("Bridge with name '%s' already exists for customer with id '%s'", bridgeRequest.getName(), customerId));
-        }
-
         Bridge bridge = doCreateBridge(customerId, bridgeRequest);
 
         LOGGER.info("Bridge with id '{}' has been created for customer '{}'", bridge.getId(), bridge.getCustomerId());
@@ -73,6 +68,10 @@ public class BridgesServiceImpl implements BridgesService {
 
     @Transactional
     protected Bridge doCreateBridge(String customerId, BridgeRequest bridgeRequest) {
+        if (bridgeDAO.findByNameAndCustomerId(bridgeRequest.getName(), customerId) != null) {
+            throw new AlreadyExistingItemException(String.format("Bridge with name '%s' already exists for customer with id '%s'", bridgeRequest.getName(), customerId));
+        }
+
         Bridge bridge = bridgeRequest.toEntity();
         bridge.setStatus(ManagedResourceStatus.ACCEPTED);
         bridge.setSubmittedAt(ZonedDateTime.now(ZoneOffset.UTC));
--- a/manager/src/main/java/com/redhat/service/bridge/manager/workers/resources/AbstractWorker.java
+++ b/manager/src/main/java/com/redhat/service/bridge/manager/workers/resources/AbstractWorker.java
@@ -43,6 +43,7 @@ public abstract class AbstractWorker<T extends ManagedResource> implements Worke
         T managedResource = load(work);
         if (Objects.isNull(managedResource)) {
             //Work has been scheduled but cannot be found. Something (horribly) wrong has happened.
+            workManager.complete(work);
             String message = String.format("Resource of type '%s' with id '%s' no longer exists in the database.", work.getType(), work.getManagedResourceId());
             throw new IllegalStateException(message);
         }
--- a/manager/src/test/java/com/redhat/service/bridge/manager/BridgesServiceTest.java
+++ b/manager/src/test/java/com/redhat/service/bridge/manager/BridgesServiceTest.java
@@ -1,7 +1,6 @@
 package com.redhat.service.bridge.manager;
 
 import java.time.ZonedDateTime;
-import java.util.ArrayList;
 import java.util.List;
 
 import javax.inject.Inject;
@@ -16,6 +15,7 @@ import com.redhat.service.bridge.infra.models.dto.ManagedResourceStatus;
 import com.redhat.service.bridge.manager.api.models.requests.BridgeRequest;
 import com.redhat.service.bridge.manager.models.Bridge;
 import com.redhat.service.bridge.manager.utils.DatabaseManagerUtils;
+import com.redhat.service.bridge.manager.utils.TestUtils;
 import com.redhat.service.bridge.test.resource.PostgresResource;
 
 import io.quarkus.test.common.QuarkusTestResource;
@@ -23,10 +23,8 @@ import io.quarkus.test.junit.QuarkusTest;
 import io.quarkus.test.junit.TestProfile;
 import io.quarkus.test.junit.mockito.InjectMock;
 
-import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
-import static org.awaitility.Awaitility.await;
 
 @QuarkusTest
 @TestProfile(WorkerSchedulerProfile.class)
@@ -84,10 +82,7 @@ public class BridgesServiceTest {
         Bridge bridge = bridgesService.createBridge(TestConstants.DEFAULT_CUSTOMER_ID, request);
 
         //Wait for Workers to complete
-        await().atMost(5, SECONDS).untilAsserted(() -> {
-            List<Bridge> bridges = bridgesService.findByShardIdWithReadyDependencies(TestConstants.SHARD_ID);
-            assertThat(bridges.size()).isEqualTo(1);
-        });
+        TestUtils.waitForBridgeToBeReady(bridgesService);
 
         Bridge retrievedBridge = bridgesService.getBridge(bridge.getId(), TestConstants.DEFAULT_CUSTOMER_ID);
         assertThat(retrievedBridge).isNotNull();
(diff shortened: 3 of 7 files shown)


manstis:
It's plausible an integration type test could be written too.

Let me have a play while waiting for you to look at the PR.

manager/src/main/java/com/redhat/service/bridge/manager/workers/resources/AbstractWorker.java
```diff
@@ -43,6 +43,7 @@ public T handleWork(Work work) {
         T managedResource = load(work);
         if (Objects.isNull(managedResource)) {
             //Work has been scheduled but cannot be found. Something (horribly) wrong has happened.
+            workManager.complete(work);
```
robpblake, line 46: @manstis We need to remove the `@Transactional` annotation from `BridgeServiceImpl.createBridge()` as it's currently leading to dirty reads in our `AbstractWorker`. More on dirty reads here: https://www.sqlshack.com/dirty-reads-and-the-read-uncommitted-isolation-level/

The problem is that the `ManagedResource` instance and `Work` are being created in the same transaction. So the following is possible:

- Transaction 1: writes the `Work` item to the table (but not yet the `ManagedResource`)
- Transaction 2 (work queue), reads the `Work` item and starts to execute. It checks for the `ManagedResource` which has not yet been written and sees it as missing. It then deletes the `Work` item
- Transaction 1: commits 

To fix this, I'd suggest the following changes:

Update to look `BridgeServiceImpl.createBridge` to look like this:

```
    @Override
    public Bridge createBridge(String customerId, BridgeRequest bridgeRequest) {
        
        Bridge bridge = doCreateBridge(customerId, bridgeRequest);

        LOGGER.info("Bridge with id '{}' has been created for customer '{}'", bridge.getId(), bridge.getCustomerId());

        workManager.schedule(bridge);

        return bridge;
    }
```

`BridgeServiceImpl.doCreateBridge()` to look like this:

```
@Transactional
    protected Bridge doCreateBridge(String customerId, BridgeRequest bridgeRequest) {
if (bridgeDAO.findByNameAndCustomerId(bridgeRequest.getName(), customerId) != null) {
            throw new AlreadyExistingItemException(String.format("Bridge with name '%s' already exists for customer with id '%s'", bridgeRequest.getName(), customerId));
        }
        
        Bridge bridge = bridgeRequest.toEntity();
        bridge.setStatus(ManagedResourceStatus.ACCEPTED);
        bridge.setSubmittedAt(ZonedDateTime.now(ZoneOffset.UTC));
        bridge.setCustomerId(customerId);
        bridge.setShardId(shardService.getAssignedShardId(bridge.getId()));
        bridgeDAO.persist(bridge);

        return bridge;
}
```
This will mean that the `ManagedResource` instance is comitted to the DB (or not) before the `Work` item.

robpblake (changes requested):
@manstis Please see my comment. Happy to jump on a call if you need more context.

cheers,
Rob

manager/src/main/java/com/redhat/service/bridge/manager/workers/resources/AbstractWorker.java
manstis: Perfect sense and consistent with deletion of Bridges and the creation/deletion of Processors.

manstis: Fixed in new commit.

manstis:
@sutaakar I added you as QE (as you have OB experience) however you may like to introduce @jomarko to things?

robpblake (approved):

manstis:
I tested this locally by creating a `Processor` with `Slack` action with duff credentials.

While the re-try loop was executing I deleted both `Connector` and `Processor` from their corresponding database tables.

I verified work still existed in the work table.

When the re-try loop was next attempted I verified both:
- The expected `java.lang.IllegalStateException: Resource of type '....Processor' with id 'xxx' no longer exists in the database` error was thrown.
- There were no records left in the work database table.

manager/src/main/java/com/redhat/service/bridge/manager/workers/resources/AbstractWorker.java
robpblake: @manstis I wanted to come back to update you on this after a quick discussion between @lucamolteni and I on Friday.

@lucamolteni correctly pointed out that Postgres doesn't support `READ_UNCOMMITTED` transaction isolation, so the dirty ready problem I called out is not correct. More here: https://www.postgresql.org/docs/12/transaction-iso.html

Instead, the problem was being caused by the immediate invocation of the vertX worker, which you have addressed in #480 

When the vertX worker was called immediately, because it was not within the same transaction as the one writing the initial data to the DB for the `Bridge` & ` Work`, it did not find the resource and hence reported that it did not exist. This is what I mistook for a dirty read.

Looking at the code we have now, I see no reason why we can't simply collapse the `doCreateBridge` into `createBridge` and `doDeleteBridge` into `deleteBridge`, placing the `Transactional` annotation back onto the `createBridge`/`deleteBridge` methods.

Apologies for making the wrong suggestion in the first place.

cheers,

Rob

manstis: LOL.. thank-you for recollecting better than myself. 

I had an impossible time trying to explain it to @lucamolteni earlier.

The change herein makes provisioning of bridges consistent with that of processors and with the de-provisioning of both. We could probably now collapse the methods (as we no longer execute the work immediately). However, since this PR brings consistency, I'd prefer to do as a separate task.

manager/src/test/java/com/redhat/service/bridge/manager/workers/resources/AbstractWorkerTest.java
```diff
@@ -0,0 +1,64 @@
...
+    BridgeWorker worker;
+
+    @Test
+    @Transactional
+    void workIsCompletedWhenResourceIsNotFound() {
+        Bridge bridge = new Bridge(TestConstants.DEFAULT_BRIDGE_NAME);
+        bridge.setCustomerId(TestConstants.DEFAULT_CUSTOMER_ID);
+        bridge.setStatus(ManagedResourceStatus.ACCEPTED);
+        bridge.setSubmittedAt(ZonedDateTime.now());
+        bridge.setId(RESOURCE_ID);
```
sutaakar, line 52: Is this code needed? The `bridge` instance is used below just to retrieve submitted time - that can be set directly

manstis: Yes, the call [here](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/475/files/6429a6ec9d4d9f2717eaf2e30ab188c6ac44c70c#diff-95c3db7d40e5c0e6398a83f2285f23a90e93db0975bafe77eebe504afdad8109R59) needs other properties set in order for it to function correctly.

sutaakar:
```
    @Test
    @Transactional
    void workIsCompletedWhenResourceIsNotFound() {
        Work work = new Work();
        work.setType(Bridge.class.getName());
        work.setSubmittedAt(ZonedDateTime.now());
        work.setManagedResourceId(RESOURCE_ID);

        assertThatThrownBy(() -> worker.handleWork(work)).isInstanceOf(IllegalStateException.class);

        verify(workManager).complete(work);
    }
```

manstis: Ha, erm, no not really.. I am clearly getting my PRs confused. Removed surplus setup.

It might well have been used before I [updated the test](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/475#issuecomment-1070876282).

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
```diff
@@ -73,6 +68,10 @@ public Bridge createBridge(String customerId, BridgeRequest bridgeRequest) {
 
     @Transactional
     protected Bridge doCreateBridge(String customerId, BridgeRequest bridgeRequest) {
+        if (bridgeDAO.findByNameAndCustomerId(bridgeRequest.getName(), customerId) != null) {
```
lucamolteni, line 71: I'm not sure this changes anything. 

First `createBridge` created a transaction and `doCreateBridge` attached to the previous one (the standard behaviour for `REQUIRED`) and now there's no transaction outside and these lines are moved inside `doCreateBridge`. 

It's still one transaction and the same one

manager/src/test/java/com/redhat/service/bridge/manager/BridgesServiceTest.java
```diff
@@ -185,6 +203,12 @@ public void getBridge() {
         BridgeRequest request = new BridgeRequest(TestConstants.DEFAULT_BRIDGE_NAME);
         Bridge bridge = bridgesService.createBridge(TestConstants.DEFAULT_CUSTOMER_ID, request);
 
+        //Wait for Workers to complete
```
lucamolteni, line 206: Was this piece of code copy and pasted? Extract to common method please

lucamolteni (changes requested):

manager/src/main/java/com/redhat/service/bridge/manager/BridgesServiceImpl.java
manstis: The call to schedule work is no longer inside the transaction and the code _structure_ is identical `deleteBridge(..)`, `createProcessor(..)` and `deleteProcessor(..)`. @robpblake has already proposed we can now dispense with the separate `X` and `doX` methods _consistently_ across both bridges and processors. I'd prefer to create another JIRA to handle the change in one go than pollute this PR (for removal of the `Work`). WDYT?

lucamolteni: I'm not sure you get what you mean, but it should be fine anyway

jiripetrlik (approved):
Looks good to me. @manstis Thank you for additional steps for testing in the related JIRA!

manstis:
@sutaakar Are you OK with this PR now?

sutaakar (approved):
Yes, looks good to me

PR #521, opened 2022-03-24
MGDOBR-336 Use Pod Name as Worker Identity to help Debug Resilient Provisioning.

lucamolteni:
This PR was tested locally on minikube, I checked the variable `hostname` was returning the name of the pod. 

[MGDOBR-336](https://issues.redhat.com/browse/MGDOBR-336)

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.

</details>


manager/src/main/java/com/redhat/service/bridge/manager/workers/id/WorkerIdProvider.java
```diff
@@ -4,13 +4,33 @@
...
-    private static final String WORKER_ID = UUID.randomUUID().toString();
+    private static final String RANDOM_STATIC_WORKER_ID = UUID.randomUUID().toString();
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(WorkerIdProvider.class);
+
+    private String findPodNameFromKubernetes() {
+        String hostname = System.getenv("HOSTNAME");
+        LOGGER.trace("POD id from hostname: {}", hostname);
+
+        if (hostname != null && !hostname.isEmpty() && !"localhost".equals(hostname)) {
```
manstis, line 20: There does not appear to be any guarantee that the environment variable `HOSTNAME` returns a lower case `String`. For the sake of avoiding future potential issues where someone has it set as `LOCALHOST` etc it'd be advisable to use `equalsIgnoreCase(..)`.

I don't suppose you want any unit tests for the change you've made too?

manager/src/main/java/com/redhat/service/bridge/manager/workers/id/WorkerIdProvider.java
```diff
@@ -4,13 +4,33 @@
...
+            return null;
+        }
+    }
 
     @Produces
     public String getWorkerId() {
-        return WORKER_ID;
+        String podNameFromKubernetes = findPodNameFromKubernetes();
+        if (podNameFromKubernetes == null) {
+            return RANDOM_STATIC_WORKER_ID;
```
r00ta, line 31: maybe it would be good to log this. 
Just an headsup: @manstis I think it would be great if in your logging PR we add the identifier of the manager pod/instance in all the logs. Not something we have to include in your PR, just saying let's consider it in another ticket

r00ta (approved):

manager/src/main/java/com/redhat/service/bridge/manager/workers/id/WorkerIdProvider.java
manstis: Great idea @r00ta.. I've been trying to think of more data for the base message (sub-classes can have more.. but my PR is just the beginning and something _concrete_ to discuss vs trying to discuss something hypothetically) :+1:

manstis: @r00ta I'll wait for this PR to be merged as I need to move `WorkerIdProvider` out of `manager` and into somewhere more accessible (less dependent). Perhaps `infra` would be a good home for it.

manager/src/main/java/com/redhat/service/bridge/manager/workers/id/WorkerIdProvider.java
```diff
@@ -4,13 +4,33 @@
...
+import org.slf4j.LoggerFactory;
+
 public class WorkerIdProvider {
 
-    private static final String WORKER_ID = UUID.randomUUID().toString();
+    private static final String RANDOM_STATIC_WORKER_ID = UUID.randomUUID().toString();
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(WorkerIdProvider.class);
+
+    private String findPodNameFromKubernetes() {
```
robpblake, line 16: Nitpick: This isn't really anything to do with Kubernetes, so should we name it a bit better e.g `findWorkerNameFromHost`?

manager/src/main/java/com/redhat/service/bridge/manager/workers/id/WorkerIdProvider.java
```diff
@@ -4,13 +4,33 @@
 
 import javax.enterprise.inject.Produces;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class WorkerIdProvider {
```
robpblake, line 10: @lucamolteni Can we add a unit test for `WorkerIdProvider` please now that we have logic in it 👍 
cheers,

Rob

lucamolteni: Testing with environment variable is complicated. I could abstract the sysenv part and test it WDYT?

robpblake: @lucamolteni Or we could simplify the implementation to look like this:

```
@ApplicationScoped
public class WorkerIdProvider {

    @ConfigProperty(name = "event-bridge.resources.worker.id")
    private String workerId;

    public String getWorkerId() {
        return workerId;
    }
}
```

With the following in the `application.properties`:

```
event-bridge.resources.worker.id=${EVENT_BRIDGE_RESOURCES_WORKER_ID:${quarkus.uuid}}
```

And then add the following to our `Deployment` for the Fleet Manager:

```
env:
     - name: EVENT_BRIDGE_RESOURCES_WORKER_ID
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
```

If you want to have control over the worker id, set the `EVENT_BRIDGE_RESOURCES_WORKER_ID` environment variable. Otherwise let Quarkus generate a random UUID.

At that point I'm not event sure how much value the `WorkerIdProvider` brings over simply injecting the value of `event-bridge.resources.worker.id` directly into `WorkManagerImpl`.

WDYT?

manager/src/main/java/com/redhat/service/bridge/manager/workers/id/WorkerIdProvider.java
jomarko: sounds as good suggestion

kustomize/base/manager/resources/deploy.yaml
```diff
@@ -18,6 +18,11 @@ spec:
         - image: event-bridge-manager
           imagePullPolicy: IfNotPresent
           name: event-bridge-manager
+          env:
+            - name: EVENT_BRIDGE_RESOURCES_WORKER_ID
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
```
manstis, line 25: I'm no expert :smile: 

Can you explain how `metadata.name` will resolve to a **different** value for each container running the `manager`? Won't `metadata.name` always resolve to `event-bridge-manager` (i.e. the name of the Kubernetes resource for our `manager`).

manstis (approved):
From what I understand I am approving.

I've added a question about how we guarantee a unique id per container.

kustomize/base/manager/resources/deploy.yaml
manstis: I had a hunt around and `metadata.name` may furnish us with a unique Pod name however for that to be useful for us we'd either need (1) a single container per Pod, (2) use something other than `metadata.name`. Our `WorkerId` is used to pick up a list of `Work` for processing. If we have the risk of the same value across multiple containers within a Pod then it's possible we'd have multiple instances of `WorkManagerImpl` processing the same work at the same time... try debugging that!

manstis: I live and learn: https://kubernetes.io/docs/concepts/workloads/pods/#workload-resources-for-managing-pods

"**Pods that run a single container**. The "one-container-per-Pod" model is the most common Kubernetes use case; in this case, you can think of a Pod as a wrapper around a single container"

jomarko:
@manstis @lucamolteni hello, the code looks fine to me. However I am not sure how to check manually these changes. Could you please guide me, if it is possible, how to check the changes manually? Can I see it in some logs of the openbridge/smart events container on minikube?

r00ta (approved):

manstis:
> @manstis @lucamolteni hello, the code looks fine to me. However I am not sure how to check manually these changes. Could you please guide me, if it is possible, how to check the changes manually? Can I see it in some logs of the openbridge/smart events container on minikube?

You cannot test is anywhere at the moment.

When https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/541 is merged we will be logging `WorkerId` (we only log `WorkerId` [here](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/manager/src/main/java/com/redhat/service/bridge/manager/workers/resources/AbstractWorker.java#L72) - and similarly for deprovisioning - at the moment which will only be visible when an unrecoverable error occurs).

IIUC even then you'll only see the UUID version of `WorkerId`. The Pod name version will only be visible in logs in a _real_ environment (unless running locally with kubernetes would use the configuration defined [here](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/521/files#diff-013c5f3d48c38e2fff2d1d2a98f54895a589ba412abecea7c52b8bb616b00818R25)).

jomarko (approved):
@manstis thank you for comment. Then I think can not do more than approve. Thank you @lucamolteni .

PR #558, opened 2022-03-31
MGDOBR-483: move dev env to sso.redhat.com

r00ta:
[MGDOBR-483](https://issues.redhat.com/browse/MGDOBR-483)

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.

</details>


dev/bin/interaction/environment-example
```diff
@@ -2,8 +2,14 @@
 # It can be used to configure required variables for your local environment
 # without the need of exporting them every time, like offline tokens etc.
 # If you want to use it, make a copy of this named "environment" in this same folder.
+# The dev and stable environment use sso.redhat.com. So if you want to target such environments, set TARGET_ENVIRONMENT=remote and the
+# variable OPENSHIFT_OFFLINE_TOKEN accordingly (retrieve the token from https://console.redhat.com/openshift/token).
```
danielezonca, line 6: Don't we have a script also for this? I know it is a single command but it helps people :)

dev/bin/interaction/environment-example
```diff
@@ -2,8 +2,14 @@
...
+# If you target your local environment, then replace the variables KEYCLOAK_USERNAME and KEYCLOAK_PASSWORD.
```
danielezonca, line 7: Is it possible to configure also the local instance to use an offline token? I would like to keep the configuration aligned

r00ta: the developer has to edit this file in any case. If the developer has to extract the offline token from the local keycloak instance he has to run an additional command with the username and the password, while in this script he can directly use the username and the password. 
It's a dev script at the end, but if you want we can do it

r00ta:
Hi @robpblake @kostola @lucamolteni @jomarko @manstis are you happy with the changes in the pipelines and the documentation?

dev/bin/interaction/environment-example
```diff
@@ -2,8 +2,14 @@
...
 
 MINIKUBE_IP=$(minikube ip)
+TARGET_ENVIRONMENT=local # valid values [local,remote]
```
manstis, line 10: Technically permissible values are `local` and "anything other than `local`" :smile:

r00ta:
> I don't use `/kustomize/startMinikubeDeployLocalDev.sh` running both `/dev/bin/manager-start.sh` and `/dev/bin/shard-start.sh` locally. Will that use SSO too?

Every local setup will use the local keycloak instance. Only dev and stable will use sso.redhat.com

dev/bin/interaction/environment-example
r00ta: fixed

dev/bin/interaction/environment-example
danielezonca: My main concern is, if I have understood correctly, that based on `TARGET_ENVIRONMENT` value, the scripts are going to use `OPENSHIFT_OFFLINE_TOKEN` or `KEYCLOAK_USERNAME`/`KEYCLOAK_PASSWORD`.
In general if we change the configuration to use the token also locally I think we should be able to get rid of all of this and keep two variables only: `SSO_REDHAT_URL` and `OPENSHIFT_OFFLINE_TOKEN` (no need to specify the target env nor keycloak variables).
Same for the if/else in `get-token.sh`

I don't know how hard is to do this different configuration so I'm fine to proceed with this PR and do a followup ticket for this cleanup (if we agree to do it).

kustomize/overlays/prod/manager/patches/deploy-config.yaml
```diff
@@ -8,7 +8,7 @@ data:
   EVENT_BRIDGE_DB_HOST: "event-bridge-db"
   EVENT_BRIDGE_DB_PORT: "5432"
   EVENT_BRIDGE_DB_SCHEMA: "event-bridge"
-  EVENT_BRIDGE_SSO_URL: "https://keycloak-event-bridge-prod.apps.openbridge-dev.fdvn.p1.openshiftapps.com/auth/realms/event-bridge-fm"
+  EVENT_BRIDGE_SSO_URL: "https://sso.redhat.com/auth/realms/redhat-external"
   EVENT_BRIDGE_SSO_CLIENT_ID: "event-bridge"
```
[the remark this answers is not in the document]
r00ta: afaik yep

PR #569, opened 2022-04-01
MGDOBR-322: refactoring of actions related classes into smaller modules

kostola:
[MGDOBR-322](https://issues.redhat.com/browse/MGDOBR-322)

## PR owner checklist

Please make sure that your PR meets the following requirements:

- [ ] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/main/kogito-build/kogito-ide-config).  
  *Run `mvn clean verify -DskipTests` so that imports are correctly set.*
- [ ] Your commit messages are clear and reference the JIRA issue e.g: "[MGDOBR-1] - $clear_explanation_of_what_you_did"
- [ ] All new functionality is tested
- [ ] Pull Request title is properly formatted: `MGDOBR-XYZ Subject`
- [ ] Pull Request contains link to the JIRA issue
- [ ] Pull Request contains link to any dependent or related Pull Request
- [ ] Pull Request contains description of the issue
- [ ] Pull Request does not include fixes for issues other than the main ticket

## PR reviewer(s) checklist

- [ ] Check if code and Github action workflows have been modified and if the modification is safe
- [ ] If the modification is safe, add the `safe to test` label

<details>
<summary>
How to trigger pipelines and use the bots:
</summary>

- <b>Run the end to end pipeline</b>  
  Annotate the pull request with the label: `safe to test`. If you want to run the pipeline again, remove and add it again.

- <b>Rebase the pull request</b>  
  Comment with: `/rebase`.

</details>

PR #590, opened 2022-04-07
MGDOBR-322: actions module refactoring

kostola:

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.

</details>

## Description

The goals of this PR are:

1. dismantle the `actions` module: it was intended to share the same code between manager and executor but it was designed when the whole app was a monolithic PoC. Now it shares action definition, validation and execution code between different services which is not totally needed (e.g. manager does validation and resolution, executor does validation and execution) and it brings unwanted dependencies in the manager (Kafka client).

2. dismantle the concepts of `ActionProvider` and `ActionProviderFactory`: they aggregate together beans that has totally different functionalities and belongs to different services: validation, resolution and execution. We need to create different independent factories for each type of functionality.

3. have a clean dependency tree


kostola:
Same goal of dismantling the action module and cleaning the dependency tree as #569 but with less modules

processor/core/src/main/java/com/redhat/service/bridge/processor/actions/kafkatopic/KafkaTopicActionBean.java
```diff
@@ -0,0 +1,11 @@
+package com.redhat.service.bridge.processor.actions.kafkatopic;
+
+import com.redhat.service.bridge.processor.actions.ActionBean;
+
+public interface KafkaTopicActionBean extends ActionBean {
```
robpblake, line 5: @kostola I'm not sure I really see the value in `KafkaTopicActionBean` and `KafakTopicAction`. IMHO, we can simplify and have `KafkaTopicAction` implement `ActionBean`.

Same for all the other Actions.

processor/core/src/main/java/com/redhat/service/bridge/processor/actions/AbstractActionBeanFactory.java
```diff
@@ -0,0 +1,31 @@
...
+    protected AbstractActionBeanFactory(Class<T> instanceClass) {
+        this.instanceClass = instanceClass;
+    }
+
+    public T get(String actionType) {
+        return getOptional(actionType)
+                .orElseThrow(() -> new ActionProviderException(String.format("No %s found for action type '%s'", instanceClass.getSimpleName(), actionType)));
+    }
+
+    public Optional<T> getOptional(String actionType) {
```
robpblake, line 26: @kostola Personally I don't think we need the `Optional` method on this class. I'd say we support `get` only as to me it makes it clearer as to expectations on what should be provided each time we add additional Actions to the platform. For example: we shouldn't have the semantics in the Shard where I can optionally as for an `ActionInvoker`, it has to exist.

robpblake (commented):
@kostola I think this is pretty nice 👍  Good job.

My main comment is: When I'm adding a new Action to the platform, how do I know what I need to do? At the minute I have to understand that I need to write a `Validator`, `Transformer` (maybe) and also an `Invoker`

My suggestion is that we have a couple of interfaces that collect together what authors of Actions need to do, to make it clearer. So I'm thinking something like this (names, signatures all just suggestions btw)

```
public interface ActionConfiguration {

   ActionValidator getValidator();
   ActionTransformer getTransformer();
}
```

and then 

```
public interface ActionRuntime {
  ActionInvokerBuilder getInvokerBuilder();
}
```
It has a couple of benefits that I can see:

- Makes it much easier as an Action author to understand what I need to write: configuration and then a runtime part
- Reduces the number of Factories we need to two: ActionConfigurationFactory and ActionRuntimeFactory

cheers,

Rob

processor/core/src/main/java/com/redhat/service/bridge/processor/actions/kafkatopic/KafkaTopicActionBean.java
kostola: I had the same idea but I wasn't sure that it could cause confusion so I went with the split.

@r00ta @lucamolteni @danielezonca what is your opinion on this? Let's go with the majority :smile:

robpblake: @kostola What confusion were you trying to avoid with the split?

robpblake:
> 

I respectfully disagree. They're not unrelated: all of the components are needed to implement the required functionality of the Action e.g I shouldn't be able to add an Action to the platform without providing a Validator (I can of course re-use an existing implementation if it gives me the functionality I need)

My concern about the `README.md` approach is:

1. People need to know to read it
2. People reviewing the PR need to know to read it and check all the components are implemented
3. If certain components are not implemented, it needs to be clear why, both at the time of the review and going forwards when other people come to look at the code.

We can avoid this overhead if we have an `interface` that defines the contract for what you need to provide to add an Action to the platform. IMHO, simple default implementations like https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/actions/src/main/java/com/redhat/service/bridge/actions/ActionTransformer.java#L11 than can easily be reused are much clearer than a missing implementation.

If I'm the only that feels this way, then let's proceed with the decomposed approach.

cheers,

Rob

kostola:
@robpblake I just pushed a refactoring that groups the access to action beans in `ActionConfigurator` (part of `processor-manager` and containing `Validator`s, `Resolver`s and `Connector`s) and `ActionRuntime` (part of `processor-executor` and containing `InvokerBuilder`s).

Imho this is the right level of compromise between having two beans that makes explicit what needs to be implemented (which to me makes sense) and not having two such beans *for every action* (which to me is unnecessarily redundant).

If you want I could add a few lines of javadoc in both "main" beans explaining what each type of sub-bean is intended to do and when it should be implemented.

Wdyt?

robpblake (approved):
Thank you @kostola . 

Non-blocking comment from me on module names: are `configuration` and `runtime` better than `manager` and `executor`

There is also the unresolved discussion on `KafkaTopicAction` and `KafkaTopicActionBean`. I personally don't think we need those, but again, not a blocking comment from me.

cheers,

Rob

bom/pom.xml
```diff
@@ -177,7 +177,64 @@
         <version>${project.version}</version>
         <classifier>javadoc</classifier>
       </dependency>
+
+      <!-- ===== ACTIONS ===== -->
```
r00ta, line 181: Remove 
```
        <groupId>com.redhat.service.bridge</groupId>
        <artifactId>actions</artifactId>
        ...
```
from line 128?

processor/executor/src/main/java/com/redhat/service/bridge/processor/actions/kafkatopic/KafkaTopicActionInvokerBuilder.java
```diff
@@ -12,22 +12,16 @@
...
+import com.redhat.service.bridge.processor.actions.ActionInvoker;
+import com.redhat.service.bridge.processor.actions.ActionInvokerBuilder;
 
 @ApplicationScoped
-public class KafkaTopicAction implements InvokableActionProvider {
-
-    public static final String TYPE = "KafkaTopic";
-
-    public static final String TOPIC_PARAM = "topic";
+public class KafkaTopicActionInvokerBuilder implements KafkaTopicActionBean, ActionInvokerBuilder {
```
r00ta, line 22: I did not get why we need the `KafkaTopicActionBean` and `KafkaTopicAction` (same for all the actions). 

iiuc, what you are trying to do with this abstraction can be achieved with generics removing the `..Bean` class for all the actions, for example you might turn this Invoker into something like 

```
public class KafkaTopicActionInvokerBuilder implements ActionInvokerBuilder<KafkaTopicAction> {
```

and change some classes accordingly.

processor/manager/src/main/java/com/redhat/service/bridge/processor/actions/ActionConfiguratorImpl.java
```diff
@@ -0,0 +1,56 @@
...
+        return getOptional(connectors, actionType);
+    }
+
+    private static <T extends ActionBean> Optional<T> getOptional(Instance<T> instances, String actionType) {
+        return instances.stream()
+                .filter(a -> a.accept(actionType))
+                .findFirst();
+    }
+
+    Collection<ActionValidator> getValidators() {
```
vaibhavjainwiz, line 45: where these functions are used?

vaibhavjainwiz (changes requested):

processor/executor/src/main/java/com/redhat/service/bridge/processor/actions/kafkatopic/KafkaTopicActionInvokerBuilder.java
kostola: I agree that the split between `...Action` and `...ActionBean` is redundant and I just refactored it to have only interfaces.

However, the generics-based solution you are proposing is not enough because we don't directly inject the beans but, instead, we have an `ActionConfigurator` "factory" where they are retrieved by type. Thus, we need the `getType` and `accept` methods defined in `ActionBean` and the action-specific interface is explicitly used to set the correct type in all the beans belonging to the same action.

processor/executor/src/main/java/com/redhat/service/bridge/processor/actions/kafkatopic/KafkaTopicActionInvoker.java
```diff
@@ -1,26 +1,26 @@
...
 
-public class KafkaTopicInvoker implements ActionInvoker {
+public class KafkaTopicActionInvoker implements ActionInvoker {
 
-    private static final Logger LOG = LoggerFactory.getLogger(KafkaTopicInvoker.class);
+    private static final Logger LOG = LoggerFactory.getLogger(KafkaTopicActionInvoker.class);
 
     private final String topic;
 
     private final ProcessorDTO processor;
```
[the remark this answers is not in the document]
kostola: That's a nice observation but I am not changing functionality with this refactoring so I'd keep it there since it was there before. We can think about this separately.

processor/manager/src/main/java/com/redhat/service/bridge/processor/actions/ActionConfiguratorImpl.java
kostola: They're used in `ActionConfiguratorImplTest` to ensure that there are no unexpected beans that are not covered by that test.

bom/pom.xml
kostola: Good catch. Done :+1:

kostola:
@robpblake thanks for the feedback.

I removed the split between `...Action` and `...ActionBean` and renamed the modules to `processor-configuration` and `processor-runtime` as you suggested.

I think we're good now :+1:

processor/manager/src/main/java/com/redhat/service/bridge/processor/actions/ActionConfiguratorImpl.java
vaibhavjainwiz: we generally don't write code to satisfy test case.

vaibhavjainwiz (approved):

processor/core/src/main/java/com/redhat/service/bridge/processor/actions/kafkatopic/KafkaTopicActionBean.java
kostola: Resolving this because this part has been refactored

processor/core/src/main/java/com/redhat/service/bridge/processor/actions/AbstractActionBeanFactory.java
kostola: Resolving this because this part has been refactored

lucamolteni (approved):

r00ta (changes requested):
Please change the CD.yaml pipeline accordingly

manager/src/main/java/com/redhat/service/bridge/manager/ActionServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+
+    @Override
+    public String getBridgeEndpoint(String bridgeId, String customerId) {
+        return bridgesService.getReadyBridge(bridgeId, customerId).getEndpoint();
+    }
+
+    @Override
+    public String getConnectorTopicName(String processorId) {
+        return resourceNamesProvider.getProcessorTopicName(processorId);
+    }
```
r00ta, line 26: I'm not sure I get why `ActionService` exposes `getBridgeEndpoint` and `getConnectorTopicName`. Are they related to actions?

PR #603, opened 2022-04-11
MGDOBR-543 Create a script to make the PR to deploy to staging

lucamolteni:
[Create a script to make the PR to deploy to staging](https://issues.redhat.com/browse/MGDOBR-543)

I'm not sure why this scripts puts the quote in the kustomization.yaml but not in the deploy-config.yaml. It shouldn't be a big problem though

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.

</details>

### DIFF
--- a/dev/bin/create-stable-pr.sh
+++ b/dev/bin/create-stable-pr.sh
@@ -1,5 +1,16 @@
 #!/bin/sh
 
+if ! command -v yq &> /dev/null
+then
+    echo "yq could not be found, install it from https://github.com/mikefarah/yq"
+    exit
+fi
+
+if (( $# == 0 )); then
+    echo "Usage: ./create-stable-br <GIT_DEV_DEPLOY_HASH>"
+    exit
+fi
+
 COMMIT_HASH=${1}
 COMMIT_DESCRIPTION=$(git log --pretty=format:"%h %s" -1 $COMMIT_HASH)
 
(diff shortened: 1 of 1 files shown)


dev/bin/create-stable-pr.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+
+COMMIT_HASH=${1}
+COMMIT_DESCRIPTION=$(git log --pretty=format:"%h %s" -1 $COMMIT_HASH)
+
+echo "DEV COMMIT: $COMMIT_HASH"
+echo "$COMMIT_DESCRIPTION"
+
+export MANAGER_NEW_TAG=$(git show $COMMIT_HASH:kustomize/overlays/prod/kustomization.yaml | yq .images[0].newTag)
```
radtriste, line 9: where does `yq` comes from ? I think it is not installed default on Fedora

Maybe add some doc to install it as a dependency (like `jq`)

lucamolteni: @radtriste added GitHub link. I wouldn't put it in the doc as it's needed only for this script

radtriste: IMO as long as we use a tool in one of the dev scripts, then we should document in the dev README file (and maybe put `optional` mention)
It could be that the tool would be reuse later in some other script(s), that would avoid to have to think about adding it afterwards. It should be mentioned the first time we introduce it ;)

robpblake (commented):
@lucamolteni It would be good to have this documented somewhere otherwise I don't think people will know to use it. At the minute the best place I can think of is in the doc that @kostola put together for our discussion on promotion process from the other day.

cheers,

Rob

PR #629, opened 2022-04-14
MGDOBR-270: Event Sources PoC

kostola:
[MGDOBR-270](https://issues.redhat.com/browse/MGDOBR-270)

## Description

This PR contains a PoC for the Event Sources feature. The focus at the moment is to have it working so no particular effort was made so far on test correctness and coverage.

## Example payload

The only supported source processor is `SlackSource`, which read messages from a specific slack channel and requires a Slack bot token to work.

Once a token with the right permissions is obtained (check the [Slack docs](https://api.slack.com/authentication/token-types#bot) for further info) here is an example payload to create the processor, by calling the existing Processor API in the manager:

```json
{
    "name": "slack-source",
    "source": {
        "type": "Slack",
        "parameters": {
            "channel": "channel-to-read-from",
            "token": "xoxb-..."
        }
    }
}
```


manager/src/main/java/com/redhat/service/smartevents/manager/api/models/responses/ProcessorResponse.java
```diff
@@ -36,7 +37,10 @@ public ProcessorResponse() {
     private String transformationTemplate;
 
     @JsonProperty("action")
-    private BaseAction action;
+    private Action action;
```
robpblake, line 40: @kostola I think this violates my expectations as a user of the API. For example, I submit this:

```
 {
    "name": "slack-source",
    "source": {
        "type": "Slack",
        "parameters": {
            "channel": "channel-to-read-from",
            "token": "xoxb-..."
        }
    }
}
```
but I then get back something with an `Action` in it:

```
{
    "name": "slack-source",
    "source": {
        "type": "Slack",
        "parameters": {
            "channel": "channel-to-read-from",
            "token": "xoxb-..."
        }
    },
   "action": "{
       "type" : "sendToBridge",
       ...
    }
}
```

The `action` is a purely internal implementation detail of how we have decided to currently ingest events into the bridge instance. The user shouldn't be aware of it IMHO.

My expectation as a user is that I get back something that represents only what I submitted e.g a `Source` configuration.

robpblake: Ignore me. It doesn't return the resolved action.

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
```diff
@@ -102,28 +103,35 @@ public Processor createProcessor(String bridgeId, String customerId, ProcessorRe
...
 
         Set<BaseFilter> requestedFilters = processorRequest.getFilters();
 
         String requestedTransformationTemplate = processorRequest.getTransformationTemplate();
-        BaseAction requestedAction = processorRequest.getAction();
 
-        BaseAction resolvedAction = actionConfigurator.getResolver(requestedAction.getType())
-                .map(resolver -> resolver.resolve(requestedAction, customerId, bridge.getId(), newProcessor.getId()))
-                .orElse(requestedAction);
+        boolean isSourceProcessor = processorRequest.getSource() != null;
```
robpblake, line 116: @kostola I think I would go with creating a `enum` `ProcessorType` and have a `getType()` call on the `Processor`. I can see a couple of major benefits:

- Rather than `getDefinition().getSource() != null`, code becomes much more readable `if processor.getType() == SOURCE`
- It makes it much easier for us to support querying on the API by Processor `type` e.g. `/api/v1/bridges/{id}/processors?type=source`

manager/src/main/java/com/redhat/service/smartevents/manager/api/models/responses/ProcessorResponse.java
robpblake: No, it does :-)

IMHO it shouldn't because it' a purely implementation detail.

manager/src/main/java/com/redhat/service/smartevents/manager/api/user/validators/processors/GatewayConstraintValidator.java
```diff
@@ -0,0 +1,107 @@
...
+    @Inject
+    GatewayConfigurator gatewayConfigurator;
+
+    @Override
+    public boolean isValid(ProcessorRequest value, ConstraintValidatorContext context) {
+        Action action = value.getAction();
+        Source source = value.getSource();
+
+        if (action == null && source == null || action != null && source != null) {
+            return false;
```
robpblake, line 46: @kostola I know this is a PoC, but we really need a clear error message here to the user that a Processor must either be a source or a sink.

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
```diff
@@ -102,28 +103,35 @@ public Processor createProcessor(String bridgeId, String customerId, ProcessorRe
...
+                ? new ProcessorDefinition(requestedFilters, requestedTransformationTemplate, processorRequest.getSource(), resolvedAction)
+                : new ProcessorDefinition(requestedFilters, requestedTransformationTemplate, processorRequest.getAction(), resolvedAction);
 
-        ProcessorDefinition definition = new ProcessorDefinition(requestedFilters, requestedTransformationTemplate, requestedAction, resolvedAction);
         newProcessor.setDefinition(definitionToJsonNode(definition));
 
         // Processor, Connector and Work should always be created in the same transaction
         processorDAO.persist(newProcessor);
-        connectorService.createConnectorEntity(newProcessor, requestedAction);
+        if (isSourceProcessor) {
```
robpblake, line 130: I think with the `getType()` on the `Processor` we don't need to change the interface of the `ConnectorService`

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/ConnectorWorker.java
```diff
@@ -193,4 +193,14 @@ protected ConnectorEntity doDeleteDependencies(ConnectorEntity connectorEntity)
         return connectorEntity;
     }
 
+    private static boolean isSourceConnector(ConnectorEntity connectorEntity) {
+        return connectorEntity.getConnectorType().contains("_source_");
```
robpblake, line 197: Is it more robust to do `connectorEntity.getProcess().getType()` ?

processor/configuration/src/main/java/com/redhat/service/smartevents/processor/sources/SourceConfigurator.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Get connector bean for specific source type.
+     * This bean is optional and must be implemented only for sources that requires
+     * a Managed Connector to work.
+     *
+     * @param sourceType desired source type
+     * @return {@link Optional} containing the bean if present, empty otherwise.
+     */
+    Optional<SourceConnector> getConnector(String sourceType);
```
robpblake, line 36: This shouldn't be `Optional` at the minute. We only support sources deployed as a Managed Connector currently.

processor/configuration/src/main/java/com/redhat/service/smartevents/processor/sources/SourceConfigurator.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Get resolver bean for specific source type.
+     * This bean is optional and must be implemented only for non-invokable sources
+     * that needs to be resolved to an invokable sources to be executed.
+     *
+     * @param sourceType desired source type
+     * @return {@link Optional} containing the bean if present, empty otherwise.
+     */
+    Optional<SourceResolver> getResolver(String sourceType);
```
robpblake, line 26: Not sure this should be `Optional` either. We must provide a fully configured `Input` action for each Source

processor/configuration/src/main/java/com/redhat/service/smartevents/processor/sources/SourceResolver.java
```diff
@@ -0,0 +1,9 @@
+package com.redhat.service.smartevents.processor.sources;
+
+import com.redhat.service.smartevents.infra.models.actions.Action;
+import com.redhat.service.smartevents.infra.models.actions.Source;
+
+public interface SourceResolver extends SourceBean {
+
+    Action resolve(Source source, String customerId, String bridgeId, String processorId);
```
robpblake, line 8: @kostola Do we need to constrain the returned type here more? In reality this can only be an `Input` action in your model, we probably shouldn't allow people to return any type of Action 🤔

processor/core/src/main/java/com/redhat/service/smartevents/processor/actions/input/InputAction.java
```diff
@@ -0,0 +1,15 @@
+package com.redhat.service.smartevents.processor.actions.input;
+
+import com.redhat.service.smartevents.processor.actions.ActionBean;
+
+public interface InputAction extends ActionBean {
```
robpblake, line 5: Some suggestions on other names perhaps to make the purpose of this `Action` clearer:

- `SourceAction`
- `IngestAction`

Just thinking out loud as `InputAction` is a little bit hazy as to it's intent.

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/input/InputActionInvoker.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public void onEvent(String event) {
+        LOG.info("InputActionInvoker::onEvent | {} - {} | {}", endpoint, cloudEventType, event);
+
+        HttpRequest<Buffer> request = client.postAbs(endpoint);
+        String token = oidcClient.getToken();
+        LOG.info("InputActionInvoker::onEvent | Token: {}", token);
+        if (token != null && !"".equals(token)) {
```
robpblake, line 52: The token is always required as we're sending to the ingress which requires auth.

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/input/InputActionInvoker.java
```diff
@@ -0,0 +1,85 @@
...
+        request.sendJsonObject(getPayload(event)).subscribe().with(
+                response -> LOG.info("RESPONSE {} {}", response.statusCode(), response.statusMessage()),
+                error -> LOG.error("ERROR", error));
+    }
+
+    private JsonObject getPayload(String event) {
+        try {
+            ObjectNode cloudEvent = MAPPER.createObjectNode();
+            cloudEvent.set("id", new TextNode(UUID.randomUUID().toString()));
+            cloudEvent.set("source", new TextNode("RHOAS"));
```
robpblake, line 65: I think it would be useful here to try and be more specific about the connector that produced this event if at all possible to help us trace back where things are coming from.

executor/src/main/java/com/redhat/service/smartevents/executor/impl/CloudEventExecutorImpl.java
```diff
@@ -0,0 +1,109 @@
...
+        LOG.info("Received event with id '{}' for Processor with name '{}' on Bridge '{}", cloudEvent.getId(), processor.getName(), processor.getBridgeId());
+
+        Map<String, Object> cloudEventData = CloudEventUtils.getMapper().convertValue(cloudEvent, Map.class);
+
+        // Filter evaluation
+        if (Boolean.TRUE.equals(filterTimer.record(() -> filterEvaluator.evaluateFilters(cloudEventData)))) {
+            LOG.info("Filters of processor '{}' matched for event with id '{}'", processor.getId(), cloudEvent.getId());
+
+            // Transformation
+            String eventToSend = transformationTimer.record(() -> transformationEvaluator.render(cloudEventData));
```
robpblake, line 69: @kostola For event sources, any transformation must produce another `CloudEvent`. Otherwise we potentially have events on the internal Kafka topic not in CloudEvent format.

Transformation/Filtering of Events from Sources are currently out-of-scope for your work, so we can postpone

executor/src/main/java/com/redhat/service/smartevents/executor/impl/CloudEventExecutorImpl.java
```diff
@@ -0,0 +1,109 @@
...
+import com.redhat.service.smartevents.infra.transformations.TransformationEvaluatorFactory;
+import com.redhat.service.smartevents.infra.utils.CloudEventUtils;
+import com.redhat.service.smartevents.processor.actions.ActionInvoker;
+
+import io.cloudevents.CloudEvent;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Timer;
+
+public class CloudEventExecutorImpl implements Executor {
```
robpblake, line 26: @kostola / @lucamolteni I think it would be very useful to have a label on the `pod` to help us determine the type of processor that this executor is handling, so that we can do things like:

```
kubectl get pod -l "com.redhat.smartevent/processor-type=sink"
```

robpblake (commented):
@kostola This is great work. Kudos!

It's a big addition to our functionality, so some comments from me.

The other thing to consider is that currently, filtering and transformation from an Event Source is out-of-scope for your work. However, I'd suggest we actually keep the Filtering part as we are starting to get some requirements in that area. 

We do not need to expose it via the UI until we're ready to fully integrate it, but I think it might be useful to keep that functionality cc @kelvah

processor/configuration/src/main/java/com/redhat/service/smartevents/processor/sources/slack/SlackSourceValidator.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Override
+    public ValidationResult isValid(Source source) {
+        Map<String, String> parameters = source.getParameters();
+
+        if (!parameters.containsKey(CHANNEL_PARAM) || parameters.get(CHANNEL_PARAM).isEmpty()) {
+            return ValidationResult.invalid(INVALID_CHANNEL_MESSAGE);
+        }
+
+        if (!parameters.containsKey(TOKEN_PARAM) || parameters.get(TOKEN_PARAM).isEmpty()) {
```
robpblake, line 27: Storing this value in plaintext in the DB is fine for this first iteration to get things working. However, we should drop that as soon as possible. https://issues.redhat.com/browse/MGDOBR-59 captures the requirement to do that and is scheduled for next sprint.

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
kelvah: I was going to add the same comment. A `ProcessorType` could be helpful. We display the type in different UI locations and having a dedicated `type` property is better than inferring it from the presence of source/action. Also useful for filtering in the future.

processor/configuration/src/main/java/com/redhat/service/smartevents/processor/sources/SourceConfigurator.java
kostola: This class is actually not used because I merged it in `GatewayConfigurator`, where this bean is not `Optional` anymore. I will remove this class.

processor/core/src/main/java/com/redhat/service/smartevents/processor/actions/input/InputAction.java
kostola: I honestly thought about `SourceAction` as first name but then I thought also it would have caused confusion, but honestly it's the clearest once you understand how it works. :smile:

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
kostola: This is one of the shortcuts I took for the first version of the PoC. I will implement it.

manager/src/main/java/com/redhat/service/smartevents/manager/api/user/validators/processors/GatewayConstraintValidator.java
kostola: This is one of the shortcuts I took for the first version of the PoC. I will implement it.

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
kostola: I think too :+1:

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/input/InputActionInvoker.java
kostola: This is one of the shortcuts I took for the first version of the PoC. I will implement it.

infra/src/main/java/com/redhat/service/smartevents/infra/models/actions/Gateway.java
```diff
@@ -4,22 +4,17 @@
...
-import javax.validation.constraints.NotNull;
-
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonInclude(JsonInclude.Include.NON_NULL)
-public class BaseAction {
+public class Gateway {
```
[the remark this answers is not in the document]
kostola: A gateway in my idea represents the touching point between a Processor and the external world in which the events pass through to flow in and out.
It is intended to be the common concept between `Source` (the gateway from which events get in the Processor) and the `Action` (the gateway from which events exit the Processor towards the external world).

r00ta (commented):
I know this is a WIP but I have a couple of initial questions: 
1) are we going to split this new feature into smaller tickets? 
2) I know it's a PoC, but I find a bit confusing that the client has to call the `processors` endpoint for both the `source processor` and the `sink processor` (with different payload of course). Since they are two completely different logical entities for the client, I'd use 2 different endpoints.

r00ta (commented):
@robpblake are we going to add this feature in one goal with this PR? My understanding was that this PR is a draft to show the code of the PoC.

robpblake:
> @robpblake are we going to add this feature in one goal with this PR? My understanding was that this PR is a draft to show the code of the PoC.

@r00ta That was my understanding as well: This PR is the PoC and chance to give feedback on the initial approach. After this, we'd update the PoC with feedback and then submit a set of smaller PRs so that we're not introducing a lot of change at once. We also need to ensure that we have updated coverage on our integration testing.

@kostola Does that match your plan ^

kostola:
>@kostola Does that match your plan

@robpblake honestly this was the initial plan but after seeing all these reviews popping out I assumed this was going to evolve directly into the feature itself, and that's why I started fixing the tests and now I was going to increase coverage of the new part.

I am now trying to understand how hard it would be to split it into several PRs.

robpblake:
> I know it's a PoC, but I find a bit confusing that the client has to call the processors endpoint for both the source processor and the sink processor (with different payload of course). Since they are two completely different logical entities for the client, I'd use 2 different endpoints.

@r00ta  I had the same initial reaction when reviewing this, but I've changed my mind to the single endpoint. So for clarity I'm personally +1 on:

- POST --> /api/v1/bridges/{id}/processors
- GET --> /api/v1/bridges/{id}/processors <-- gets all processors
- GET --> /api/v1/bridges/{id}/processors?type={sink|source} <-- gets all processors of a specific "type"

with the type of `Processor` defined as a `type` parameter on the response payload e.g:

```
{
  "type": "sink"
} 
```

versus

- POST --> /api/v1/bridges/{id}/sink_processor
- POST --> /api/v1/bridges/{id}/source_processor
- GET --> /api/v1/bridges/{id}/sink_processor
- GET --> /api/v1/bridges/{id}/source_processor

Couple of reasons why I changed my mind:

1. It felt simpler to deal with a single endpoint with filtering capabilities rather than the client having to understand and combine different endpoints e.g `/api/v1/bridges/{id}/processors` vs union of `/api/v1/bridges/{id}/sink_processor` and `/api/v1/bridges/{id}/source_processor` to list all my processors
2. Personal preference, but the definitions of Sink and Source processors are "similar enough" to not require a separate endpoint only difference being one contains a `source` declaration, the other an `action`

cheers,

Rob

cc @kostola

robpblake:
> @robpblake @r00ta I looked at the code and this is my analysis based on the existing tasks:
> 
> * Most of this code would fall under [MGDOBR-322](https://issues.redhat.com/browse/MGDOBR-322)
> * The changes required for [MGDOBR-323](https://issues.redhat.com/browse/MGDOBR-323) and [MGDOBR-324](https://issues.redhat.com/browse/MGDOBR-324) are **really small and deeply related to MGDOBR-322**, which means there's no sense in having separated PRs for these two tasks.
> * [MGDOBR-326](https://issues.redhat.com/browse/MGDOBR-326) is not needed anymore since there's no change in the way the operator deploys source processors with this design.
> * The only task we can move to a separated PR could be [MGDOBR-325](https://issues.redhat.com/browse/MGDOBR-325), since we need a dedicated executor for source processors that handles plain events instead of CloudEvents. However, the footprint of these changes is way less than all the rest.
> 
> Let me know if you want to move MGDOBR-325 logic to a new PR or if we should proceed with this one this time.
> 
> I think also we can have a brainstorming call to discuss the changes in depth with all the devs to optimize our time.
> 
> cc @lucamolteni

@kostola Your suggested split between manager-side changes -322,-323,-324 in one PR and then shard side changes in -325 sound reasonable to me. As we've said before, it's just about striking a balance with the amount of changes in each PR.

Rather than setting up a meeting, what I'd suggest is that you give everyone a deadline e.g COB tomorrow to give you feedback on this PR as it contains the vast majority of changes IMHO as is pretty clear. If there are anything that cannot be resolved in comments on this PR, then we can have a call. From my perspective you've done a great job here and I think there isn't too much to discuss.

We also need to make sure we have a plan to update our integration tests to cover this new functionality.

cheers,
Rob

r00ta:
> @robpblake @r00ta I looked at the code and this is my analysis based on the existing tasks:
> 
> * Most of this code would fall under [MGDOBR-322](https://issues.redhat.com/browse/MGDOBR-322)
> * The changes required for [MGDOBR-323](https://issues.redhat.com/browse/MGDOBR-323) and [MGDOBR-324](https://issues.redhat.com/browse/MGDOBR-324) are **really small and deeply related to MGDOBR-322**, which means there's no sense in having separated PRs for these two tasks.
> * [MGDOBR-326](https://issues.redhat.com/browse/MGDOBR-326) is not needed anymore since there's no change in the way the operator deploys source processors with this design.
> * The only task we can move to a separated PR could be [MGDOBR-325](https://issues.redhat.com/browse/MGDOBR-325), since we need a dedicated executor for source processors that handles plain events instead of CloudEvents. However, the footprint of these changes is way less than all the rest.
> 
> Let me know if you want to move MGDOBR-325 logic to a new PR or if we should proceed with this one this time.
> 
> I think also we can have a brainstorming call to discuss the changes in depth with all the devs to optimize our time.
> 
> cc @lucamolteni

my2c: It would be great if you manage to refactor the codebase with separated tickets (for example, if we go for the 2 enpoints `sink_processors` and `source_processors` you could migrate the current `processors` enpoint to `sink_processors` and keep the PR scoped to that specific goal).

kostola:
@r00ta I explained in detail in my previous comment why I believe there's no point in splitting more than two PRs.
Also your example for a further split, using two separate endpoints for source processors and sink processors, is **explicitly not wanted** in the Event Sources design document. See the ["How does the user create a Source Processor" paragraph](https://docs.google.com/document/d/1bbiks0BYedetZCSDxSRXzNYZsbheEd0yy69a5tc4yQY/edit#bookmark=id.j2ht5zvwnsl9) and the [rejected proposal of adding a new explicit API](https://docs.google.com/document/d/1bbiks0BYedetZCSDxSRXzNYZsbheEd0yy69a5tc4yQY/edit#bookmark=id.2rtiit1fq3ga).

@robpblake what I'm going to do now is create two separate PR, one for the "manager" side and one for the "executor" side (the shard operator is almost untouched) and then start working on them because we need to deliver this.

r00ta:
> @r00ta I explained in detail in my previous comment why I believe there's no point in splitting more than two PRs. Also your example for a further split, using two separate endpoints for source processors and sink processors, is **explicitly not wanted** in the Event Sources design document. See the ["How does the user create a Source Processor" paragraph](https://docs.google.com/document/d/1bbiks0BYedetZCSDxSRXzNYZsbheEd0yy69a5tc4yQY/edit#bookmark=id.j2ht5zvwnsl9) and the [rejected proposal of adding a new explicit API](https://docs.google.com/document/d/1bbiks0BYedetZCSDxSRXzNYZsbheEd0yy69a5tc4yQY/edit#bookmark=id.2rtiit1fq3ga).
> 
> @robpblake what I'm going to do now is create two separate PRs, one for the "manager" side and one for the "executor" side (the shard operator is almost untouched) and then start working on them because we need to deliver this.

According to @robpblake comments, we have to update that document with the specification that we do want 2 endpoints.

kostola:
@r00ta I don't want two separate endpoints

r00ta:
> @r00ta I don't want two separate endpoints for the same reasons that @robpblake explained

Sorry I misread @robpblake comment. I leave this up to you then

processor/configuration/src/main/java/com/redhat/service/smartevents/processor/sources/slack/SlackSourceResolver.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Override
+    public Action resolve(Source source, String customerId, String bridgeId, String processorId) {
+        Action resolvedAction = new Action();
+        resolvedAction.setType(InputAction.TYPE);
+
+        try {
+            resolvedAction.setParameters(Map.of(
+                    InputAction.ENDPOINT_PARAM, getBridgeWebhookUrl(customerId, bridgeId),
+                    InputAction.CLOUD_EVENT_TYPE, "SlackSource"));
```
jomarko, line 31: Reading the code, it is sometimes hard to understand when type is `Slack` and when `SlackSource`

processor/core/src/main/java/com/redhat/service/smartevents/processor/GatewayBean.java
```diff
@@ -0,0 +1,11 @@
+package com.redhat.service.smartevents.processor;
+
+import com.redhat.service.smartevents.infra.models.gateways.Gateway;
+
+public interface GatewayBean<T extends Gateway> {
+    String getType();
+
+    default boolean accept(String gatewayType) {
+        return getType().equals(gatewayType);
```
jomarko, line 9: Could we please use `Objects.equals(a, b)` pattern instead of `a.equals(b)`. The first is NullPointerException safe.

processor/core/src/main/java/com/redhat/service/smartevents/processor/actions/input/InputAction.java
```diff
@@ -0,0 +1,16 @@
+package com.redhat.service.smartevents.processor.actions.input;
+
+import com.redhat.service.smartevents.infra.models.gateways.Action;
+import com.redhat.service.smartevents.processor.GatewayBean;
+
+public interface InputAction extends GatewayBean<Action> {
```
jomarko, line 6: Is this some 5th action next to slack, webhook, sendtobridge and kafka?

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/input/InputActionInvoker.java
```diff
@@ -0,0 +1,85 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(InputActionInvoker.class);
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    private final String endpoint;
+    private final String cloudEventType;
+    private final WebClient client;
+    private final AbstractOidcClient oidcClient;
+
+    public InputActionInvoker(String endpoint, String cloudEventType, WebClient client, AbstractOidcClient oidcClient) {
```
jomarko, line 33: Shouldn't we have some enum for `cloudEventType`? maybe in some future ticket?

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/input/InputActionInvoker.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+
+    private static String encode(ObjectNode event) {
+        try {
+            return MAPPER.writeValueAsString(event);
+        } catch (JsonProcessingException e) {
+            throw new CloudEventSerializationException("Failed to encode CloudEvent");
+        }
+    }
```
jomarko, line 84: Shouldn't we use existing `CloudEventUtils`?

jomarko (commented):
Thank you for this contribution @kostola . I have some comments/questions to code inline.

Also general comment about tests. I think we updated existing one with 'Sink', but not sure we have for 'Source'?

Do we have updates for README files?

Not sure I can try this manually? You mentioned needed some slack bot set-up in the demo video?

r00ta:
@kostola can we close this?

kostola:
Closed because the feature was developed in PRs derived from this PoC and they are all merged now :+1:

PR #641, opened 2022-04-19
MGDOBR-598: Failed to Update the same as existing Filter definition to the Processor

manstis:
[https://issues.redhat.com/browse/MGDOBR-598](https://issues.redhat.com/browse/MGDOBR-598)

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.

</details>

### DIFF
--- a/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeIngressServiceImpl.java
+++ b/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeIngressServiceImpl.java
@@ -77,6 +77,12 @@ public class BridgeIngressServiceImpl implements BridgeIngressService {
 
             // create or update the secrets for the bridgeIngress
             createOrUpdateBridgeIngressSecret(bridgeIngress, bridgeDTO);
+        } else {
+            LOGGER.info("Existing BridgeIngress '{}' matched the expected specification. No update required. Notifying manager that is it ready.", bridgeDTO.getId());
+            bridgeDTO.setStatus(ManagedResourceStatus.READY);
+            managerClient.notifyBridgeStatusChange(bridgeDTO).subscribe().with(
+                    success -> LOGGER.debug("Ready notification for BridgeIngress '{}' has been sent to the manager successfully", bridgeDTO.getId()),
+                    failure -> LOGGER.error("Failed to send updated status to Manager for entity of type '{}'", BridgeDTO.class.getSimpleName(), failure));
         }
     }
 
--- a/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeIngressServiceImpl.java
+++ b/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeIngressServiceImpl.java
@@ -77,12 +77,6 @@ public class BridgeIngressServiceImpl implements BridgeIngressService {
 
             // create or update the secrets for the bridgeIngress
             createOrUpdateBridgeIngressSecret(bridgeIngress, bridgeDTO);
-        } else {
-            LOGGER.info("Existing BridgeIngress '{}' matched the expected specification. No update required. Notifying manager that is it ready.", bridgeDTO.getId());
-            bridgeDTO.setStatus(ManagedResourceStatus.READY);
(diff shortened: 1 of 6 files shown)


robpblake:
@manstis I'll ask what might be an incredibly naive question: would it be simpler to check in the Fleet Manager at the point that the user calls the API for the update if there are actually changes required? At the minute we're doing a full invocation of our deployment pipeline to simply do a null-op.

cheers,

Rob

manstis:
Hey @robpblake it's not a naive question.

I prefer keeping the life-cycle code consistent for all CRUD operations; but in this case it is an unnecessary round-trip so I've added the check you propose to the `manager`. I've kept the `operator` check in this PR too as it ensures integrity of the `Processor`'s `status` for when we handle `Processor`'s `action` updates. In this scenario the `Processor` may need to move to `ACCEPTED` state to (re-)provision a Managed Connector by the `Worker` mechanism but the Kubernetes CRD spec would be unchanged... I'm guessing this could be a valid scenario as the mechanism to perform `action` updates is undetermined at the moment.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeIngressServiceImpl.java
```diff
@@ -77,6 +77,12 @@ public void createBridgeIngress(BridgeDTO bridgeDTO) {
 
             // create or update the secrets for the bridgeIngress
             createOrUpdateBridgeIngressSecret(bridgeIngress, bridgeDTO);
+        } else {
+            LOGGER.info("Existing BridgeIngress '{}' matched the expected specification. No update required. Notifying manager that is it ready.", bridgeDTO.getId());
+            bridgeDTO.setStatus(ManagedResourceStatus.READY);
+            managerClient.notifyBridgeStatusChange(bridgeDTO).subscribe().with(
+                    success -> LOGGER.debug("Ready notification for BridgeIngress '{}' has been sent to the manager successfully", bridgeDTO.getId()),
+                    failure -> LOGGER.error("Failed to send updated status to Manager for entity of type '{}'", BridgeDTO.class.getSimpleName(), failure));
```
r00ta, line 85: well this is not always true. 
Consider the following scenario: 
1) Something goes wrong and the resource is in a FAILED state. 
2) This method is called and there is no change in the CRD
3) -> with the current implementation you update the manager with the READY state, while the resource is in a FAILED state. 

In short, I think you should return the status you find in the `existing` resource.

r00ta (changes requested):

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeIngressServiceImpl.java
manstis: Hi @r00ta the best I can do in the `operator` (without significant refactoring) is `processorDTO.setStatus(existing.getStatus().isReady() ? ManagedResourceStatus.READY : ManagedResourceStatus.FAILED)`... i.e. if the CRD was "ready" I signal the `manager` that the `ManagedResource` is `READY` otherwise it must have `FAILED`. The CRD controller only sets the `ManagedResourceStatus` to either `READY` or `FAILED` so it seems reasonable. WDYT?

manstis: @r00ta given @robpblake [comment](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/641#issuecomment-1102861125) IDK whether I should remove the check from `manager` as given your insightful comment if a CRD was `FAILED` and I do the check in the `manager` (as Rob proposed and I've implemented) we'll never get the operator to try and re-provision the updated definition... unless I also check if the state is `READY`.. hmm.. let me add that as a commit until we've finalised our discussion.

manstis:
@robpblake Following discussion with @r00ta we decided to keep the check only in the `manager`.

This PR has now been updated accordingly. IDK if you want to re-review it.

r00ta (approved):

Srihari1192 (approved):

lucamolteni (approved):

PR #646, opened 2022-04-20
MGDOBR-322: introduce Gateway

kostola:
[MGDOBR-322](https://issues.redhat.com/browse/MGDOBR-322)

## Description

This PR belongs to the [Event Sources epic](https://issues.redhat.com/browse/MGDOBR-270) and is part of a sequence of PRs derived from the Event Sources PoC (#629)

## PR owner checklist

Please make sure that your PR meets the following requirements:

- [ ] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/main/kogito-build/kogito-ide-config).  
  *Run `mvn clean verify -DskipTests` so that imports are correctly set.*
- [ ] Your commit messages are clear and reference the JIRA issue e.g: "[MGDOBR-1] - $clear_explanation_of_what_you_did"
- [ ] All new functionality is tested
- [ ] Pull Request title is properly formatted: `MGDOBR-XYZ Subject`
- [ ] Pull Request contains link to the JIRA issue
- [ ] Pull Request contains link to any dependent or related Pull Request
- [ ] Pull Request contains description of the issue
- [ ] Pull Request does not include fixes for issues other than the main ticket

## PR reviewer(s) checklist

- [ ] Check if code and Github action workflows have been modified and if the modification is safe
- [ ] If the modification is safe, add the `safe to test` label

<details>
<summary>
How to trigger pipelines and use the bots:
</summary>

- <b>Run the end to end pipeline</b>  
  Annotate the pull request with the label: `safe to test`. If you want to run the pipeline again, remove and add it again.

- <b>Rebase the pull request</b>  
  Comment with: `/rebase`.

</details>

PR #651, opened 2022-04-21
MGDOBR-322: refactor ProcessorServiceTest

kostola:

**Blocked by #646**

## Description

### Why this refactoring?

The existing `ProcessorServiceTest` was quite complex and was in fact performing two completely separated kind of tests:

- Test `ProcessorService` logic (as one would expect from the test name).
- Test the interaction between `ProcessorService` and Managed Connectors client APIs (which involves `ProcessorService` but also `ConnectorService` and `ConnectorWorker`).

This refactor splits these two sets of tests in two separate classes: `ProcessorServiceTest` and `ProcessorServiceConnectorTest`. This makes it easier to extend the test logic once _Source Processors_ will be added (in a separate PR).

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.

</details>

### DIFF
--- a/manager/src/test/java/com/redhat/service/smartevents/manager/ProcessorServiceConnectorTest.java
+++ b/manager/src/test/java/com/redhat/service/smartevents/manager/ProcessorServiceConnectorTest.java
@@ -1,0 +1,328 @@
+package com.redhat.service.smartevents.manager;
+
+import java.util.List;
+import java.util.Map;
+
+import javax.inject.Inject;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+
+import com.openshift.cloud.api.connector.models.Connector;
+import com.openshift.cloud.api.connector.models.ConnectorRequest;
+import com.openshift.cloud.api.connector.models.ConnectorState;
+import com.openshift.cloud.api.connector.models.ConnectorStatusStatus;
+import com.openshift.cloud.api.kas.auth.models.Topic;
+import com.redhat.service.smartevents.infra.exceptions.definitions.platform.InternalPlatformException;
+import com.redhat.service.smartevents.infra.models.dto.ManagedResourceStatus;
+import com.redhat.service.smartevents.infra.models.gateways.Action;
+import com.redhat.service.smartevents.manager.api.models.requests.ProcessorRequest;
+import com.redhat.service.smartevents.manager.connectors.ConnectorsApiClient;
+import com.redhat.service.smartevents.manager.dao.BridgeDAO;
+import com.redhat.service.smartevents.manager.dao.ConnectorsDAO;
+import com.redhat.service.smartevents.manager.dao.ProcessorDAO;
+import com.redhat.service.smartevents.manager.models.Bridge;
+import com.redhat.service.smartevents.manager.models.ConnectorEntity;
+import com.redhat.service.smartevents.manager.models.Processor;
+import com.redhat.service.smartevents.manager.providers.ResourceNamesProvider;
+import com.redhat.service.smartevents.manager.utils.DatabaseManagerUtils;
+import com.redhat.service.smartevents.manager.utils.Fixtures;
+import com.redhat.service.smartevents.processor.actions.slack.SlackAction;
+import com.redhat.service.smartevents.rhoas.RhoasTopicAccessType;
+import com.redhat.service.smartevents.test.resource.PostgresResource;
+
+import io.quarkus.test.common.QuarkusTestResource;
+import io.quarkus.test.junit.QuarkusTest;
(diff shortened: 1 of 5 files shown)


manager/src/test/java/com/redhat/service/smartevents/manager/ProcessorServiceConnectorTest.java
```diff
@@ -0,0 +1,328 @@
...
+
+    @Test
+    void createConnectorFailureOnExternalConnectorCreation() {
+        Bridge b = createPersistBridge(READY);
+
+        Action slackAction = createSlackAction();
+        ProcessorRequest processorRequest = new ProcessorRequest("ManagedConnectorProcessor", slackAction);
+
+        doThrow(new InternalPlatformException(RhoasServiceImpl.createFailureErrorMessageFor("errorDeletingConnector"), new RuntimeException("error")))
+                .when(connectorsApiClient).deleteConnector(anyString());
```
manstis, line 158: Why are you mocking behaviour for `deleteConnector(..)` in a `createConnectorFailure...` test?

manager/src/test/java/com/redhat/service/smartevents/manager/ProcessorServiceTest.java
```diff
@@ -9,533 +8,435 @@
...
+    void getProcessors_bridgeDoesNotExist() {
+        assertThatExceptionOfType(ItemNotFoundException.class)
+                .isThrownBy(() -> processorService.getProcessors(NON_EXISTING_BRIDGE_ID, DEFAULT_CUSTOMER_ID, QUERY_INFO));
     }
 
     @Test
-    public void deleteProcessor_processorDoesNotExist() {
-        Bridge b = createPersistBridge(ManagedResourceStatus.READY);
-        assertThatExceptionOfType(ItemNotFoundException.class).isThrownBy(() -> processorService.deleteProcessor(b.getId(), "doesNotExist", b.getCustomerId()));
+    void getProcessorsCount() {
```
manstis, line 328: It's a bit weird having a `getXXX()` method returning `void` that does an assertion.

manager/src/test/java/com/redhat/service/smartevents/manager/ProcessorServiceConnectorTest.java
manstis: I investigated and see this is just a copy and paste, so no blame on your part.

I think the test should mock this behaviour instead:
```
        doThrow(new InternalPlatformException(RhoasServiceImpl.createFailureErrorMessageFor("errorDeletingConnector"), new RuntimeException("error")))
                .when(connectorsApiClient).createConnector(any(ConnectorEntity.class));
```
I tried the modified test here and 
- The test still works as expected (i.e. passes)
- It's actually testing the right bit of code (based on the name of the test).

manager/src/test/java/com/redhat/service/smartevents/manager/ProcessorServiceTest.java
kostola: Yes I agree. I prefixed all the test methods with `"test"` to avoid confusion.

manager/src/test/java/com/redhat/service/smartevents/manager/ProcessorServiceConnectorTest.java
kostola: I think in this refactoring we should focus on the split, then since you wrote most of that code and you feel it can be improved, you can do it in a follow-up PR. Wdyt @manstis ?

manstis: Well, for the sake of a 1 line change I think it'd be OK adding it to this PR.

There are other changes you've made to the code-base other than splitting for this PR too.

But if you're adamant it cannot be accommodated I'll do later.

danielezonca (approved):
This PR is mainly a refactoring so I'm ok to proceed and merge: it has a better modularization so it is an improvement.
My feeling is that we should revisit similar test classes: try to isolate entirely each class with mocks and test each method that has "no logic" (i.e. [`ProcessorService.getProcessorsCount`](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/eef7827cefabd44188015d863c247d355c2e5784/manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java#L223-L227)) is quite time consuming to maintain it is not probably adding "real" coverage.
I think we can decide to skip unit testing methods with a single flow (aka no branches) if there is at least another test (end to end if possible) that is invoking this codepath.

Btw let's move the discussion somewhere else (i.e. sprint retrospective), fine for me to merge :)

sutaakar (approved):

PR #662, opened 2022-04-23
MGDOBR-322: add "type" and "source" to Processor entity, definition and DTOs

kostola:

## Description

This PR does two things:
- Extends the `"source"` field, introduced to `ProcessorRequest` with #654, to `ProcessorDefinition`, `ProcessorDTO`, `ProcessorResponse` and `BridgeExecutorSpec` so that the information can be correctly propagated around in all the system.
- Introduce a new `ProcessorType` enum (`SOURCE` or `SINK`) and a `"type"` field to `Processor` entity (with a dedicated column in the corresponding table) and all the other DTOs that require it.

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.

</details>


infra/src/main/java/com/redhat/service/smartevents/infra/models/processors/ProcessorDefinition.java
```diff
@@ -18,6 +19,9 @@ public class ProcessorDefinition {
     @JsonProperty("requestedAction")
     private Action requestedAction;
 
+    @JsonProperty("requestedSource")
```
manstis, line 22: I'll have a proper look next week but biggest gripe is that this change really suggests we need a base `ProcessorDefinition` abstract class with two sub-classes; one for Sinks (with the Action field) and the other for Sources (with the Source field). That'd obviously have impact elsewhere.

sutaakar:
@kostola  Please adjust operator test with the missing processor type in https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/shard-operator-integration-tests/src/test/resources/features/bridgeexecutor-deploy.feature

kostola:
> Please adjust operator test with the missing processor type

@sutaakar done :+1:

infra/src/main/java/com/redhat/service/smartevents/infra/models/processors/ProcessorDefinition.java
kostola: In theory this makes sense, but practically it would require quite some changes in the code that are not worth at this point. We can definitely think about this if the `ProcessorDefinition` becomes more complex.

sutaakar (approved):

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/resources/BridgeExecutor.java
```diff
@@ -155,6 +164,12 @@ public BridgeExecutor build() {
             bridgeExecutorSpec.setCustomerId(customerId);
             bridgeExecutorSpec.setProcessorName(processorName);
 
+            if (processorType != null) {
```
robpblake, line 167: @kostola Why this `!= null` check only on `processorType`? All the other fields are required e.g `customerId` so I think we should be consistent in our approach here.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/resources/BridgeExecutorSpec.java
```diff
@@ -11,6 +11,8 @@ public class BridgeExecutorSpec {
 
     private String customerId;
 
+    private String processorType;
```
robpblake, line 14: @kostola Doesn't this require our CRDs to be regenerated with the additional `processorType` field for the spec?

robpblake (changes requested):
@kostola Small change is required to the DB schema please to remove defaulting to `SINK` on the Processor table. I've also provided some other minor feedback

One other question: How are we handling CRD update for already running Processors?

infra/src/main/java/com/redhat/service/smartevents/infra/models/processors/ProcessorDefinition.java
jomarko: @kostola from @manstis review message, I would say, we need a jira to track this

lucamolteni (approved):

infra/src/main/java/com/redhat/service/smartevents/infra/models/filters/BaseFilter.java
```diff
@@ -49,5 +49,5 @@ public String getType() {
 
     @NotNull
     @JsonIgnore
-    public abstract T getValue();
+    public abstract Object getValue();
```
manstis, line 52: I am trying to understand why you made this change?

It seems to move filters further away from the "typed" form I thought we were moving towards.

Was it just to simplify the generics?

infra/src/main/java/com/redhat/service/smartevents/infra/models/processors/ProcessorDefinition.java
robpblake: @manstis No I see your point, and happy to take a follow up issue into the backlog, but at this point I'm inclined to agree with @kostola in terms of effort vs reward of doing the work. 

In particular if we have dedicated constructors for the two ways in which a `ProcessorDefinition` can be created e.g one that requires `Source` and resolved `Action` and the other than requires requested `Action` and resolved `Action` then I think that helps to ensure we don't create the objects incorrectly.

cheers,

Rob

manager/src/main/java/com/redhat/service/smartevents/manager/models/Processor.java
```diff
@@ -51,6 +55,10 @@ public class Processor extends ManagedDefinedResource {
 
     public static final String BRIDGE_ID_PARAM = "bridgeId";
 
+    @Column(name = "type")
```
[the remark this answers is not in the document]
kostola: Done :+1:

manager/src/main/resources/db/migration/0.1/V0.1.9__update_processor.sql
```diff
@@ -0,0 +1,2 @@
+alter table PROCESSOR
+    add column "type" varchar(255) default 'SINK';
```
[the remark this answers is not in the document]
kostola: I updated the migration. Please take a look and let me know if it works :+1:

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/resources/BridgeExecutor.java
kostola: Done :+1:

infra/src/main/java/com/redhat/service/smartevents/infra/models/filters/BaseFilter.java
kostola: The generic type in `BaseFilter` was used only for the return type of the `getValue` method, but Java allows to override a method specifying a stricter return type than the superclass without the need of generics (i.e. return type in the subclass is a subclass of the return type in the superclass).

Thus, using plain `Object` in `BaseFilter` achieves the same goal and also eliminates all the warnings that SonarCloud was throwing because `BaseFilter` was used *everywhere* in the codebase except in one place *without the generic type specified* (i.e. `BaseFilter` instead of `BaseFilter<...>`). SonarCloud warnings are the main reason why I did this.

kostola:
> It seems to me we miss a test for ProcessorType.SOURCE. Or am I missing sth?

@jomarko this PR only adds fields to entity, definition and DTOs, but the business logic to actually create source processor will be added in a following PR (#664). I already added all the tests to cover source processors there.

infra/src/main/java/com/redhat/service/smartevents/infra/models/processors/ProcessorDefinition.java
kostola: @robpblake @manstis I created a follow-up task for this: [MGDOBR-635](https://issues.redhat.com/browse/MGDOBR-635)

manstis (approved):

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsServiceImpl.java
```diff
@@ -51,7 +51,6 @@ public void createConnectorEntity(Processor processor, Action action) {
         persistConnectorEntity(processor, topicName, actionConnector.getConnectorType(), actionConnector.connectorPayload(action, topicName));
     }
 
-    @Transactional(Transactional.TxType.MANDATORY)
```
robpblake, line 54: @kostola Was there a reason why this is required?

robpblake, line 54: @kostola Was there a reason why this is required?

manager/src/main/resources/db/migration/0.1/V0.1.9__update_processor.sql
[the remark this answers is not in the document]
robpblake: Thanks @kostola . Looks great.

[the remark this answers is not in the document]
robpblake: Thanks @kostola . Looks great.

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsServiceImpl.java
kostola: Well I got this warning message when starting the manager locally:

```
@Transactional(value = MANDATORY) will have no effect on method com.redhat.service.smartevents.manager.connectors.ConnectorsServiceImpl.persistConnectorEntity() because the method is private
```

so I removed it

kostola: Replied [here](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/662#discussion_r859647623)

manstis: @robpblake The `private` method is invoked from a `public` method [with](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/a67296e1e6ee9711a85d82b1e0747c3fb95d7717/manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsServiceImpl.java#L42) the annotation; so I think we're OK.

@kostola a `private` method cannot be proxied by the codegen for the annotation.. so it is disregarded. That said I think we're OK with your change, as I mention above.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/resources/BridgeExecutorSpec.java
r00ta: The new CRDs are generated in the `target` folder when you compile the application. When you merge the new code the bot will include the CRDs update in the `update kustomization images` pull request

jomarko (approved):

PR #663, opened 2022-04-23
MGDOBR-322: add SlackSource and InputAction

kostola:
[MGDOBR-322](https://issues.redhat.com/browse/MGDOBR-322)

## Description

This PR belongs to the [Event Sources epic](https://issues.redhat.com/browse/MGDOBR-270) and is part of a sequence of PRs derived from the Event Sources PoC (#629)

## PR owner checklist

Please make sure that your PR meets the following requirements:

- [ ] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/main/kogito-build/kogito-ide-config).  
  *Run `mvn clean verify -DskipTests` so that imports are correctly set.*
- [ ] Your commit messages are clear and reference the JIRA issue e.g: "[MGDOBR-1] - $clear_explanation_of_what_you_did"
- [ ] All new functionality is tested
- [ ] Pull Request title is properly formatted: `MGDOBR-XYZ Subject`
- [ ] Pull Request contains link to the JIRA issue
- [ ] Pull Request contains link to any dependent or related Pull Request
- [ ] Pull Request contains description of the issue
- [ ] Pull Request does not include fixes for issues other than the main ticket

## PR reviewer(s) checklist

- [ ] Check if code and Github action workflows have been modified and if the modification is safe
- [ ] If the modification is safe, add the `safe to test` label

<details>
<summary>
How to trigger pipelines and use the bots:
</summary>

- <b>Run the end to end pipeline</b>  
  Annotate the pull request with the label: `safe to test`. If you want to run the pipeline again, remove and add it again.

- <b>Rebase the pull request</b>  
  Comment with: `/rebase`.

</details>

PR #664, opened 2022-04-24
MGDOBR-270: Event Sources business logic

kostola:
[MGDOBR-270](https://issues.redhat.com/browse/MGDOBR-270)

**Blocked by:**
- #662 
- #663 

## Description

This PR is the *final one* of the sequence. It contains the business logic that will make the first version of the *Event Sources* feature actually usable.

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.

</details>


executor/src/main/java/com/redhat/service/smartevents/executor/ExecutorsService.java
```diff
@@ -30,10 +26,9 @@ public class ExecutorsService {
     @Incoming(EVENTS_IN_CHANNEL)
     public CompletionStage<Void> processBridgeEvent(final Message<String> message) {
         try {
-            CloudEvent cloudEvent = CloudEventUtils.decode(message.getPayload());
             Executor executor = executorsProvider.getExecutor();
             try {
-                executor.onEvent(cloudEvent);
+                executor.onEvent(message.getPayload());
```
danielezonca, line 31: We should move in the opposite direction: I would add the CloudEvents parsing here and preserve `executor.onEvent(CloudEvents)` method.
The whole system is designed to manage CloudEvents and in the future RHOC is going to do the conversion for us so at that point we will probably going to have only CloudEvents everywhere.

My suggestion is to check if this a valid CloudEvents, if not, "convert" it as CloudEvents (adding missing metadata) here

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/ConnectorWorker.java
```diff
@@ -193,4 +193,14 @@ protected ConnectorEntity doDeleteDependencies(ConnectorEntity connectorEntity)
         return connectorEntity;
     }
 
+    private static boolean isSourceConnector(ConnectorEntity connectorEntity) {
+        return connectorEntity.getConnectorType().contains("_source_");
```
danielezonca, line 197: Can we use a proper `type` check instead of a contains?

manager/src/test/java/com/redhat/service/smartevents/manager/connectors/ConnectorsServiceTest.java
```diff
@@ -33,77 +44,85 @@ class ConnectorsServiceTest {
...
+        }
+
         processor.setId(TEST_PROCESSOR_ID);
         processor.setName(TEST_PROCESSOR_NAME);
+        processor.setDefinition(processorDefinition);
         return processor;
     }
 
-    private Action testSlackAction() {
+    private static Action testSlackAction() {
```
danielezonca, line 125: Minor note (and it is already on main so not introduced by this PR): I find confusing having a method named `test` if it is not a test :)

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/source/SourceActionInvokerBuilder.java
```diff
@@ -0,0 +1,29 @@
...
+        String endpoint = getOrThrowException(action, ENDPOINT_PARAM);
+        String cloudEventType = getOrThrowException(action, CLOUD_EVENT_TYPE_PARAM);
+        AbstractOidcClient abstractOidcClient = getOidcClient();
+        return new SourceActionInvoker(endpoint, cloudEventType, webClient, abstractOidcClient);
+    }
+
+    private static String getOrThrowException(Action action, String parameterName) {
+        if (action.getParameters() == null || !action.getParameters().containsKey(parameterName)) {
+            throw new GatewayProviderException(String.format("Missing %s parameter from action", parameterName));
+        }
```
danielezonca, line 26: Is this validation also performed during Action validation?

danielezonca (commented):
Not a full review, just some comments :)

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/source/SourceActionInvoker.java
```diff
@@ -0,0 +1,90 @@
...
+        this.endpoint = endpoint;
+        this.cloudEventType = cloudEventType;
+        this.client = client;
+        this.oidcClient = oidcClient;
+        this.idGenerator = idGenerator;
+    }
+
+    @Override
+    public void onEvent(String event) {
+        LOG.info("SourceActionInvoker::onEvent | {} - {} | {}", endpoint, cloudEventType, event);
```
robpblake, line 51: @kostola Lots of logging in here that I think is a bit noisy and is also printing senstive data into the logs i.e the token. Can we tidy it up a bit please.

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
```diff
@@ -302,16 +317,4 @@ public ProcessorResponse toResponse(Processor processor) {
 
         return processorResponse;
     }
-
-    JsonNode definitionToJsonNode(ProcessorDefinition definition) {
```
robpblake, line 306: @kostola Does the change in serialization/deserialization impact existing entities in the database? I.e when we deploy this to `dev`, will it fail to deserialize existing entities?

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/source/SourceActionInvoker.java
```diff
@@ -0,0 +1,90 @@
...
+            request = request.bearerTokenAuthentication(token);
+        }
+        request.sendJsonObject(getPayload(event)).subscribe().with(
+                response -> LOG.info("RESPONSE {} {}", response.statusCode(), response.statusMessage()),
+                error -> LOG.error("ERROR", error));
+    }
+
+    private JsonObject getPayload(String event) {
+        try {
+            ObjectNode cloudEvent = MAPPER.createObjectNode();
```
robpblake, line 67: @kostola Do you have an example of the Event put onto the Kafka Topic by the Slack Source Connector?

I'm interested to understand if the Event contains headers, because at the minute we're putting the Kafka Event headers into the `data` part of the `CloudEvent`. So from a user experience perspective, if I want to match on a header from a Source, I'm interested to understand if they have to write a filter that targets the `data` element of the `CloudEvent`.

e.g. if the Kafka Event produced looks like this:

```
{
 "header_1":"foo"
}
```
Do I as a user have to write the following filter to match on it:

```
{
  "filters": [
    {
      "type": "StringEquals", 
      "key": "data.header_1",
      "value": "foo"
    }
  ]
}
```

Does that match user expectations or would you as a user expect the headers to be hoisted as custom attributes on the `CloudEvent`? 🤔

executor/src/main/java/com/redhat/service/smartevents/executor/Executor.java
```diff
@@ -50,31 +54,41 @@ public Executor(ProcessorDTO processor, FilterEvaluatorFactory filterEvaluatorFa
...
             // DO NOTHING;
         }
     }
 
+    @SuppressWarnings("unchecked")
+    private Map<String, Object> toEventMap(String event) {
+        try {
+            return mapper.readValue(event, Map.class);
+        } catch (JsonProcessingException e) {
+            LOG.error("JsonProcessingException when generating event map for '{}'", event, e);
```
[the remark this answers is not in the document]
robpblake: @jomarko Yes it's correct. If a `Throwable` is the last parameter of a log statement, SLF4j will print a stacktrace.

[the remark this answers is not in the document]
robpblake: @jomarko Yes it's correct. If a `Throwable` is the last parameter of a log statement, SLF4j will print a stacktrace.

manager/src/test/java/com/redhat/service/smartevents/manager/connectors/ConnectorsServiceTest.java
kostola: I renamed those methods to avoid confusion :+1:

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/source/SourceActionInvokerBuilder.java
kostola: There's no action validation phase for `SourceAction` because it's an internal action used only by source processors. However, the existence of those parameters should be guaranteed by the unit tests of the source resolvers.

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/source/SourceActionInvoker.java
kostola: Cleaned this part :+1:

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/source/SourceActionInvoker.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Override
+    public void onEvent(String event) {
+        LOG.info("SourceActionInvoker::onEvent | {} - {} | {}", endpoint, cloudEventType, event);
+
+        HttpRequest<Buffer> request = client.postAbs(endpoint);
+        String token = oidcClient.getToken();
+        LOG.info("SourceActionInvoker::onEvent | Token: {}", token);
+        if (token != null && !"".equals(token)) {
```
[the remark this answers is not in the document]
kostola: Done :+1:

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
kostola: Nope it should work because the column is always `jsonb` and the payload is compatible with the `ProcessorDefinition` bean

robpblake: Thanks!

manstis:
@kostola I will look at this tomorrow.

manstis:
@kostola I notice there are some significant comments from @robpblake and @danielezonca.

Therefore rather than add more opinion to those already made; I'll defer my review until later (if needed).

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/ConnectorWorker.java
kostola: Done :+1:

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/source/SourceActionInvokerBuilder.java
kostola: With the newest refactoring `SourceAction` is not needed anymore and standard `WebhookAction` can be used instead. Thus, I removed all this code.

executor/src/main/java/com/redhat/service/smartevents/executor/ExecutorImpl.java
```diff
@@ -0,0 +1,121 @@
...
+        LOG.debug("Received event with id '{}' and type '{}' in processor with name '{}' of bridge '{}", event.getId(), event.getType(), processor.getName(), processor.getBridgeId());
+
+        // Filter evaluation
+        if (!matchesFilters(eventMap)) {
+            LOG.debug("Filters of processor '{}' did not match for event with id '{}' and type '{}'", processor.getId(), event.getId(), event.getType());
+            return;
+        }
+        LOG.info("Filters of processor '{}' matched for event with id '{}' and type '{}'", processor.getId(), event.getId(), event.getType());
+        // Transformation
+        String eventToSend = applyTransformations(eventMap);
```
robpblake, line 78: @kostola For a `Source` processor, how do we guarantee that this transforms into a `CloudEvent`? From what I can see we simply send whatever the customer transforms into a JSON payload and send it to the Ingress via Webhook Action. Am I correct?

I think we have two options here:

- If the Processor is of type `Source`, then always marshall the transformed event back into a CloudEvent format
- Prevent the user from applying a transformation to a `Source` processor. As discussed before, that is out-of-scope for this work.

cheers,

Rob

executor/src/main/java/com/redhat/service/smartevents/executor/ExecutorService.java
```diff
@@ -0,0 +1,77 @@
...
+    @Inject
+    Executor executor;
+
+    @Inject
+    ObjectMapper mapper;
+
+    private Function<String, CloudEvent> toCloudEvent;
+
+    @PostConstruct
+    void init() {
```
robpblake, line 47: @kostola Any reason not to just in-line this in the `processEvent` method? Avoids anyone having to remember to call the `init` method. 

Not a blocking comment, just an observation.

kostola: I inlined it in `processEvent` to simplify the code as you suggested

executor/src/main/java/com/redhat/service/smartevents/executor/ExecutorImpl.java
kostola: There's no guarantee that the transformed event will always be a `CloudEvent` with the current implementation.

I refactored the code to include your second suggestion, which is also what I believe is the best solution at the moment:

> Prevent the user from applying a transformation to a Source processor

This allows us to finalize the first version of Event Sources quickly and think more in depth about transformations later. I honestly think source processors feature can initially be merged like this without compromising its usage.

executor/src/main/java/com/redhat/service/smartevents/executor/ExecutorsService.java
kostola: I refactored everything to use `CloudEvent`s both in source and sink processors as you suggested.
This had two additional major benefits:
* Filtering is now consistent between both type of processors, as they always deal with a `CloudEvent` structure.
* The special `SourceAction` internal action is not needed anymore and standard `Webhook` action can be used to resolve `Source`s. Thus, I deleted it completely.

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/source/SourceActionInvoker.java
kostola: With the refactoring [described here](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/664#discussion_r862846024) now the event received from a source Managed Connector is wrapped in a `CloudEvent` before reaching the action invoker, so the filtering is consistent betweek source and sink processors.

I honestly have no idea about Kafka Event headers. We are totally not dealing with them now.

Imho this should be discussed separately.

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
```diff
@@ -156,11 +167,16 @@ public Processor updateProcessor(String bridgeId, String processorId, String cus
         if (!Objects.equals(existingAction, processorRequest.getAction())) {
             throw new BadRequestException("It is not possible to update the Processor's Action.");
         }
+        if (!Objects.equals(existingSource, processorRequest.getSource())) {
+            throw new BadRequestException("It is not possible to update the Processor's Source.");
+        }
```
robpblake, line 173: @kostola Can you please add a check in here that prevents a user from trying to update the `type` of a `Processor` between `sink` and `source`.

cheers,

Rob

dev/SourceProcessors.md
```diff
@@ -0,0 +1,153 @@
...
+    {
+      "key": "data.text",
+      "type": "StringContains",
+      "values": [ "hello", "hi" ]
+    }
+  ]
+}
+```
+
+## Currently implemented sources
```
robpblake, line 37: @kostola We have `ACTIONS.md` in the root of the repo for the list of `Actions` we support. I think we should have something similar for the list of `Sources` that we support. Our other general user-facing docs e.g `ACTIONS.md`, `FILTERS.md` and `TRANSFORMATIONS.md` will also need updating to reflect the introduction of `Source Processors`. 

Those `.md` files are useful for the docs team who are doing their initial first iteration of customer-facing content now.

cheers,

Rob

manager/src/main/java/com/redhat/service/smartevents/manager/api/models/requests/ProcessorRequest.java
```diff
@@ -42,13 +44,29 @@ public ProcessorRequest(String name, Action action) {
...
 
+    @JsonIgnore
+    public ProcessorType getType() {
+        if (getSource() != null) {
+            return ProcessorType.SOURCE;
+        }
+        if (getAction() != null) {
+            return ProcessorType.SINK;
+        }
+        return null;
```
sutaakar, line 67: Wouldn't it have sense to throw  an exception in here (fail fast)?
If the processor type cannot be resolved than IMHO something is wrong and we shouldn't continue.

sutaakar (commented):
One minor comment fro me above, otherwise looks good to me.
Checked just the code, will leave the complex testing on Jozef.

executor/src/main/java/com/redhat/service/smartevents/executor/ExecutorImpl.java
robpblake: @kostola I agree with you. I've created https://issues.redhat.com/browse/MGDOBR-682 to capture the need for `Transformation` on an Event Source.

cheers,
Rob

lucamolteni (approved):

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/source/SourceActionInvoker.java
robpblake: @kostola @lucamolteni I'm pretty sure this will impact on the S3 Source Connector (and maybe other connectors). From what I can see of the payload shared by @lucamolteni last week, the S3 Connector is using Kafka headers to store some of the data about the event emitted from S3. So we need to understand how we're going to bind the Kafka headers used by the Connector into the wrapped `CloudEvent` so that a user can filter on them in their `Processor`. For example, the Events produced by the S3 Connector seem to have the following headers:

```
Headers: {CamelAwsS3BucketName=test-connector-1, CamelAwsS3ContentControl=null, CamelAwsS3ContentDisposition=null, CamelAwsS3ContentEncoding=null, CamelAwsS3ContentLength=59650, CamelAwsS3ContentType=image/jpeg, CamelAwsS3ETag="4721d67d808c4854581fe119a914564e", CamelAwsS3ExpirationTime=null, CamelAwsS3Key=side-monkey.jpeg, CamelAwsS3LastModified=2022-04-29T13:16:01Z, CamelAwsS3Metadata={}, CamelAwsS3ReplicationStatus=null, CamelAwsS3ServerSideEncryption=null, CamelAwsS3StorageClass=null, CamelAwsS3VersionId=null, CamelMessageTimestamp=1651238161000}
```

If I as user wanted to write a `Filter` targeting the `CamelAwsS3Key`, how would I do that? 

My instinct is that we make these headers Extension Context Attributes (https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#extension-context-attributes) and then place the event payload into the `data` element of the `CloudEvent` as we do now. So taking the S3 Connector payload as an example, we'd end up with a `CloudEvent` looking like:

```
{
  "id" : "blah",   
  "source" : "RHOSE",     
  "specversion" : "1.0",
  "camelawss3bucketname": "test-connector-1",   <-- headers
  "camelawss3contentcontrol": "null", <-- headers
  ...
  "data": {
  }
}
```
Then as a user, I can write my `Filter` as follows:

```
{
  "filters": [
    {
      "type": "StringEquals", 
      "key": "camelawss3key",
      "value": "foo"
    }
  ]
}
```

@lucamolteni Can you confirm I'm correct on the structure of the Kafka Event from S3 Connector please?

So basically rule would be:

- If Source Connector uses Kafka Headers, these are hoisted to Extension Context Attributes
- Payload from the Source Connector is preserved as-is into the `data` part of the `CloudEvent`

cheers,

Rob

cc @danielezonca

manager/src/main/java/com/redhat/service/smartevents/manager/api/models/requests/ProcessorRequest.java
kostola: Thanks for the hint. Done :+1:

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
kostola: Good hint. I introduced an additional check on processor type just to be sure (+ tests).

processor/runtime/src/main/java/com/redhat/service/smartevents/processor/actions/source/SourceActionInvoker.java
kostola: @danielezonca @lucamolteni @robpblake I think this discussion is getting big enough to require a dedicated task...wdyt?

robpblake: Thanks @kostola @lucamolteni 👍

sutaakar (approved):

dev/SourceProcessors.md
kostola: I updated all these files and added a new `SOURCES.md` file :+1:

robpblake (approved):
Very nice. Thank you!

SOURCES.md
```diff
@@ -0,0 +1,84 @@
...
+## Parameters of a Source
+
+Each `Source` has 2 parameters to specify:
+- `type`: the type of the `Source`. This must be one of the supported `Source` types listed below
+  - Attempting to use an unknown `Source` type will result in an Error from the Bridge API.
+- `parameters`: A key/value map of configuration parameters for the `Source`
+  - Only string for the `key` and `value` of the parameters are supported.
+  - The required parameters are `Source` specific and documented in the list of supported `Sources`
+
+## Supported Source Types
```
manstis, line 24: @kelvah FYI.. this seems to be the only "Source Processor" we need to worry about in the UI at the moment.

manager/src/main/resources/db/migration/0.1/V0.1.10__update_connector.sql
```diff
@@ -0,0 +1,15 @@
...
+alter table CONNECTOR
+    rename column connector_type to connector_type_id;
+
+-- add new "type" column
+alter table CONNECTOR
+    add column "type" varchar(255);
+
+-- update existing connectors "type" with 'SINK' value (the only possible one before this migration)
+update CONNECTOR
+    set "type"='SINK';
```
manstis, line 11: Would it not be safer to include `where type is null`?

manstis (approved):

manager/src/main/resources/db/migration/0.1/V0.1.10__update_connector.sql
kostola: there can't be source connectors before merging this PR

manstis: Indeed, but there can be afterwards and this is re-deployed to an environment unless we always delete the database.

It was only a comment; I approved anyway.

PR #679, opened 2022-04-26
MGDOBR-636: MC: Return HTTP-410 [GONE] when deleted not HTTP-404 [NOT_FOUND]

manstis:
[https://issues.redhat.com/browse/MGDOBR-636](https://issues.redhat.com/browse/MGDOBR-636)

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.

</details>

### DIFF
--- a/manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
+++ b/manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
@@ -73,9 +73,11 @@ public class ConnectorsApiClientImpl implements ConnectorsApiClient {
             switch (e.getCode()) {
                 case HttpStatus.SC_NOT_FOUND:
                 case HttpStatus.SC_GONE:
+                    //These "exceptions" are handled by RHOSE API as expected states
                     LOGGER.info("Connector with id '{}' could not be found in Connector Namespace '{}'.", connectorExternalId, mcNamespaceId);
                     return null;
                 default:
+                    //All other exceptions can be reported
                     String message =
                             String.format("Failed to retrieve Connector with id '%s', from Connector Namespace '%s' with HTTP Response Code '%s'", connectorExternalId, mcNamespaceId, e.getCode());
                     throw new ConnectorGetException(message, e);
--- a/manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
+++ b/manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
@@ -73,10 +73,12 @@ public class ConnectorsApiClientImpl implements ConnectorsApiClient {
             switch (e.getCode()) {
                 case HttpStatus.SC_NOT_FOUND:
                 case HttpStatus.SC_GONE:
+                    LOGGER.info("Connector could not be found in MC Fleet Manager.");
                     return null;
+                default:
+                    LOGGER.error("Failed to retrieve connector: {}", e.getMessage());
+                    throw new ConnectorGetException("Error while retrieving the connector on MC Fleet Manager", e);
             }
-            LOGGER.error("Failed to retrieve connector: {}", e.getMessage());
-            throw new ConnectorGetException("Error while retrieving the connector on MC Fleet Manager", e);
         }
     }
 
--- a/manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
+++ b/manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
@@ -73,11 +73,12 @@ public class ConnectorsApiClientImpl implements ConnectorsApiClient {
             switch (e.getCode()) {
                 case HttpStatus.SC_NOT_FOUND:
                 case HttpStatus.SC_GONE:
-                    LOGGER.info("Connector could not be found in MC Fleet Manager.");
+                    LOGGER.info("Connector with id '{}' could not be found in Connector Namespace '{}'.", connectorExternalId, mcNamespaceId);
                     return null;
                 default:
-                    LOGGER.error("Failed to retrieve connector: {}", e.getMessage());
-                    throw new ConnectorGetException("Error while retrieving the connector on MC Fleet Manager", e);
+                    String message =
+                            String.format("Failed to retrieve Connector with id '%s', from Connector Namespace '%s' with HTTP Response Code '%s'", connectorExternalId, mcNamespaceId, e.getCode());
+                    throw new ConnectorGetException(message, e);
             }
         }
     }
--- a/manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
+++ b/manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
@@ -90,6 +90,7 @@ public class ConnectorsApiClientImpl implements ConnectorsApiClient {
         try {
             return connectorsAPI.createConnector(true, connectorRequest);
         } catch (ApiException e) {
+            LOGGER.error(e.getMessage());
             throw new ConnectorCreationException("Error while creating the connector on MC Fleet Manager", e);
         }
     }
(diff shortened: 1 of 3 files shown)


sutaakar:
@manstis Can you please check Sonar errors?

manstis:
> @manstis Can you please check Sonar errors?

@sutaakar I have re-worked it slightly. Let's see what SonarCloud makes of it now.

manstis:
@sutaakar SonarCloud is evidently happy with my change!

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
```diff
@@ -67,15 +67,18 @@ public Connector getConnector(String connectorExternalId) {
...
-            }
+            LOGGER.debug("Retrieving Connector with ID '{}'", connectorExternalId);
             return connectorsAPI.getConnector(connectorExternalId);
         } catch (ApiException e) {
-            if (e.getCode() == HttpStatus.SC_NOT_FOUND) {
-                return null;
+            switch (e.getCode()) {
+                case HttpStatus.SC_NOT_FOUND:
+                case HttpStatus.SC_GONE:
+                    LOGGER.info("Connector could not be found in MC Fleet Manager.");
```
robpblake, line 76: @manstis Can we make the log statements a bit more helpful e.g 

```
LOGGER.info("Connector with id '{}' could not be found in Connector Namespace '{}'.", connectorExternalId, mcNamespaceId);
```
That should make it much easier for us to debug problems in the future.

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
```diff
@@ -67,15 +67,18 @@ public Connector getConnector(String connectorExternalId) {
...
+                    return null;
+                default:
+                    LOGGER.error("Failed to retrieve connector: {}", e.getMessage());
+                    throw new ConnectorGetException("Error while retrieving the connector on MC Fleet Manager", e);
```
robpblake, line 80: @manstis Same point here for the `Exception` message. If the `Exception` message is:

```
String message = String.format("Failed to retrieve Connector with id '%s', from Connector Namespace '%s'", connectorExternalId, mcNamespaceId);
throw new ConnectorGetException(message, e)
```
Then we don't need the additional `LOGGER.error` here IMHO. It will be caught and logged in the `ConnectorWorker` giving us a single statement in the log, with the chained stacktrace.

robpblake (changes requested):
@manstis Can we just make the log statements and exception message a bit clearer please to aid in future debugging.

cheers,

Rob

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
manstis: Done.

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
manstis: Done.

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
```diff
@@ -86,6 +90,7 @@ public Connector createConnector(ConnectorRequest connectorRequest) {
         try {
             return connectorsAPI.createConnector(true, connectorRequest);
         } catch (ApiException e) {
+            LOGGER.error(e.getMessage());
```
robpblake, line 93: @manstis Let's be consistent in our approach please. So rather than `LOGGER.error` here (and in the `deleteConnector` method), can we take the same approach as we did in the `getConnector` workflow, so something like:

```
String message = String.format("Failed to create Connector of type '%s', in Connector Namespace '%s'. Received HTTP Response '%s'", connectorEntity.getConnectorType(), mcNamespace, e.getCode());
throw new ConnectorCreationException(message, e);
```

cheers,

Rob

robpblake (changes requested):

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
manstis: @sutaakar I'm not sure @robpblake opinion is correct.

From what I saw on [MGDOBR-634](https://issues.redhat.com/browse/MGDOBR-634) we only get the message from `ConnectorGetException` and the underlying cause is not recorded. I think I should either log the original error here or add `e.printStackTrace()` to `AbstractWorker` exception handler. WDYT?

robpblake: @manstis I think the log statement in `AbstractWorker` needs to change from:

```
LOGGER.info("Failed to delete dependencies for '{}' [{}].\n"
                        + "Work status: {}\n"
                        + "{}",
                        managedResource.getName(),
                        managedResource.getId(),
                        work,
                        e.getMessage());
```

to 

```
LOGGER.info("Failed to delete dependencies for '{}' [{}].\n"
                        + "Work status: {}\n"
                        + "{}",
                        managedResource.getName(),
                        managedResource.getId(),
                        work,
                        e);
```

Note the drop of the `e.getMessage()` call.

cheers,

Rob

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
manstis: I'm not 100% convinced this will show the original cause in our error logging.

The implementation on `main` does `throw new ConnectorGetException("Error while retrieving the connector on MC Fleet Manager", e)` (which is comparable to your proposal) and on [MGDOBR-634](https://issues.redhat.com/browse/MGDOBR-634) we are not seeing the underlying cause being logged; which was the reason for @sutaakar reporting the issue.

Unless I am missing something I don't think your proposal guarantees the underlying cause is logged?

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
sutaakar: So the complete stacktrace with causes is printed in `AbstractWorker`? Fine for me.
So the API client will just wrap the error it gets from external system and pass it above.

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
manstis: Unsurprisingly I missed something stupid.

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
manstis: Thanks @robpblake I will find a rock to lay under until the day passes. Update made.

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsApiClientImpl.java
```diff
@@ -67,15 +67,19 @@ public Connector getConnector(String connectorExternalId) {
...
-                LOGGER.debug(String.format("Retrieving Connector with ID '%s'", connectorExternalId));
-            }
+            LOGGER.debug("Retrieving Connector with ID '{}'", connectorExternalId);
             return connectorsAPI.getConnector(connectorExternalId);
         } catch (ApiException e) {
-            if (e.getCode() == HttpStatus.SC_NOT_FOUND) {
-                return null;
+            switch (e.getCode()) {
+                case HttpStatus.SC_NOT_FOUND:
+                case HttpStatus.SC_GONE:
```
lucamolteni, line 75: Could you please add a clarifying comment on why these two branches are handled differently? 

Something like ```// this is expected, don't do anything ``` and ``` // unknown status code```

lucamolteni (approved):

robpblake (approved):
Thanks @manstis Looks good!

sutaakar:
For some reason the MC wasn't created properly in PR check tests, there is no stacktrace as the operation is async

sutaakar (approved):

manstis:
> OK to merge as the (e2e) failure is investigated and unrelated.

PR #733, opened 2022-05-05
NO JIRA Add a script to delete Bridges and Processors created by E2E

sutaakar:

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>

### DIFF
--- a/dev/bin/interaction/delete-test-resources.sh
+++ b/dev/bin/interaction/delete-test-resources.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+source "$(dirname "${BASH_SOURCE[0]}")/common.sh"
+
+# Script used to delete Bridges and Processors used in E2E tests
+# Script will delete Bridges without Processors, for Bridges with Processors it will delete Processors if they are in proper state
+# Test Bridges are identified using "test-" prefix in their names
+
+# If Bridge contains Processors then only Processors are deleted by the script execution. If user wants to delete also Bridges then the script has to be executed again after some time (so the Processors can be removed first)!
+
+BRIDGES=($(curl -s -H "Authorization: $OB_TOKEN" -X GET $MANAGER_URL/api/v1/bridges | jq '.items[] | select(.name | startswith("test-")) | .id' | tr -d \"))
+
+for i in "${BRIDGES[@]}"
+do
+   NUMBER_OF_PROCESSORS=$(curl -s -H "Authorization: $OB_TOKEN" -X GET $MANAGER_URL/api/v1/bridges/$i/processors | jq '.items | length')
+  if [ "$NUMBER_OF_PROCESSORS" == "0" ]; then
+    echo "No processors in Bridge $i."
+    BRIDGE_STATUS=$(curl -s -H "Authorization: $OB_TOKEN" -X GET $MANAGER_URL/api/v1/bridges/$i | jq '.status' | tr -d \")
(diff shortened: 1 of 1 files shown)


dev/bin/interaction/delete-test-resources.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+source "$(dirname "${BASH_SOURCE[0]}")/common.sh"
+
+# Script used to delete Bridges and Processors used in E2E tests
+# Script will delete Bridges without Processors, for Bridges with Processors it will delete Processors if they are in proper state
+
+BRIDGES=($(curl -s -H "Authorization: $OB_TOKEN" -X GET $MANAGER_URL/api/v1/bridges | jq '.items[] | select(.name | startswith("test-")) | .id' | tr -d \"))
```
jiripetrlik, line 8: I would suggest to move "test-" prefix to a separate variable at the beginning of the script. It will be more clear how the script finds out bridges. Something like BRIDGE_PREFIX="test-".

jiripetrlik (approved):
I have only one minor suggestion for improvement.

dev/bin/interaction/delete-test-resources.sh
sutaakar: Unfortunately it is not so easy as the jq filter is specified using single quote, making properly replacement complicated. I have added some comments mentioning "test-" filtering in name.

sutaakar:
@lucamolteni @r00ta Can you please take a look?

dev/bin/interaction/delete-test-resources.sh
```diff
@@ -0,0 +1,39 @@
...
+      PROCESSOR_STATUS=$(curl -s -H "Authorization: $OB_TOKEN" -X GET $MANAGER_URL/api/v1/bridges/$i/processors/$j | jq '.status' | tr -d \")
+      if [ "$PROCESSOR_STATUS" == "ready" ] || [ "$PROCESSOR_STATUS" == "failed" ]; then
+        echo "Deleting Processor $j"
+        curl -s -H "Authorization: $OB_TOKEN" -X DELETE $MANAGER_URL/api/v1/bridges/$i/processors/$j
+      else
+        echo "Processor $j status $PROCESSOR_STATUS is incompatible with automated deletion, please remove it manually"
+      fi
+    done
+  fi
+done
```
r00ta, line 38: In the case you have processors for the bridge you are not deleting the bridge itself then

sutaakar: Yes, because right now it is not possible to delete Bridges which contains Processes.
So the possibility was either to implement some sort of waiting loop - that would bring some complexity..... or leave it like this so user should rerun the script several times. So when there is a Bridge with Processors then first run will delete Processors, user should wait some time and then rerun the script again to remove Bridges.

r00ta: Could you add a comment about this? So that the expectation is that this script is not going to delete all the resources at the first execution but it has to be run multiple times

r00ta (approved):

dev/bin/interaction/delete-test-resources.sh
sutaakar: Added a comment, added an echo mentioning a need to rerun the script if the Bridge has some Processors

PR #771, opened 2022-05-19
POC - MGDOBR 404

lucamolteni:
[MGDOBR-404](https://issues.redhat.com/browse/MGDOBR-404)

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>


dev/bin/interaction/create-action.sh
```diff
@@ -57,8 +57,8 @@ elif [ "${action_type}" = 'slack-plain' ]; then
    "action": {
       "type": "Slack",
       "parameters": {
-         "channel": "mc",
-         "webhookUrl": '"\"$SLACK_WEBHOOK_URL\""'
+         "slack_channel": "mc",
+         "slack_webhook_url": '"\"$SLACK_WEBHOOK_URL\""'
       }
```
r00ta, line 62: Just brainstorming: we could change the `type` to `Connector` and provide the connector type inside the `parameters`. In this way we can better separate what is a connector and what is not.

processor/configuration/src/main/java/com/redhat/service/smartevents/processor/actions/slack/SlackConnectorAction.java
```diff
@@ -0,0 +1,20 @@
+package com.redhat.service.smartevents.processor.actions.slack;
+
+import javax.enterprise.context.ApplicationScoped;
+
+import com.redhat.service.smartevents.processor.actions.generic.GenericConnectorAction;
+
+@ApplicationScoped
+public class SlackConnectorAction extends GenericConnectorAction {
```
r00ta, line 8: Brainstorming: we could remove all the connector specific classes with the following workflow: 
1) We get a connector action
2) if the `type` is for example `slack_sink_0.1`, we look for the `slack_sink_0.1.json` file
3) we use `slack_sink_0.1.json` for the validation

In this way we can
1) add a new connector by just adding the json in the `resources` folder
2) no need to write a single line of code for any connector

On the other side, if we keep connector-specific classes we might add additional checks on the payload. For example if a parameter is an URL, the json schema validation would just check that it is a string, but we could also check that the value is a valid URL.

r00ta (commented):
Thank you for starting with this, I think it will improve a lot the way we will manage the integration with connectors

vaibhavjainwiz:
@Brainstorming : Currently we have our own contract which we expose to external user. Which provide us full control how we want to provide support for it in future. For example In future if we have to capture some extra field for Slack action then we could add new parameter.
```
"name": "$action_name"',
   "action": {
      "type": "Slack",
      "parameters": {
         "channel": "mc",
         "webhookUrl": '"$SLACK_WEBHOOK_URL"'
      }
   }
```
On other hand if we choose to use MC schema then we can't add/remove field on bases of our requirement. It will also tightly bind RHOSE with MC only. 
**Cons : **

- If our customer requirs to connect with an application whose MC is not yet implemented then we can't write custome connect logic for it.
- If MC choose to add some new paramter or rename existing paramter then our application users start getting error.

PR #784, opened 2022-05-24
chore: test update to test PR#776

gowriswarupk:
Hi,

Just a test update with a new line to test the new openAPI updater change. 

thanks.


manager/src/main/resources/application.properties
```diff
@@ -11,7 +11,7 @@ quarkus.swagger-ui.always-include=true
 quarkus.smallrye-openapi.store-schema-directory=target/generated/
 quarkus.smallrye-openapi.info-title=Red Hat Openshift SmartEvents Fleet Manager
 quarkus.smallrye-openapi.info-version=0.0.1
-quarkus.smallrye-openapi.info-description=The api exposed by the fleet manager of the SmartEvents service.
+quarkus.smallrye-openapi.info-description=The API exposed by the fleet manager of the SmartEvents service.
```
danielezonca, line 14: Please update also the saved file to match 
https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/openapi/openapi.yaml#L5

.github/workflows/CI.yaml
```diff
@@ -69,5 +69,5 @@ jobs:
       - name: Validate openapi spec
         if: ${{ always() }}
         run: |
-          differences=$(diff openapi/openapi.yaml manager/target/generated/openapi.yaml)
-          if [ $? -ne 0 ]; then echo "openapi/openapi.yaml is out of sync. Please move the manager/target/generated/openapi.yaml into openapi/openapi.yaml"; false ; fi
+          differences=$(diff openapi/rhose-api.yaml manager/target/generated/rhose-api.yaml)
```
danielezonca, line 72: This job runs using the `pull_request_target` event to safely access secrets (aka it takes job definition from main and not from the branch) so the PR check will be red until we merge, as alternative we can do an initial PR that does only the CI change and is backward compatible: like, if `openapi.yaml` exists, use it, otherwise use `rhose-api.yaml`.
It might be an overkill to be honest 🤔 

@r00ta @lucamolteni @sutaakar @manstis 
Any preference?

sutaakar: It is fine for me to merge it will failing PR check in such case, the functionality can be tested on merged code.

PR #755, opened 2022-05-13
[MGDOBR-59] Sensitive Parameter Handling for Source and Actions

robpblake:
[MGDOBR-59](https://issues.redhat.com/browse/MGDOBR-59)

This PR is the first draft of sensitive parameter handling for `Actions` and `Sources`. Broadly it introduces the following changes:

1. Sensitive Parameters for `Actions` and `Sources` are no longer stored in the database in plaintext. We store a masked value. This masked value is returned from the API as now, but obviously does not contain the sensitive value. This allows customers to see what they have configured as parameters, but not the value where it is sensitive.
2. Sensitive Parameters are stored into a Vault implementation. In this case, they are stored into our AWS Secrets Manager implementation
3. We no longer store the definition of the requested `Connector` in the database as it has the potential to contain sensitive data. Instead, we assemble the `Connector` request body at runtime 

This is a draft. Code standard is not final.

## PR owner checklist

Please make sure that your PR meets the following requirements:

- [ ] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/main/kogito-build/kogito-ide-config).  
  *Run `mvn clean verify -DskipTests` so that imports are correctly set.*
- [ ] Your commit messages are clear and reference the JIRA issue e.g: "[MGDOBR-1] - $clear_explanation_of_what_you_did"
- [ ] The layers in the `kustomize` folder have been updated accordingly.
- [ ] All new functionality is tested
- [ ] Pull Request title is properly formatted: `MGDOBR-XYZ Subject`
- [ ] Pull Request contains link to the JIRA issue
- [ ] Pull Request contains link to any dependent or related Pull Request
- [ ] Pull Request contains description of the issue
- [ ] Pull Request does not include fixes for issues other than the main ticket

## PR reviewer(s) checklist

- [ ] Check if code and Github action workflows have been modified and if the modification is safe
- [ ] If the modification is safe, add the `safe to test` label

<details>
<summary>
How to trigger pipelines and use the bots:
</summary>

- <b>Run the end to end pipeline</b>  
  Annotate the pull request with the label: `safe to test`. If you want to run the pipeline again, remove and add it again.

- <b>Rebase the pull request</b>  
  Comment with: `/rebase`.
- <b>Deploy BOT</b>
  Comment with `/deploy <dev,stable> when the PR has been merged to deploy to a target environment.

</details>

PR #856, opened 2022-06-16
MGDOBR-516: Allow a user to update Action Parameters on a Sink Processor

manstis:
[https://issues.redhat.com/browse/MGDOBR-516](https://issues.redhat.com/browse/MGDOBR-516)

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>

### DIFF
--- a/manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
+++ b/manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
@@ -118,7 +118,6 @@ public class ProcessorServiceImpl implements ProcessorService {
         newProcessor.setBridge(bridge);
         newProcessor.setShardId(shardService.getAssignedShardId(newProcessor.getId()));
         newProcessor.setOwner(owner);
-        newProcessor.setGeneration(0);
 
         Set<BaseFilter> requestedFilters = processorRequest.getFilters();
         String requestedTransformationTemplate = processorRequest.getTransformationTemplate();
@@ -235,7 +234,7 @@ public class ProcessorServiceImpl implements ProcessorService {
 
         // Processor, Connector and Work should always be created in the same transaction
         // Since updates to the Action are unsupported we do not need to update the Connector record.
-        connectorService.updateConnectorEntityDefinition(existingProcessor);
+        connectorService.updateConnectorEntity(existingProcessor);
         workManager.schedule(existingProcessor);
         metricsService.onOperationStart(existingProcessor, MetricsOperation.MODIFY);
 
--- a/manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
+++ b/manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
@@ -114,9 +114,11 @@ public class ProcessorServiceImpl implements ProcessorService {
         newProcessor.setName(processorRequest.getName());
         newProcessor.setSubmittedAt(ZonedDateTime.now());
         newProcessor.setStatus(ManagedResourceStatus.ACCEPTED);
+        newProcessor.setDependencyStatus(ManagedResourceStatus.ACCEPTED);
         newProcessor.setBridge(bridge);
         newProcessor.setShardId(shardService.getAssignedShardId(newProcessor.getId()));
         newProcessor.setOwner(owner);
+        newProcessor.setGeneration(0);
 
         Set<BaseFilter> requestedFilters = processorRequest.getFilters();
         String requestedTransformationTemplate = processorRequest.getTransformationTemplate();
@@ -175,9 +177,11 @@ public class ProcessorServiceImpl implements ProcessorService {
                     customerId));
         }
         ProcessorDefinition existingDefinition = existingProcessor.getDefinition();
+        Set<BaseFilter> existingFilters = existingDefinition.getFilters();
+        String existingTransformationTemplate = existingDefinition.getTransformationTemplate();
         Action existingAction = existingDefinition.getRequestedAction();
-        Action existingResolvedAction = existingDefinition.getResolvedAction();
         Source existingSource = existingDefinition.getRequestedSource();
+        Action existingResolvedAction = existingDefinition.getResolvedAction();
 
         // Validate update.
         // Name cannot be updated.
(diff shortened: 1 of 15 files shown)


manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
```diff
@@ -187,20 +187,28 @@ public Processor updateProcessor(String bridgeId, String processorId, String cus
         if (!Objects.equals(existingProcessor.getType(), processorRequest.getType())) {
             throw new BadRequestException("It is not possible to update the Processor's Type.");
         }
-        // See https://issues.redhat.com/browse/MGDOBR-516 for updating Action support
-        if (!Objects.equals(existingAction, processorRequest.getAction())) {
-            throw new BadRequestException("It is not possible to update the Processor's Action.");
+        if (existingProcessor.getType() == ProcessorType.SINK) {
+            if (!Objects.equals(existingAction.getType(), processorRequest.getAction().getType())) {
```
lucamolteni, line 191: Why duplicate the if only for the different error message?

manager/src/test/java/com/redhat/service/smartevents/manager/ProcessorServiceConnectorTest.java
```diff
@@ -98,11 +99,14 @@ void createConnectorSuccess() {
         ProcessorRequest processorRequest = new ProcessorRequest("ManagedConnectorProcessor", slackAction);
 
         //Emulate successful External Connector creation
-        Connector externalConnector = stubbedExternalConnector("connectorExternalId");
+        Connector externalConnector = spy(stubbedExternalConnector("connectorExternalId"));
```
lucamolteni, line 102: Why the spy here?

lucamolteni (changes requested):

lucamolteni:
> I've verified updates are applied for a WebHook action locally.

I thought this was for the sink processor

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/ConnectorWorker.java
```diff
@@ -90,6 +95,17 @@ public ConnectorEntity createDependencies(Work work, ConnectorEntity connectorEn
             return connectorEntity;
         }
         if (status.getState() == ConnectorState.READY) {
+            // If the Connector is ready but differs to that required we need to patch it.
+            JsonNode existingConnectorDefinition = MAPPER.valueToTree(connector.getConnector());
+            JsonNode updatedConnectorDefinition = connectorEntity.getDefinition();
+            if (!Objects.equals(existingConnectorDefinition, updatedConnectorDefinition)) {
```
sutaakar, line 101: Some connectors contain sensitive information which are not returned, i.e. Slack WebHook URL. As a result the `Objects.equals` will always return false, the Slack Processor will never change state to ready.

robpblake: Hmmm..that's a pain. Couple of initial thoughts here @manstis 

1. Have a `generation` field on the `ManagedResource` which is monotonically increased by 1 each time we accept an update from the user. This would let us know how many times a resource has been updated in total and would solve your problem here in that if the `generation` you're working on is > 1, then you know it has to be an update and a `patch` is required to RHOC. 
2. Have the ability to define an operation on the Work item e.g `CREATE`, `MODIFY`, `DELETE`
3. Something else :-)

I think I lean towards 1 as it's basically what the kube API does. Just some suggestions for you.

cheers,

Rob

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
manstis: What would you propose? 

I cannot use the generic term `Gateway` as that is an internal detail. 

`Action` and `Source` are known terms to the User so surely it's better to keep terminology in error messages consistent with their understanding?

manager/src/test/java/com/redhat/service/smartevents/manager/ProcessorServiceConnectorTest.java
manstis: Because I was mocking invocation of it's `getConnector()` on L#108 but this can be better achieved by subclassing it and overriding the method. I cannot _set_ the connector value when constructing the value as it does not exist until runtime and needs to be read from the database.

manstis:
@lucamolteni I've made an update.

> I've verified updates are applied for a WebHook action locally.

> I thought this was for the sink processor

`WebHook` is a Sink processor. Have I misunderstood you?

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
```diff
@@ -187,20 +187,28 @@ public Processor updateProcessor(String bridgeId, String processorId, String cus
...
+                ? resolveSource(processorRequest.getSource(), customerId, bridgeId, processorId)
+                : resolveAction(processorRequest.getAction(), customerId, bridgeId, processorId);
         ProcessorDefinition updatedDefinition = existingProcessor.getType() == ProcessorType.SOURCE
-                ? new ProcessorDefinition(updatedFilters, updatedTransformationTemplate, existingSource, existingResolvedAction)
-                : new ProcessorDefinition(updatedFilters, updatedTransformationTemplate, existingAction, existingResolvedAction);
+                ? new ProcessorDefinition(updatedFilters, updatedTransformationTemplate, updatedSource, updatedResolvedAction)
+                : new ProcessorDefinition(updatedFilters, updatedTransformationTemplate, updatedAction, updatedResolvedAction);
 
         // No need to update CRD if the definition is unchanged
         if (existingDefinition.equals(updatedDefinition)) {
```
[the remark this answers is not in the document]
manstis: No, since this check also considers Filters and Transformations. 

However I will need to refactor it given @sutaakar comments about secret parameters too.

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/ConnectorWorker.java
manstis: @sutaakar We don't seem to _hide_ any parameters at the moment; but the problem will become an issue with https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/755. I have tested this PR locally with a WebHook `Action` and successfully updated the URL.. but of course it will become a problem so I will look into a solution.

manstis: @sutaakar @robpblake If you agree with my comment above; then it is best fixed in that PR where we can [retrieve](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/755/files#diff-fb73e1aeb94e2e7cdd2a9b560302e23913c137a3ca8d68c5bf4b3602044ff889R61-R71) *all* parameters (public and secret) and compare to that deployed to Managed Connectors. That seems the most viable solution. WDYT?

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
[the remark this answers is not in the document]
manstis: @robpblake This is a short-circuit to prevent an unnecessary CRD provisioning cycle if the definition had not changed.

However now we have the approaching issue of _secret_ parameters this check can no longer work.

I don't think we have much of an option other than to remove it.

WDYT?

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/ConnectorWorker.java
manstis: @robpblake @sutaakar I think it may be best to handle the oncoming issue in the PR I referenced; where the issue is introduced. That PR [builds](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/8eeffd45d19018b5c7c589d877e08aa9c0e35f68/manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/ConnectorWorker.java#L133) the `Connector` definition from both the [public and secret parameters](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/8eeffd45d19018b5c7c589d877e08aa9c0e35f68/manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsServiceImpl.java#L75-L100) before invoking the RHOC API.

manstis: @sutaakar 

> (that is the reason for failing E2E PR check)

Are you sure they fail for that reason? The change to hide _secret_ parameters isn't merged yet.

robpblake: @manstis The secret parameters are hidden by the RHOC API. It does not return sensitive parameters on the `GET` of the Connector details...

It would be more obvious if the value was returned masked or using some sort of redacted string e.g something like this:

```
{
  "sensitive_value": "******"
}
```

rather than

```
{
 "sensitive_value": ""
}
```

manstis: > The secret parameters are hidden by the RHOC API. It does not return sensitive parameters on the GET of the Connector details...

Yikes. That's going to affect everything.

manstis:
@robpblake @sutaakar @kostola I've put this PR into draft while I think of what to do.

lucamolteni:
> @lucamolteni I've made an update.
> 
> > I've verified updates are applied for a WebHook action locally.
> 
> > I thought this was for the sink processor
> 
> `WebHook` is a Sink processor. Have I misunderstood you?

My bad, I probably have mixed those up :) Sorry

lucamolteni (approved):

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/ConnectorWorker.java
```diff
@@ -90,6 +91,16 @@ public ConnectorEntity createDependencies(Work work, ConnectorEntity connectorEn
             return connectorEntity;
         }
         if (status.getState() == ConnectorState.READY) {
+            // If the Connector is ready but differs to that required we need to patch it.
+            JsonNode updatedConnectorDefinition = connectorEntity.getDefinition();
+            if (connectorEntity.getGeneration() > 0) {
+                LOGGER.debug("Managed Connector for '{}' [{}] was found but with a different definition. Patching definition.",
+                        connectorEntity.getName(),
+                        connectorEntity.getId());
+                connectorsApi.updateConnector(connector.getId(), updatedConnectorDefinition);
```
sutaakar, line 100: @manstis For some reason this call doesn't actually patch the connector, not sure why. Have you tried it by yourself?

manstis: That'd explain things then.

manstis:
This PR is blocked by https://issues.redhat.com/browse/MGDCTRS-1080.

manstis:
@sutaakar The issue with RHOC has been resolved and I can confirm updates to a Slack Processor now work.

@robpblake Ready for your eyes when you have some spare time.

sutaakar (approved):
Looks good to me, works as expected

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
```diff
@@ -114,9 +114,11 @@ private Processor doCreateProcessor(Bridge bridge, String customerId, String own
         newProcessor.setName(processorRequest.getName());
         newProcessor.setSubmittedAt(ZonedDateTime.now());
         newProcessor.setStatus(ManagedResourceStatus.ACCEPTED);
+        newProcessor.setDependencyStatus(ManagedResourceStatus.ACCEPTED);
         newProcessor.setBridge(bridge);
         newProcessor.setShardId(shardService.getAssignedShardId(newProcessor.getId()));
         newProcessor.setOwner(owner);
+        newProcessor.setGeneration(0);
```
robpblake, line 121: Nit pick: We don't really need this call as that's the default for a `long`

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsService.java
```diff
@@ -7,4 +7,6 @@ public interface ConnectorsService {
     void createConnectorEntity(Processor processor);
 
     void deleteConnectorEntity(Processor processor);
+
+    void updateConnectorEntityDefinition(Processor processor);
```
robpblake, line 11: Nitpick: For consistency across the methods of this class, I think this should be named `updateConnectorEntity`

robpblake (approved):
@manstis Thanks very much! Some minor nitpick comments from me. See what you think. If you disagree, all good to merge from my perspective.

cheers,

Rob

manager/src/main/java/com/redhat/service/smartevents/manager/connectors/ConnectorsService.java
manstis: Done.

PR #875, opened 2022-06-23
MGDOBR-779: Make BridgeInstance error handling strategy editable

manstis:
[https://issues.redhat.com/browse/MGDOBR-779](https://issues.redhat.com/browse/MGDOBR-779)

Following the virtual Face to Face the approach taken in this PR was considered suitable.

Details of the Bridge instance status will be available when [MGDOBR-52](https://issues.redhat.com/browse/MGDOBR-52) is completed.

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>

### DIFF
--- a/manager/src/main/java/com/redhat/service/smartevents/manager/BridgesServiceImpl.java
+++ b/manager/src/main/java/com/redhat/service/smartevents/manager/BridgesServiceImpl.java
@@ -4,6 +4,7 @@ import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
@@ -15,6 +16,7 @@ import org.slf4j.LoggerFactory;
 
 import com.redhat.service.smartevents.infra.api.APIConstants;
 import com.redhat.service.smartevents.infra.exceptions.definitions.user.AlreadyExistingItemException;
+import com.redhat.service.smartevents.infra.exceptions.definitions.user.BadRequestException;
 import com.redhat.service.smartevents.infra.exceptions.definitions.user.BridgeLifecycleException;
 import com.redhat.service.smartevents.infra.exceptions.definitions.user.ItemNotFoundException;
 import com.redhat.service.smartevents.infra.models.ListResult;
@@ -99,6 +101,61 @@ public class BridgesServiceImpl implements BridgesService {
         return bridge;
     }
 
+    @Override
+    @Transactional
+    public Bridge updateBridge(String bridgeId, String customerId, BridgeRequest bridgeRequest) {
+        // Extract existing definition
+        Bridge existingBridge = getBridge(bridgeId, customerId);
+
+        if (!existingBridge.isActionable()) {
+            throw new BridgeLifecycleException(String.format("Bridge with id '%s' for customer '%s' is not in an actionable state.",
+                    bridgeId,
+                    customerId));
+        }
+
+        BridgeDefinition existingDefinition = existingBridge.getDefinition();
+
+        // Validate update.
+        // Name cannot be updated.
+        if (!Objects.equals(existingBridge.getName(), bridgeRequest.getName())) {
+            throw new BadRequestException("It is not possible to update the Bridge's name.");
+        }
+        // Cloud Provider cannot be updated.
+        if (!Objects.equals(existingBridge.getCloudProvider(), bridgeRequest.getCloudProvider())) {
+            throw new BadRequestException("It is not possible to update the Bridge's Cloud Provider.");
+        }
+        // Cloud Region cannot be updated.
+        if (!Objects.equals(existingBridge.getRegion(), bridgeRequest.getRegion())) {
+            throw new BadRequestException("It is not possible to update the Bridge's Region.");
+        }
+
+        // Construct updated definition
+        Action updatedErrorHandler = bridgeRequest.getErrorHandler();
+        BridgeDefinition updatedDefinition = new BridgeDefinition(updatedErrorHandler);
+
+        // No need to update CRD if the definition is unchanged
+        if (Objects.equals(existingDefinition, updatedDefinition)) {
+            return existingBridge;
+        }
+
+        // Create new definition copying existing properties
+        existingBridge.setModifiedAt(ZonedDateTime.now());
+        existingBridge.setStatus(ManagedResourceStatus.ACCEPTED);
+        existingBridge.setDependencyStatus(ManagedResourceStatus.ACCEPTED);
+        existingBridge.setDefinition(updatedDefinition);
+        existingBridge.setGeneration(existingBridge.getGeneration() + 1);
+
+        // Bridge and Work should always be created in the same transaction
+        workManager.schedule(existingBridge);
+        metricsService.onOperationStart(existingBridge, MetricsOperation.MODIFY);
+
+        LOGGER.info("Bridge with id '{}' for customer '{}' has been marked for update",
+                existingBridge.getId(),
+                existingBridge.getCustomerId());
+
+        return existingBridge;
+    }
+
     @Transactional
     @Override
     public Bridge getBridge(String id) {
@@ -136,19 +193,17 @@ public class BridgesServiceImpl implements BridgesService {
     @Transactional
     public void deleteBridge(String id, String customerId) {
         Long processorsCount = processorService.getProcessorsCount(id, customerId);
-        ListResult<Processor> hiddenProcessors = processorService.getHiddenProcessors(id, customerId);
+        Optional<Processor> errorHandler = processorService.getErrorHandler(id, customerId);
 
-        if (processorsCount != hiddenProcessors.getTotal()) {
+        if (processorsCount > (errorHandler.isPresent() ? 1 : 0)) {
             // See https://issues.redhat.com/browse/MGDOBR-43
             throw new BridgeLifecycleException("It is not possible to delete a Bridge instance with active Processors.");
         }
 
         Bridge bridge = findByIdAndCustomerId(id, customerId);
-        if (!isBridgeDeletable(bridge)) {
+        if (!bridge.isActionable()) {
             throw new BridgeLifecycleException("Bridge could only be deleted if its in READY/FAILED state.");
         }
-
-        hiddenProcessors.getItems().forEach(p -> processorService.deleteProcessor(id, p.getId(), customerId));
 
         LOGGER.info("Bridge with id '{}' for customer '{}' has been marked for deletion", bridge.getId(), bridge.getCustomerId());
 
@@ -159,11 +214,6 @@ public class BridgesServiceImpl implements BridgesService {
         metricsService.onOperationStart(bridge, MetricsOperation.DELETE);
 
         LOGGER.info("Bridge with id '{}' for customer '{}' has been marked for deletion", bridge.getId(), bridge.getCustomerId());
-    }
-
-    private boolean isBridgeDeletable(Bridge bridge) {
-        // bridge could only be deleted if its in READY or FAILED state
-        return bridge.getStatus() == ManagedResourceStatus.READY || bridge.getStatus() == ManagedResourceStatus.FAILED;
     }
 
     @Transactional
--- a/manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorService.java
+++ b/manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorService.java
@@ -1,6 +1,7 @@
 package com.redhat.service.smartevents.manager;
 
 import java.util.List;
+import java.util.Optional;
 
 import com.redhat.service.smartevents.infra.models.ListResult;
 import com.redhat.service.smartevents.infra.models.QueryProcessorResourceInfo;
@@ -15,9 +16,13 @@ public interface ProcessorService {
 
     Processor createProcessor(String bridgeId, String customerId, String owner, ProcessorRequest processorRequest);
 
+    Processor updateProcessor(String bridgeId, String processorId, String customerId, ProcessorRequest processorRequest);
+
+    Optional<Processor> getErrorHandler(String bridgeId, String customerId);
+
     Processor createErrorHandlerProcessor(String bridgeId, String customerId, String owner, ProcessorRequest processorRequest);
 
-    Processor updateProcessor(String bridgeId, String processorId, String customerId, ProcessorRequest processorRequest);
+    Processor updateErrorHandlerProcessor(String bridgeId, String processorId, String customerId, ProcessorRequest processorRequest);
 
     List<Processor> findByShardIdWithReadyDependencies(String shardId);
 
@@ -25,9 +30,7 @@ public interface ProcessorService {
 
     Long getProcessorsCount(String bridgeId, String customerId);
 
-    ListResult<Processor> getUserVisibleProcessors(String bridgeId, String customerId, QueryProcessorResourceInfo queryInfo);
-
-    ListResult<Processor> getHiddenProcessors(String bridgeId, String customerId);
+    ListResult<Processor> getProcessors(String bridgeId, String customerId, QueryProcessorResourceInfo queryInfo);
 
     void deleteProcessor(String bridgeId, String processorId, String customerId);
 
--- a/manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
+++ b/manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
@@ -4,6 +4,7 @@ import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 
 import javax.enterprise.context.ApplicationScoped;
@@ -37,7 +38,6 @@ import com.redhat.service.smartevents.manager.dao.ProcessorDAO;
 import com.redhat.service.smartevents.manager.metrics.MetricsOperation;
 import com.redhat.service.smartevents.manager.metrics.MetricsService;
 import com.redhat.service.smartevents.manager.models.Bridge;
-import com.redhat.service.smartevents.manager.models.ManagedResource;
 import com.redhat.service.smartevents.manager.models.Processor;
 import com.redhat.service.smartevents.manager.providers.InternalKafkaConfigurationProvider;
 import com.redhat.service.smartevents.manager.providers.ResourceNamesProvider;
@@ -90,12 +90,21 @@ public class ProcessorServiceImpl implements ProcessorService {
     }
 
     @Override
+    public Optional<Processor> getErrorHandler(String bridgeId, String customerId) {
+        ListResult<Processor> hiddenProcessors = processorDAO.findHiddenByBridgeIdAndCustomerId(bridgeId, customerId, new QueryProcessorResourceInfo());
+        return hiddenProcessors.getItems()
+                .stream()
+                .filter(p -> p.getType() == ProcessorType.ERROR_HANDLER)
+                .findFirst();
+    }
+
+    @Override
     @Transactional
     public Processor createProcessor(String bridgeId, String customerId, String owner, ProcessorRequest processorRequest) {
         // We cannot deploy Processors to a Bridge that is not available. This throws an Exception if the Bridge is not READY.
         Bridge bridge = bridgesService.getReadyBridge(bridgeId, customerId);
 
-        return doCreateProcessor(bridge, customerId, owner, processorRequest.getType(), processorRequest);
+        return doCreateProcessor(bridge, customerId, owner, processorRequest.getType(), processorRequest, 0);
     }
 
     @Override
@@ -105,13 +114,13 @@ public class ProcessorServiceImpl implements ProcessorService {
 
         ProcessorType processorType = processorRequest.getType();
         if (processorType != ProcessorType.SINK) {
-            throw new InternalPlatformException("Unable to configure error handler");
+            throw new InternalPlatformException("Unable to create Error Handler processor. An Action must be defined.");
         }
 
-        return doCreateProcessor(bridge, customerId, owner, ProcessorType.ERROR_HANDLER, processorRequest);
+        return doCreateProcessor(bridge, customerId, owner, ProcessorType.ERROR_HANDLER, processorRequest, bridge.getGeneration());
     }
 
-    private Processor doCreateProcessor(Bridge bridge, String customerId, String owner, ProcessorType processorType, ProcessorRequest processorRequest) {
+    private Processor doCreateProcessor(Bridge bridge, String customerId, String owner, ProcessorType processorType, ProcessorRequest processorRequest, long generation) {
         String bridgeId = bridge.getId();
         if (processorDAO.findByBridgeIdAndName(bridgeId, processorRequest.getName()) != null) {
             throw new AlreadyExistingItemException("Processor with name '" + processorRequest.getName() + "' already exists for bridge with id '" + bridgeId + "' for customer '" + customerId + "'");
@@ -125,6 +134,7 @@ public class ProcessorServiceImpl implements ProcessorService {
         newProcessor.setDependencyStatus(ManagedResourceStatus.ACCEPTED);
         newProcessor.setBridge(bridge);
         newProcessor.setShardId(shardService.getAssignedShard(newProcessor.getId()).getId());
+        newProcessor.setGeneration(generation);
         newProcessor.setOwner(owner);
 
         Set<BaseFilter> requestedFilters = processorRequest.getFilters();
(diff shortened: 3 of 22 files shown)


manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
```diff
@@ -166,9 +166,6 @@ public Processor updateProcessor(String bridgeId, String processorId, String cus
 
         // Extract existing definition
         Processor existingProcessor = getProcessor(bridgeId, processorId, customerId);
-        if (existingProcessor.getType() == ProcessorType.ERROR_HANDLER) {
```
manstis, line 169: It is now possible to update an Error Handler Processor.

manager/src/main/java/com/redhat/service/smartevents/manager/BridgesServiceImpl.java
```diff
@@ -135,12 +182,10 @@ public void deleteBridge(String id, String customerId) {
         }
 
         Bridge bridge = findByIdAndCustomerId(id, customerId);
-        if (!isBridgeDeletable(bridge)) {
+        if (!isBridgeActionable(bridge)) {
             throw new BridgeLifecycleException("Bridge could only be deleted if its in READY/FAILED state.");
         }
 
-        hiddenProcessors.getItems().forEach(p -> processorService.deleteProcessor(id, p.getId(), customerId));
```
manstis, line 142: I have moved deletion of the Error Handler processor to `BridgeWorker` for consistency.

r00ta:
Please do not merge this until I remove the label -> I'm checking that all the knative changes are working on main/dev

manager/src/main/java/com/redhat/service/smartevents/manager/BridgesServiceImpl.java
```diff
@@ -90,6 +92,61 @@ public Bridge createBridge(String customerId, String organisationId, String owne
...
+
+        if (!existingBridge.isActionable()) {
+            throw new BridgeLifecycleException(String.format("Bridge with id '%s' for customer '%s' is not in an actionable state.",
+                    bridgeId,
+                    customerId));
+        }
+        // We don't want to allow updating an Error Handler if the previous configuration has not been provisioned.
+        // This can lead to problems trying to remove the Error Handler if it is not in READY state.
+        ListResult<Processor> hiddenProcessors = processorService.getHiddenProcessors(bridgeId, customerId);
+        if (!hiddenProcessors.getItems().stream().allMatch(ManagedResource::isActionable)) {
```
manstis, line 109: This is interesting....

To make things "simple" we cannot allow an Error Handler to be updated unless it is `READY`.

However Users know nothing of Error Handler processor provisioning as we chose to keep it internal.

Any better ideas to how to solve the problem?

We could make the `Bridge` stay in `PREPARING` until it's Error Handler is `READY` instead.

manstis:
Tried and tested locally and it works as expected:
# Create a `Bridge` without an Error Handler - OK
# Update the `Bridge` and add an Error Handler - OK
# Update the `Bridge` and remove the Error Handler - OK

manstis:
I need to write a few more Unit Tests for the latest commits.

manager/src/main/java/com/redhat/service/smartevents/manager/BridgesServiceImpl.java
sutaakar: Fine for me to keep `Bridge` in `PROVISIONING` state until all the components are fully provisioned.

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/BridgeWorker.java
```diff
@@ -73,41 +75,47 @@ public Bridge createDependencies(Work work, Bridge bridge) {
...
-        String customerId = bridge.getCustomerId();
-        ListResult<Processor> processors = processorService.getHiddenProcessors(bridgeId, customerId);
-
+        // If an Error Handler processor exists assume it is to be updated otherwise create it!
         // This assumes we can only have one ErrorHandler Processor per Bridge
-        if (processors.getTotal() > 0) {
-            return;
+        if (hiddenProcessors.getTotal() > 0) {
+            // Update Error Handler processor if already present
+            Processor errorHandler = hiddenProcessors.getItems().get(0);
```
sutaakar, line 107: a nitpick - maybe throw an exception in case there are several hidden Processors as it is not possible to identify which one is error handler?

manager/src/main/java/com/redhat/service/smartevents/manager/BridgesServiceImpl.java
danielezonca: We had an initial discussion about this with @kelvah because I agree that we need to make this time to apply a change visible to users.
I agree that we can leverage `Bridge` intermediate statuses to represent this.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
```diff
@@ -86,6 +86,12 @@ public void createBridgeExecutor(ProcessorDTO processorDTO) {
 
             // create or update the secrets for the bridgeExecutor
             createOrUpdateBridgeExecutorSecret(bridgeExecutor, processorDTO);
+        } else {
```
sutaakar, line 89: In what case is this branch called? Why do we need this new code here?

manager/src/main/java/com/redhat/service/smartevents/manager/api/user/BridgesAPI.java
```diff
@@ -94,6 +95,23 @@ public Response createBridge(@Valid BridgeRequest bridgeRequest) {
         return Response.accepted(bridgesService.toResponse(bridge)).build();
     }
 
+    @APIResponses(value = {
+            @APIResponse(description = "Accepted.", responseCode = "202",
+                    content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(implementation = BridgeResponse.class))),
+            @APIResponse(description = "Bad request.", responseCode = "400", content = @Content(mediaType = MediaType.APPLICATION_JSON)),
+            @APIResponse(description = "Unauthorized.", responseCode = "401"),
+            @APIResponse(description = "Forbidden.", responseCode = "403"),
```
sutaakar, line 103: Missing 404 response code

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
manstis: It _could_ happen if an attempt is made to re-provision a `Processor` without change.

I experienced this exact issue with `Bridge` on this PR (when updating an Error Handler the `Bridge` itself is not changed but our provisioning mechanism passes it through the Shard Operator by virtue of the `Bridge` moving from `ACCEPTED`->`PREPARING` and its dependencies being `READY`).

I'd err on the side of caution and keep this code in the `BridgeExecutorServiceImpl` too. We didn't know it was an issue for `Bridge`'s until this PR to update the Error Handler. It does not harm, but does prevent a potential issue in the future.

sutaakar:
Code looks ok otherwise, briefly checked locally and seems to work properly. Will do more tests once comments are addressed.

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/BridgeWorker.java
manstis: I have refactored slightly to handle the first Error Handler processor found.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
sutaakar: ok, fine for me

manstis:
I have tested the following scenarios locally:
- Create a `Bridge` without Error Handler - OK
- Add an Error Handler to the `Bridge` - OK
- Update the Error Handler on the `Bridge` - OK
- Delete the Error Handler from the `Bridge` - OK
- Delete the `Bridge` - OK
- Create a `Bridge` with an Error Handler - OK
- Update the Error Handler on the `Bridge` - OK
- Delete the `Bridge` (and the Error Handler is deleted automatically) - OK

manager/src/main/java/com/redhat/service/smartevents/manager/BridgesServiceImpl.java
robpblake: Just to be clear on the implications for the Bridge `status` lifecycle here:

- If the error handler fails to deploy, is the containing Bridge instance marked as `FAILED`? If not, how does the user know if their error handler is broken if we failed to deploy it?
- If a user updates an error handler, and it fails to deploy, is the containing Bridge instance marked as `FAILED`?

In particular, that last scenario doesn't feel correct to me. I don't think we'd want to mark a bridge as `FAILED` if the user does a bad update to the error handler.

It feels like the error handler needs to be its own resource, perhaps at `bridges/{id}/error_handler` so that we can separate the lifecycle of the error_handler from the bridge instance on which is it deployed. That would mean that the user can determine if the error handler is working (or not) and update it without impacting the status of the Bridge e.g making a bad update to the error handler doesn't cause a bridge to go into a `FAILED` status.

manstis: Hi @robpblake well...... 

The original Error Handler proposal had the `Processors` listed a long with every other `SINK` or `SOURCE` `Processor` however "we" deemed it unsuitable for Users to be aware of internal implementation details and hence hid Error Handler `Processor`'s from view... 

That was OK when you could only create them.

Now we need to support updating the Error Handler definition (or even removing or adding it entirely) we have to prevent the operation if the underlying `Processor` is not `READY` otherwise we run into all sorts of race condition issues. 

This PR, whilst supporting updating the Error Handler, does change the life-cycle to exactly as you fear.

An Error Handler `Processor` is now "just" a dependency of the `Bridge`... hence, as with other dependencies (i.e. RHOC and `Processor`'s), if provisioning of the dependency fails the parent is moved into a `FAILED` state too. Meaning both of your horror scenarios are true.

It sounds like we need a meeting (you, @danielezonca, @kelvah, me and anyone else interest....)

manager/src/main/java/com/redhat/service/smartevents/manager/BridgesServiceImpl.java
```diff
@@ -90,6 +91,53 @@ public Bridge createBridge(String customerId, String organisationId, String owne
...
+
+        // Validate update.
+        // Name cannot be updated.
+        if (!Objects.equals(existingBridge.getName(), bridgeRequest.getName())) {
+            throw new BadRequestException("It is not possible to update the Bridge's name.");
+        }
+
+        // Construct updated definition
+        Action updatedErrorHandler = bridgeRequest.getErrorHandler();
+        BridgeDefinition updatedDefinition = new BridgeDefinition(Objects.nonNull(updatedErrorHandler) ? updatedErrorHandler : null);
```
sutaakar, line 116: `Objects.nonNull(updatedErrorHandler) ? updatedErrorHandler : null`
AFAIK that is the same as using `updatedErrorHandler` directly, right? Or is there some specific meaning for this construct?

manager/src/main/java/com/redhat/service/smartevents/manager/BridgesServiceImpl.java
```diff
@@ -90,6 +91,53 @@ public Bridge createBridge(String customerId, String organisationId, String owne
...
+
+        // Create new definition copying existing properties
+        existingBridge.setModifiedAt(ZonedDateTime.now());
+        existingBridge.setStatus(ManagedResourceStatus.ACCEPTED);
+        existingBridge.setDependencyStatus(ManagedResourceStatus.ACCEPTED);
+        existingBridge.setDefinition(updatedDefinition);
+        existingBridge.setGeneration(existingBridge.getGeneration() + 1);
+
+        // Bridge and Work should always be created in the same transaction
+        workManager.schedule(existingBridge);
```
sutaakar, line 131: AFAIK the work handling should be synchronized with https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/920 (adjusting the PR once the other one is merged)

manstis: Correct. I would suggest #920 be merged first.

The _approach_ to editing Bridges is still not agreed across the team.

I _thought_ we was going to have agreement on Friday but it never came to be.

I'll move this back to draft.

manager/src/main/java/com/redhat/service/smartevents/manager/BridgesServiceImpl.java
manstis: hahaha, yes it is.. IDK what I was thinking. Changed.

manstis:
I have repeated my earlier tests of the following scenarios locally:

- Create a Bridge without Error Handler - OK
- Add an Error Handler to the Bridge - OK
- Update the Error Handler on the Bridge - OK
- Delete the Error Handler from the Bridge - OK
- Delete the Bridge - OK
- Create a Bridge with an Error Handler - OK
- Update the Error Handler on the Bridge - OK
- Delete the Bridge (and the Error Handler is deleted automatically) - OK

manstis:
@robpblake @danielezonca @sutaakar This PR is ready for review now.

sutaakar (approved):
Looks good to me

manstis:
**NOTE TO SELF** When https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/899 is merged this PR needs to be updated to prevent the Cloud Provider and Region being updated on a Bridge update...

manager/src/main/java/com/redhat/service/smartevents/manager/BridgesServiceImpl.java
```diff
@@ -90,6 +91,53 @@ public Bridge createBridge(String customerId, String organisationId, String owne
...
+        if (!Objects.equals(existingBridge.getName(), bridgeRequest.getName())) {
+            throw new BadRequestException("It is not possible to update the Bridge's name.");
+        }
+
+        // Construct updated definition
+        Action updatedErrorHandler = bridgeRequest.getErrorHandler();
+        BridgeDefinition updatedDefinition = new BridgeDefinition(updatedErrorHandler);
+
+        // No need to update CRD if the definition is unchanged
+        if (existingDefinition.equals(updatedDefinition)) {
```
jomarko, line 119: Could we please use `Objects.equals()` ? reading this class, not sure `existingDefinition` is `!= null`.

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
```diff
@@ -103,10 +102,10 @@ public Processor createErrorHandlerProcessor(String bridgeId, String customerId,
             throw new InternalPlatformException("Unable to configure error handler");
         }
 
-        return doCreateProcessor(bridge, customerId, owner, ProcessorType.ERROR_HANDLER, processorRequest);
+        return doCreateProcessor(bridge, customerId, owner, ProcessorType.ERROR_HANDLER, processorRequest, bridge.getGeneration());
     }
 
-    private Processor doCreateProcessor(Bridge bridge, String customerId, String owner, ProcessorType processorType, ProcessorRequest processorRequest) {
+    private Processor doCreateProcessor(Bridge bridge, String customerId, String owner, ProcessorType processorType, ProcessorRequest processorRequest, long generation) {
```
jomarko, line 108: wouldn't be more suitable to have some self explaining enum `GenerationStrategy long` rather than `long generation`?

manager/src/test/java/com/redhat/service/smartevents/manager/api/user/BridgesAPITest.java
```diff
@@ -331,6 +334,23 @@ public void testDeleteBridgeWithActiveProcessors() {
...
+                new BridgeDTO(bridgeResponse.getId(),
+                        bridgeResponse.getName(),
+                        bridgeResponse.getEndpoint(),
+                        DEFAULT_CUSTOMER_ID,
+                        DEFAULT_USER_NAME,
+                        READY,
+                        new KafkaConnectionDTO()));
+
+        TestUtils.deleteBridge(bridgeResponse.getId()).then().statusCode(202);
+    }
```
jomarko, line 352: Is there also an error handler state that should cause bridge deletion to fail? if yes, should we cover it by test?

jomarko (commented):
Thank you @manstis . See my comments inline and https://github.com/manstis/sandbox/pull/5/files

manager/src/main/java/com/redhat/service/smartevents/manager/BridgesServiceImpl.java
manstis: Done. I forgot your fondness of `Objects.equals(..)` (and I do not disagree, it is a good idea).

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
manstis: No, it needs to be a numerical type.

`generation` increases each time a `Bridge` is [saved](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/875/files#diff-66557df441afc3933a43c346564ae30380002fb47b8a0f12612ebcf2e71ece79R128). When updating a `Bridge` instance in` BridgeWorker` if the Error Handler `Processor` generation [is not equal to](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/875/files#diff-1319ec61fecd04bc61a3e44b2bffba77efe39a0885cf8d585aa9b01b97736859R122) the that of the `Bridge` it means the `Processor` is out of date and needs updating too. We then [set](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/875/files#diff-58ef592841ec60f16af57955f1413b4962e3111a85558c21e9f346bbf10ae4f9R122) the Error Handler `Processor` generation to be the same as that for the `Bridge.

jomarko: Ok, then it is fair to no use enum. However I do not see any `+ 1` mechanism in `BridgeWorker` as you mentioned in `ConnectorWorker`

manager/src/test/java/com/redhat/service/smartevents/manager/api/user/BridgesAPITest.java
manstis: New test added.

jomarko (approved):

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
manstis: That's a great question!

I've tried to answer it at least three times; but I am failing! I recall I was setting the Error Handler generation to equal that of the `Bridge` to avoid unnecessary updates that I saw happening when testing; but I cannot remember the exact circumstances. It also leads me to then believe the same should be done in `ConnectorWorker` (set to that of the `Processor`) and not the increment as it does now. I am also concerned setting a `Processor` generation to that of the `Bridge` is wrong for regular processors not _tied_ to the life-cycle of the `Bridge`.

I need to investigate more..

manstis: https://issues.redhat.com/browse/MGDOBR-975

manstis:
@danielezonca @robpblake This is good to review/merge now.

I've added the checks to prevent a `Bridge` `CloudProvider` and/or `Region` from being updated.

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorService.java
```diff
@@ -15,9 +15,11 @@ public interface ProcessorService {
 
     Processor createProcessor(String bridgeId, String customerId, String owner, ProcessorRequest processorRequest);
 
+    Processor updateProcessor(String bridgeId, String processorId, String customerId, ProcessorRequest processorRequest);
+
     Processor createErrorHandlerProcessor(String bridgeId, String customerId, String owner, ProcessorRequest processorRequest);
 
-    Processor updateProcessor(String bridgeId, String processorId, String customerId, ProcessorRequest processorRequest);
+    Processor updateErrorHandlerProcessor(String bridgeId, String processorId, String customerId, ProcessorRequest processorRequest);
```
robpblake, line 23: @manstis Does it simplify our life and make things clearer if we have a `Optional<Processor> getErrorHandler(String bridgeId)` method?  It seems a little confusing to me (I'm easily confused I appreciate) that in some places we're very explicit e.g create/update Error handler, but in other's we rely on knowledge of us understanding that the list of hidden processors exists and may contain the error handler.

It existed that way before this PR, so wonder if that's a follow up change we want to make if you think it sounds sensible.

cheers,

Rob

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/BridgeWorker.java
```diff
@@ -72,41 +76,61 @@ public Bridge createDependencies(Work work, Bridge bridge) {
...
 
-        String bridgeId = bridge.getId();
-        String customerId = bridge.getCustomerId();
-        ListResult<Processor> processors = processorService.getHiddenProcessors(bridgeId, customerId);
-
+        // If an Error Handler processor exists assume it is to be updated otherwise create it!
         // This assumes we can only have one ErrorHandler Processor per Bridge
-        if (processors.getTotal() > 0) {
-            return;
+        if (hiddenProcessors.getTotal() > 0) {
```
robpblake, line 114: @manstis We could extract this into a `getErrorHandler` method which returns the `Optional` as we have similar logic in the delete flow. Or see my previous comment about providing a `getErrorHandler` method on the `ProcessorService` which feels perhaps more appropriate.

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/BridgeWorker.java
```diff
@@ -72,41 +76,61 @@ public Bridge createDependencies(Work work, Bridge bridge) {
...
                 RhoasTopicAccessType.CONSUMER_AND_PRODUCER);
 
-        // We don't need to wait for the Bridge to be READY to create the Error Handler.
-        createErrorHandlerProcessor(bridge);
+        // We don't need to wait for the Bridge to be READY to handle the Error Handler.
+        createOrUpdateOrDeleteErrorHandlerProcessor(bridge);
+
+        // Keep the Bridge in PREPARING until the Error Handler processor is READY.
+        // This ensures Users cannot update the Bridge Error Handler until it is available.
+        if (isErrorHandlerProcessorReady(bridge)) {
```
robpblake, line 84: @manstis Checking my understanding here: The `Bridge` will enter the `FAILED` state only after exceeding the time limit for the provisioning process if the ErrorHandler fails to deploy?

If the `ErrorHandler` has failed to deploy, shouldn't we immediately mark the `Bridge` as failed within this loop as well?

cheers,

Rob

robpblake (commented):
@manstis Couple of questions/comments from me. Thanks very much!

cheers,
Rob

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
```diff
@@ -108,10 +107,10 @@ public Processor createErrorHandlerProcessor(String bridgeId, String customerId,
             throw new InternalPlatformException("Unable to configure error handler");
```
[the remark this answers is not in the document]
manstis: Updated.

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorService.java
manstis: Done. It does avoid a duplication of code in `BridgeWorker`.

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/BridgeWorker.java
manstis: Done. It does avoid a duplication of code in `BridgeWorker`.

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/BridgeWorker.java
manstis: Well........ you have to go and open a whoop-ar5e can of worms; don't you :laughing: 

The code here would put a `Bridge`'s dependencies into `READY` so that the `Bridge` itself would continue to be re-provisioned by the Operator and the `Bridge` move into `READY` itself. This would happen if the Error Handler was either successfully updated or failed leading to a scenario where the `Bridge` is `READY` but Error Handler `FAILED`. 

Your comment suggests the preferred scenario is that, if the Error Handler fails we short-circuit and fail the `Bridge` too.

I've updated this PR to do the same. It is definitely consistent with parent-child dependencies elsewhere.

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorService.java
```diff
@@ -15,19 +16,21 @@ public interface ProcessorService {
...
     List<Processor> findByShardIdWithReadyDependencies(String shardId);
 
     Processor updateProcessorStatus(ProcessorDTO processorDTO);
 
     Long getProcessorsCount(String bridgeId, String customerId);
 
-    ListResult<Processor> getUserVisibleProcessors(String bridgeId, String customerId, QueryProcessorResourceInfo queryInfo);
-
-    ListResult<Processor> getHiddenProcessors(String bridgeId, String customerId);
+    ListResult<Processor> getProcessors(String bridgeId, String customerId, QueryProcessorResourceInfo queryInfo);
```
danielezonca, line 33: I like the `UserVisible`/`Hidden` naming to make clear that there are hidden processors, now this is not visible anymore but "implicit"

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/BridgeWorker.java
robpblake: @manstis That was my understanding of the outcome of the discussions with the other week at the F2F. I thought that we had agreed:

- Failure to deploy the Error Handler would result in the `Bridge` additionally entering the `FAILED` status, as the user has no direct visibility on the Error Handler
- The work you are doing to add a `status_message` to the `Bridge` resource in #968  would provide the user with a bit more information as to why the `Bridge` is in `FAILED` status.

Do you feel anything I'm suggesting here is not in alignment with those discussions?

cheers,

Rob

danielezonca (approved):

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorService.java
manstis: @danielezonca We now only have `ListResult<Processor> getProcessors(..)` (all `SINK` and `SOURCE`) and `Optional<Processor> getErrorHandler(..)`. Everywhere we were using `getHiddenProcessors(..)` we examining the list for the first `ERROR_HANDLER`. We only support defining one Error Handler at the moment so I think this change is more aligned.

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/BridgeWorker.java
manstis: No, no @robpblake I agree and have made the change in this PR.

robpblake (approved):

PR #917, opened 2022-07-05
WIP: Update knative 1.2

matzew:
Updating to the new Knative 1.2 version (backed by the Serverless Operator 1.23)

I guess I have no JIRA, but I can create on.

* Updates to Eventing Core, from 1.1 to 1.2 but just core (no IMC /MT Broker), sources from [here](https://github.com/openshift/knative-eventing/blob/release-v1.2/openshift/release/artifacts/knative-eventing-core.yaml)
* Updates to KnativeKafka from 1.1 to 1.2, sources from [here](https://github.com/openshift-knative/eventing-kafka-broker/blob/release-v1.2/openshift/release/artifacts/eventing-kafka.yaml)
* Make use of the proper `external.topic` annotation in the code


matzew:
Are you somewhere pointing to the midstream branch of the serverless operator? If so, that should change to point to the 1.23 branch

r00ta:
> Are you somewhere pointing to the midstream branch of the serverless operator? If so, that should change to point to the 1.23 branch

We install the serverless operator on the `dev` and `stable` cluster manually

matzew:
Currently the "Knative Manifests" are manually copied over. Do you plan to at least get some "download" script that downloads the bits and places them to the `knative/resources` folder?

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/providers/GlobalConfigurationsConstants.java
```diff
@@ -25,6 +25,6 @@ public class GlobalConfigurationsConstants {
     public static final String KNATIVE_KAFKA_TOPIC_SECRET_REF_NAME_CONFIGMAP = "auth.secret.ref.name";
     public static final String KNATIVE_KAFKA_TOPIC_TOPIC_NAME_CONFIGMAP = "topic.name";
 
-    public static final String KNATIVE_BROKER_EXTERNAL_TOPIC_ANNOTATION_NAME = "x-kafka.eventing.knative.dev/external.topic";
+    public static final String KNATIVE_BROKER_EXTERNAL_TOPIC_ANNOTATION_NAME = "kafka.eventing.knative.dev/external.topic";
```
matzew, line 28: @r00ta I think I got all those occurrence of that annotation, but worth to double check

matzew:
IMO this PR gives an impression on how to move forward

r00ta:
@matzew I think we are good to go with this, wdyt?

r00ta (approved):

dev/bin/knative/resources/knative-eventing.yaml
```diff
@@ -3275,12 +3279,12 @@ metadata:
   labels:
     eventing.knative.dev/release: devel
   annotations:
-    knative.dev/example-checksum: "f8e5a744"
+    knative.dev/example-checksum: "9185c153"
```
jiripetrlik, line 3282: Is this hash mandatory? It seems a little bit suspicious to me to add some hashes there. Can you please explain.

dev/bin/knative/resources/knative-kafka.yaml
```diff
@@ -1,25 +1,54 @@
 ---
 ---
+
+# Copyright 2020 The Knative Authors
```
jiripetrlik, line 4: Is it some legal requirement to have the license headers here? It seems to me that it is added in various places and the header is inconsistent. The year is different etc. If necessary would it be possible to have only one license for one file?

dev/bin/knative/resources/knative-kafka.yaml
```diff
@@ -1,25 +1,54 @@
...
 metadata:
   name: kafka-broker-config
   namespace: knative-eventing
   labels:
-    kafka.eventing.knative.dev/release: v1.1.3
+    kafka.eventing.knative.dev/release: v1.2.3
 data:
-  default.topic.partitions: "1"
-  default.topic.replication.factor: "1"
+  default.topic.partitions: "10"
```
jiripetrlik, line 26: Can you please explain why we need more partitions?

dev/bin/knative/resources/knative-kafka.yaml
```diff
@@ -255,33 +330,432 @@ rules:
...
   labels:
-    kafka.eventing.knative.dev/release: v1.1.3
+    kafka.eventing.knative.dev/release: v1.2.3
   name: config-kafka-leader-election
   namespace: knative-eventing
   annotations:
     knative.dev/example-checksum: "96896b00"
 data:
   _example: |
+    ################################
```
jiripetrlik, line 638: I expect that we do not want to provide any examples how to configure knative. Can you please remove this?

dev/bin/knative/resources/knative-kafka.yaml
```diff
@@ -255,33 +330,432 @@ rules:
...
+metadata:
+  name: config-kafka-source-defaults
+  namespace: knative-eventing
+  labels:
+    kafka.eventing.knative.dev/release: v1.2.3
+  annotations:
+    knative.dev/example-checksum: "b6ed351d"
+data:
+  _example: |
+    ################################
```
jiripetrlik, line 702: Can you please remove this example?

dev/bin/knative/resources/knative-eventing.yaml
r00ta: @jiripetrlik this file is copied from here https://github.com/openshift-knative/eventing-kafka-broker/blob/release-v1.2/openshift/release/artifacts/eventing-kafka.yaml . It's just the standard knative installation. 

Was there a specific reason not to use directly the link https://github.com/openshift-knative/eventing-kafka-broker/blob/release-v1.2/openshift/release/artifacts/eventing-kafka.yaml in the installation script? At the end the link points to a release, I think it should be fine just to keep this reference to a fixed version. wdyt @matzew ?

dev/bin/knative/resources/knative-kafka.yaml
matzew: To be clear all upstream released manifests have these headers. Here they are basically copied from the upstream. I don't think there is any issue with this, but I can be wrong about implications.

If you want to remove feel free to a custom script to remove them.

dev/bin/knative/resources/knative-kafka.yaml
matzew: We, Serverless (and Knative) have these values as default. Since long time

I think initially it was just `1`, but that does not really make sense... Going with different values is NOT recommended (partion of `1` is generally just a toy)

dev/bin/knative/resources/knative-kafka.yaml
matzew: This is part of the upstream manifests, Feel free to remove them, if you don't need them.

I'd recommend against developing some script that removes this (in fact the actual serverless opeartor ships manifests w/ exactly the same). There is no harm on this

dev/bin/knative/resources/knative-kafka.yaml
matzew: Product does have this too, so there is no reason to change these scripts here, on the DEV folder, IMO

jiripetrlik (approved):
@matzew 
Hello, thank you for explanations. Looks good to me.

PR #929, opened 2022-07-08
[DRAFT] MGDOBR-441: Shard may never delete resources if it crashes after sending DELETING status call

manstis:
**=== DRAFT ===** 

This is to gather opinions before I spend much time writing tests.

[https://issues.redhat.com/browse/MGDOBR-441](https://issues.redhat.com/browse/MGDOBR-441)

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>

### DIFF
--- a/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
+++ b/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
@@ -97,6 +97,8 @@ public class BridgeExecutorServiceImpl implements BridgeExecutorService {
             return true;
         }
         // Does exist, but it has already entered the Kubernetes reconciliation loop. Don't update it.
+        // The expected definition will always have READY:Unknown set by BridgeExecutor.Builder.build().
+        // READY:False is set by the first iteration through k8s's reconciliation loop.
         if (existing.getStatus()
                 .getConditionByType(ConditionTypeConstants.READY)
                 .filter(c -> c.getStatus() == ConditionStatus.False).isPresent()) {
--- a/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
+++ b/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
@@ -22,6 +22,8 @@ import com.redhat.service.smartevents.shard.operator.providers.GlobalConfigurati
 import com.redhat.service.smartevents.shard.operator.providers.TemplateImportConfig;
 import com.redhat.service.smartevents.shard.operator.providers.TemplateProvider;
 import com.redhat.service.smartevents.shard.operator.resources.BridgeExecutor;
+import com.redhat.service.smartevents.shard.operator.resources.ConditionStatus;
+import com.redhat.service.smartevents.shard.operator.resources.ConditionTypeConstants;
 import com.redhat.service.smartevents.shard.operator.utils.Constants;
 import com.redhat.service.smartevents.shard.operator.utils.DeploymentSpecUtils;
 import com.redhat.service.smartevents.shard.operator.utils.LabelsBuilder;
@@ -78,7 +80,7 @@ public class BridgeExecutorServiceImpl implements BridgeExecutorService {
                 .withName(BridgeExecutor.resolveResourceName(processorDTO.getId()))
                 .get();
 
-        if (existing == null || !expected.getSpec().equals(existing.getSpec())) {
+        if (createOrReplace(expected, existing)) {
             BridgeExecutor bridgeExecutor = kubernetesClient
                     .resources(BridgeExecutor.class)
                     .inNamespace(namespace.getMetadata().getName())
@@ -87,6 +89,21 @@ public class BridgeExecutorServiceImpl implements BridgeExecutorService {
             // create or update the secrets for the bridgeExecutor
             createOrUpdateBridgeExecutorSecret(bridgeExecutor, processorDTO);
         }
+    }
+
+    private boolean createOrReplace(BridgeExecutor expected, BridgeExecutor existing) {
+        // Does not exist. Create it.
+        if (existing == null) {
+            return true;
+        }
+        // Does exist, but it has already entered the Kubernetes reconciliation loop. Don't update it.
+        if (existing.getStatus()
+                .getConditionByType(ConditionTypeConstants.READY)
+                .filter(c -> c.getStatus() == ConditionStatus.False).isPresent()) {
+            return false;
+        }
+        // Specs differ. Update it.
+        return !expected.getSpec().equals(existing.getSpec());
     }
 
     @Override
--- a/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
+++ b/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
@@ -94,6 +94,7 @@ public class BridgeExecutorServiceImpl implements BridgeExecutorService {
     private boolean createOrReplace(BridgeExecutor expected, BridgeExecutor existing) {
         // Does not exist. Create it.
         if (existing == null) {
+            LOGGER.info("Existing BridgeExecutor does not exist. Creating a new instance.");
             return true;
         }
         // Does exist, but it has already entered the Kubernetes reconciliation loop. Don't update it.
@@ -102,6 +103,7 @@ public class BridgeExecutorServiceImpl implements BridgeExecutorService {
         if (existing.getStatus()
                 .getConditionByType(ConditionTypeConstants.READY)
                 .filter(c -> c.getStatus() == ConditionStatus.False).isPresent()) {
+            LOGGER.info("Existing BridgeExecutor has been marked as READY:False. Skipping.");
             return false;
         }
         // Specs differ. Update it.
--- a/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeIngressServiceImpl.java
+++ b/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeIngressServiceImpl.java
@@ -79,6 +79,7 @@ public class BridgeIngressServiceImpl implements BridgeIngressService {
     private boolean createOrReplace(BridgeIngress expected, BridgeIngress existing) {
         // Does not exist. Create it.
         if (existing == null) {
+            LOGGER.info("Existing BridgeIngress does not exist. Creating a new instance.");
             return true;
         }
         // Does exist, but it has already entered the Kubernetes reconciliation loop. Don't update it.
@@ -87,6 +88,7 @@ public class BridgeIngressServiceImpl implements BridgeIngressService {
         if (existing.getStatus()
                 .getConditionByType(ConditionTypeConstants.READY)
                 .filter(c -> c.getStatus() == ConditionStatus.False).isPresent()) {
+            LOGGER.info("Existing BridgeIngress has been marked as READY:False. Skipping.");
             return false;
         }
         // Specs differ. Update it.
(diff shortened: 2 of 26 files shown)


manstis:
Hi @robpblake @r00ta you're both probably the most knowledgeable in this area so I'll ask your opinions.

The issue is that the Operator sends a status update to the Manager that "something" is _going_ to happen before the "something" has actually completed. This is nice (and solved another issue) but means if the Operator dies before the "something" has completed it is never picked up again and completed when the Operator comes back to life.

This PR now also returns PROVISIONING/DELETING resources to the Operator (i.e. the "something" that may not have completed). I then check whether the status of the CRD is READY i.e. fully reconciled before asking Kubernetes to do its thing. If the CRD is not READY it _suggests_ a reconciliation is already in progress which _should_ prevent the same CRD being worked on by multiple invocations of our Operator<->Manager polling loop.

WDYT (before I write loads of tests to be told my approach stinks)?

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
```diff
@@ -89,6 +91,25 @@ public void createBridgeExecutor(ProcessorDTO processorDTO) {
...
+        }
+        // Does exist, but it has already entered the Kubernetes reconciliation loop. Don't update it.
+        // The expected definition will always have READY:Unknown set by BridgeExecutor.Builder.build().
+        // READY:False is set by the first iteration through k8s's reconciliation loop.
+        if (existing.getStatus()
+                .getConditionByType(ConditionTypeConstants.READY)
+                .filter(c -> c.getStatus() == ConditionStatus.False).isPresent()) {
+            LOGGER.info("Existing BridgeExecutor has been marked as READY:False. Skipping.");
+            return false;
+        }
```
r00ta, line 108: I'm not sure I get this piece here. If the condition status is not ready **but** the specs are different I think we should update the CRD. In other words I think we should remove these lines, or am I missing anything?

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
```diff
@@ -78,7 +80,7 @@ public void createBridgeExecutor(ProcessorDTO processorDTO) {
                 .withName(BridgeExecutor.resolveResourceName(processorDTO.getId()))
                 .get();
 
-        if (existing == null || !expected.getSpec().equals(existing.getSpec())) {
+        if (createOrReplace(expected, existing)) {
```
r00ta, line 83: nitpick: `createOrReplace` is not very informative here. Maybe `isExecutorEqual`?

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeIngressServiceImpl.java
```diff
@@ -63,7 +65,7 @@ public void createBridgeIngress(BridgeDTO bridgeDTO) {
                 .withName(BridgeIngress.resolveResourceName(bridgeDTO.getId()))
                 .get();
 
-        if (existing == null || !expected.getSpec().equals(existing.getSpec())) {
+        if (createOrReplace(expected, existing)) {
```
r00ta, line 68: same, `isIngressEqual`?

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeIngressServiceImpl.java
```diff
@@ -74,6 +76,25 @@ public void createBridgeIngress(BridgeDTO bridgeDTO) {
...
+        }
+        // Does exist, but it has already entered the Kubernetes reconciliation loop. Don't update it.
+        // The expected definition will always have READY:Unknown set by BridgeIngress.Builder.build().
+        // READY:False is set by the first iteration through k8s's reconciliation loop.
+        if (existing.getStatus()
+                .getConditionByType(ConditionTypeConstants.READY)
+                .filter(c -> c.getStatus() == ConditionStatus.False).isPresent()) {
+            LOGGER.info("Existing BridgeIngress has been marked as READY:False. Skipping.");
+            return false;
+        }
```
r00ta, line 93: same

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/ManagerSyncServiceImpl.java
```diff
@@ -62,6 +62,7 @@ protected Uni<Object> doBridges() {
         return managerClient.fetchBridgesToDeployOrDelete()
                 .onItem().transformToUni(x -> Uni.createFrom().item(x.stream().map(y -> {
                     if (y.getStatus().equals(ManagedResourceStatus.PREPARING)) { // Bridges to deploy
+                        LOGGER.info("Found Bridge in PREPARING state. Moving to PROVISIONING.");
```
r00ta, line 65: Add the `bridgeId` in the log statement? (same in all the other logs)

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/ManagerSyncServiceImpl.java
```diff
@@ -62,6 +62,7 @@ protected Uni<Object> doBridges() {
         return managerClient.fetchBridgesToDeployOrDelete()
                 .onItem().transformToUni(x -> Uni.createFrom().item(x.stream().map(y -> {
                     if (y.getStatus().equals(ManagedResourceStatus.PREPARING)) { // Bridges to deploy
```
r00ta, line 64: what is we add here `if (y.getStatus().equals(ManagedResourceStatus.PREPARING) || y.getStatus().equals(ManagedResourceStatus.PROVISIONING))`? 

I know we would send another `PROVISIONING` update to the manager but I think it would be anyway more consistent

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeExecutorController.java
```diff
@@ -76,12 +78,32 @@ public List<EventSource> prepareEventSources(EventSourceContext<BridgeExecutor>
...
+        if (status.getConditionByType(ConditionTypeConstants.READY)
+                .filter(c -> c.getStatus() != ConditionStatus.False)
+                .isPresent()) {
+            LOGGER.info("Marking bridgeExecutor '{}' in namespace '{}' as augmenting reconciliation.",
+                    bridgeExecutor.getMetadata().getName(),
+                    bridgeExecutor.getMetadata().getNamespace());
+            bridgeExecutor.getStatus().markConditionFalse(ConditionTypeConstants.READY);
+            bridgeExecutor.getStatus().markConditionTrue(ConditionTypeConstants.AUGMENTATION, ConditionReasonConstants.RECONCILIATION_PROGRESSING);
+            return UpdateControl.updateStatus(bridgeExecutor);
+        }
```
r00ta, line 101: Why do we need this?

r00ta (commented):
I know it's a draft, I was just surfing around

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
manstis: Hi @r00ta I'll answer here; but it is probably a general response to your other questions too.

The issue is this: When the `operator` polls the `manager` it receives a `Bridge`/`Processor` that is in the `PREPARED` state. The `operator` then immediately signals to the `manager` that the resources is `PROVISIONING`. If the `operator` dies at this point we are stuck with a resource flagged as `PROVISIONING` that is no longer detected by the `operator` polling the `manager`. The resource never gets completed.

My change now also returns resources from the `manager` that are in `PROVISIONING`. However this has the knock on affect that if Resource X is already being provisioned by the `operator` when a subsequent poll completes the `operator` has two Resource X's to provision; one "in flight" and the other fresh from the subsequent poll.

What do I do!?!

When a new resource is being provisioned the `existing` is `null` and the `expected` has a Status of `Ready:Unknown`. This causes us to invoke request creation and our Controller reconciliation loop starts provisioning the resource. As the first step in the loop I then [set](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/2ef7e7530e8b5cf0ccdebc1dbf66479582cf69c3/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeIngressController.java#L91-L99) the Status to `Ready:False`. 

Should the `operator` die at this point:-
- When it comes back up k8s continues with the reconciliation loop.
- If a poll completes from the `operator`->`manager` the `existing` has a Status `Ready:False` and the check above fails and consequentially we do not signal to k8s to provision the resource (it is already being handled by k8s recovery in the previous bullet point).

Should the operator not die but a subsequent poll completes:
- If a poll completes from the `operator`->`manager` the `existing` has a Status `Ready:False` and the check above fails and consequentially we do not signal to k8s to provision the resource (it is already being handled by k8s recovery in the previous bullet point).

So, I am using the CRD's Status to indicate whether k8s is already handling it.

`Ready:Unknown` -> We need to tell k8s and invoke the controller/reconciliation.
`Ready:False` -> We've already told k8s to handle it.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeExecutorController.java
manstis: See my comment above.

I won't reply to the other minor comments as I am more interested in your thoughts of the the high-level approach first.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
r00ta: Uhm I still don't get why this piece of code is useful. 
1) If the operator sent the PROVIISONING update but crashed before creating the CRD -> when it starts again it would fetch the resource from the manager and create it
2) If the operator sent the PROVISIONING update and created the CRD it can keep fetching the same resource, but the `createOrReplace` would not have any affect given that the spec is the same. 

Why do we need 
```java 
        if (existing.getStatus()
                .getConditionByType(ConditionTypeConstants.READY)
                .filter(c -> c.getStatus() == ConditionStatus.False).isPresent()) {
            LOGGER.info("Existing BridgeExecutor has been marked as READY:False. Skipping.");
            return false;
        }
```
?

The process should be already resilient if the manager returns the istances in the PROVISIONING state

r00ta:
```
In this scenario k8s would continue the reconciliation of (1) and we would also initiate (2) so there would be two reconciliations in process at the same time for the same resource... but please tell me if this is incorrect!!!
```
Yes, this is not correct. The operator sdk guarantees that there is always **exactly** one reconcile loop running for each resource -> the concurrency is already managed by the operator sdk, it basically keeps a queue of triggers and there is only one reconcile loop running for every resource

manstis: Thanks for the information. That changes _everything_!

manstis: Thanks for the information. That changes _everything_!

manstis:
Closing due to the [revelation](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/929#discussion_r920492964) provided by @r00ta.

I'll open a new PR in due course.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
manstis: Change no longer needed.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeIngressServiceImpl.java
manstis: Change no longer needed.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/ManagerSyncServiceImpl.java
manstis: To be honest I prefer the separation.

We won't get the duplicate (misleading) status update and the log entries are more contextually correct.

Let's wait for my revised PR to go to review and see what everybody thinks.

PR #947, opened 2022-07-14
[2d81c] Update kustomization images

r00ta:


manstis:
@r00ta sooooo.. this will _deploy_ my `MC_xxx` change to dev now?

PR #968, opened 2022-07-22
MGDOBR-52: Provide customers with a "status message" on RHOSE deployment failures

manstis:
[https://issues.redhat.com/browse/MGDOBR-52](https://issues.redhat.com/browse/MGDOBR-52)

Happy to set up a meeting to talk about when anyone feels like looking...

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>


infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorHelper.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.redhat.service.smartevents.infra.exceptions.definitions.platform.UnspecifiedProvisioningFailureException;
+
+import io.quarkus.runtime.Quarkus;
+
+@ApplicationScoped
+public class BridgeErrorHelper {
```
manstis, line 17: I could move this into `ErrorService` or rename...

It's used to map _any_ error occurring provisioning/deprovisioning to a `BridgeError` (or use a default).

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/definitions/platform/ProvisioningFailureException.java
```diff
@@ -0,0 +1,14 @@
+package com.redhat.service.smartevents.infra.exceptions.definitions.platform;
+
+import com.redhat.service.smartevents.infra.exceptions.ProvidesReason;
+
+public class ProvisioningFailureException extends InternalPlatformException implements ProvidesReason {
```
manstis, line 5: I only have one exception type for where we explicitly handle certain scenarios... I could replace with more explicit types for each scenario and delete the `ProvidesReason` interface.. but open for opinions.

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/definitions/platform/UnspecifiedProvisioningFailureException.java
```diff
@@ -0,0 +1,12 @@
+package com.redhat.service.smartevents.infra.exceptions.definitions.platform;
+
+public class UnspecifiedProvisioningFailureException extends InternalPlatformException {
```
manstis, line 3: Fallback for any generic `Exception`'s that do not map to explicit `BridgeError`'s.

infra/src/main/java/com/redhat/service/smartevents/infra/models/dto/BridgeStatusWrapperDTO.java
```diff
@@ -0,0 +1,63 @@
+package com.redhat.service.smartevents.infra.models.dto;
+
+import java.util.Objects;
+
+import javax.validation.constraints.NotNull;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.redhat.service.smartevents.infra.exceptions.BridgeError;
+
+public class BridgeStatusWrapperDTO {
```
manstis, line 10: We used to send just the `Bridge` to the `manager` from the `operator`.

I now return the `Bridge` and any `BridgeError` that led to the `operator` to fail.

infra/src/main/java/com/redhat/service/smartevents/infra/models/dto/ProcessorStatusWrapperDTO.java
```diff
@@ -0,0 +1,63 @@
+package com.redhat.service.smartevents.infra.models.dto;
+
+import java.util.Objects;
+
+import javax.validation.constraints.NotNull;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.redhat.service.smartevents.infra.exceptions.BridgeError;
+
+public class ProcessorStatusWrapperDTO {
```
manstis, line 10: We used to send just the `Processor` to the `manager` from the `operator`.

I now return the `Processor` and any `BridgeError` that led to the `operator` to fail.

manager/src/main/java/com/redhat/service/smartevents/manager/BridgesService.java
```diff
@@ -25,7 +26,7 @@ public interface BridgesService {
 
     List<Bridge> findByShardIdWithReadyDependencies(String shardId);
 
-    Bridge updateBridge(BridgeDTO bridgeDTO);
+    Bridge updateBridgeStatus(BridgeStatusWrapperDTO statusWrapperDTO);
```
manstis, line 29: Just making the method name consistent with that used for `Processor`'s (and closer to what it actually does).

manager/src/main/java/com/redhat/service/smartevents/manager/dao/WorkErrorDAO.java
```diff
@@ -0,0 +1,27 @@
...
+import javax.transaction.Transactional;
+
+import com.redhat.service.smartevents.manager.models.WorkError;
+
+import io.quarkus.hibernate.orm.panache.PanacheRepositoryBase;
+import io.quarkus.panache.common.Parameters;
+
+@Transactional
+@ApplicationScoped
+public class WorkErrorDAO implements PanacheRepositoryBase<WorkError, String> {
```
manstis, line 15: A new entity to record an error that happened when (de-)provisioning something.

manager/src/test/java/com/redhat/service/smartevents/manager/workers/resources/BridgeWorkerTest.java
```diff
@@ -115,6 +119,11 @@ void testProvisionWorkWithKnownResource(ManagedResourceStatus status,
         assertThat(refreshed.getDependencyStatus()).isEqualTo(dependencyStatusWhenComplete);
         verify(rhoasServiceMock).createTopicAndGrantAccessFor(TEST_TOPIC_NAME, RhoasTopicAccessType.CONSUMER_AND_PRODUCER);
         verify(workManager, times(isWorkComplete ? 0 : 1)).reschedule(any());
+
+        if (throwRhoasError) {
```
manstis, line 123: Woot! A single test showing records are created if the `Worker` fails..

If my approach seems reasonable I'd obviously write a lot more tests...

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeExecutorController.java
```diff
@@ -34,16 +36,18 @@
...
 import io.javaoperatorsdk.operator.api.reconciler.Reconciler;
+import io.javaoperatorsdk.operator.api.reconciler.RetryInfo;
 import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
 import io.javaoperatorsdk.operator.processing.event.source.EventSource;
 
 @ApplicationScoped
 @ControllerConfiguration(labelSelector = LabelsBuilder.RECONCILER_LABEL_SELECTOR)
 public class BridgeExecutorController implements Reconciler<BridgeExecutor>,
-        EventSourceInitializer<BridgeExecutor> {
+        EventSourceInitializer<BridgeExecutor>, ErrorStatusHandler<BridgeExecutor> {
```
manstis, line 50: [`ErrorStatusHandler`](https://javaoperatorsdk.io/docs/features#setting-error-status-after-last-retry-attempt) is called if k8s fails to provision something.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeExecutorController.java
```diff
@@ -97,12 +101,16 @@ public UpdateControl<BridgeExecutor> reconcile(BridgeExecutor bridgeExecutor, Co
             LOGGER.info("Executor deployment BridgeProcessor: '{}' in namespace '{}' is NOT ready", bridgeExecutor.getMetadata().getName(),
                     bridgeExecutor.getMetadata().getNamespace());
 
-            bridgeExecutor.getStatus().setConditionsFromDeployment(deployment);
+            setStatusFromDeployment(bridgeExecutor, deployment);
```
manstis, line 104: I moved the method `setConditionsFromDeployment(deployment)` from `CustomResourceStatus` as it was only used here.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeExecutorController.java
```diff
@@ -158,10 +162,14 @@ public DeleteControl cleanup(BridgeExecutor bridgeExecutor, Context context) {
         return DeleteControl.defaultDelete();
     }
 
-    private void notifyDeploymentFailure(BridgeExecutor bridgeExecutor, String failureReason) {
-        LOGGER.warn("Processor deployment BridgeExecutor: '{}' in namespace '{}' has failed with reason: '{}'",
-                bridgeExecutor.getMetadata().getName(), bridgeExecutor.getMetadata().getNamespace(), failureReason);
-        notifyManager(bridgeExecutor, ManagedResourceStatus.FAILED);
+    @Override
+    public Optional<BridgeExecutor> updateErrorStatus(BridgeExecutor bridgeExecutor, RetryInfo retryInfo, RuntimeException e) {
```
manstis, line 166: This is my implementation of `ErrorStatusHandler`.

If k8s failed to provision something I update the CRD status and notify the `manager.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeIngressController.java
```diff
@@ -32,16 +35,18 @@
...
 import io.javaoperatorsdk.operator.api.reconciler.Reconciler;
+import io.javaoperatorsdk.operator.api.reconciler.RetryInfo;
 import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
 import io.javaoperatorsdk.operator.processing.event.source.EventSource;
 
 @ApplicationScoped
 @ControllerConfiguration(labelSelector = LabelsBuilder.RECONCILER_LABEL_SELECTOR)
 public class BridgeIngressController implements Reconciler<BridgeIngress>,
-        EventSourceInitializer<BridgeIngress> {
+        EventSourceInitializer<BridgeIngress>, ErrorStatusHandler<BridgeIngress> {
```
manstis, line 49: Likewise for `Bridge`'s.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/resources/CustomResourceStatus.java
```diff
@@ -9,17 +9,16 @@
...
 
-import io.fabric8.kubernetes.api.model.apps.Deployment;
-import io.fabric8.kubernetes.client.internal.readiness.Readiness;
+import io.javaoperatorsdk.operator.api.ObservedGenerationAware;
 
 /**
  * Common interface for Kubernetes Custom Resource status
  */
-public abstract class CustomResourceStatus {
+public abstract class CustomResourceStatus implements ObservedGenerationAware {
```
manstis, line 18: This is considered ["best practice"](https://javaoperatorsdk.io/docs/features#automatic-observed-generation-handling) and prevents an additional reconciliation loop from running now I've added an `ErrorStatusHandler`.

"Note that this update request will also produce an event, and will result in a reconciliation **if the controller is not generation aware**"

manstis:
@r00ta As always, your opinion is welcome!

This PR adds code to _store_ errors during (de-)provisioning from either the `operator` or our `Workers` (in the `manager`).

I would extend our API to have `smartevents_mgmt/v1/bridges/{id}/status` etc to return the errors.

An empty collection meaning "OK". I've not written that bit yet.

WDYT?

infra/src/main/java/com/redhat/service/smartevents/infra/models/dto/BridgeStatusWrapperDTO.java
r00ta, line 10: What about adding the error code in the `BridgeDTO` in case of failure? In that case we might simply return the error code in the response and then it's up to the user to look up the error catalog for more details

r00ta (commented):
Hi @manstis , my 2 cents is that you already added a lot of logic in this, I would have done it slightly differently: 
1) add an `errorCode` in the `BridgeDTO` and in the `ProcessorDTO`
2) Add a column in the Bridge and Processor table
3) If the column is not null, just return it as part of the response with an `errorCode` property

It would be responsability of the user to look up for the error details using the error catalog. 

I'd avoid to wrap the BridgeDTO and the ProcessorDTO with a wrapper as the info might be already added in these POJO, as well as I'd not create a separated table to keep track of the errors as I'd expect there can be just one error for the resource at a time. 
If the feature you aim to implement is to provide the user a full history of what went wrong since the instance was created, then of course you need a dedicated table as you did.

robpblake (changes requested):
@manstis My suggestion here is for us to start simple and provide a single `status_message` field on our API resources that expose a `status` field. The purpose of that `status_message` IMHO is to provide a human readable description of why the resource is in the `status` it is. Most likely it will only be used at this point when things go into the `FAILED` status.

Some other thoughts I had when looking at this PR:

1. I think the `status_message` should be either actionable by the user or if it's something that the user can do nothing about, we should consider whether a generic message is OK. I think we should avoid leaking internal implementation details over which the user has no control.
2. The `status_message` can be set by either the Fleet Manager or the Fleet Shard, depending on where we are in the provisioning of a Resource.
3. I think we can start with the Shard optionally returning a `status_message` if it wants to. Right now I'd suggest the Fleet Manager probably just passes that message straight through to our customers, but we'll need to consider if that makes sense. Given that the API between Fleet Manager --> Shard is completely internal to the service, we can make the generation/translation of the `status_message` as elaborate as we want in the future, but I'd suggest we just start simple.

cheers,

Rob

manstis:
@r00ta @robpblake I've updated this PR as follows:-
- Single `BridgeError` can be returned from the Operator to the Manager.
- Worker mechanism can also record a single `BridgeError` in the Manager.
- `BridgeError` is wrapped into an _instance_ that has a UUID and is logged with the original exception.
- `BridgeResponse` and `ProcessorResponse` include a User friendly form of the `BridgeError.reason` including the UUID (for DevOps to use to look into the logs to debug the issue). i.e. the status is at our existing endpoints although we could change this to be `/smartevents_mgmt/v1/bridges/{bridgeId}/status` and `/smartevents_mgmt/v1/bridges/{bridgeId}/processors/{processorId}/status`.

For example:

**Mapping RuntimeException to BridgeError**
```
2022-08-01 16:30:24,952 INFO  [com.red.ser.sma.inf.exc.BridgeErrorHelper] (main) Mapping 'com.redhat.service.smartevents.infra.exceptions.definitions.platform.InternalPlatformException' to BridgeError: com.redhat.service.smartevents.infra.exceptions.definitions.platform.InternalPlatformException: error
	at com.redhat.service.smartevents.manager.RhoasServiceImpl_ClientProxy.createTopicAndGrantAccessFor(Unknown Source)
	at com.redhat.service.smartevents.manager.workers.resources.BridgeWorker.createDependencies(BridgeWorker.java:73)
	at com.redhat.service.smartevents.manager.workers.resources.BridgeWorker.createDependencies(BridgeWorker.java:28)
...
2022-08-01 16:30:24,956 INFO  [com.red.ser.sma.inf.exc.BridgeErrorHelper] (main) Mapped 'com.redhat.service.smartevents.infra.exceptions.definitions.platform.InternalPlatformException' to 'BridgeErrorInstance{bridgeError=Error [id=9, code=OPENBRIDGE-9, reason=There was an internal exception. This is supposed not to be fixable by the user, so you should open a bug with all the information you have., type=USER], uuid='3678fc66-94a3-4c23-9500-40646476e7f2'}'
```

**Response to User**
```
[OPENBRIDGE-9] There was an internal exception. This is supposed not to be fixable by the user, so you should open a bug with all the information you have. Please contact support quoting UUID: 3678fc66-94a3-4c23-9500-40646476e7f2
```
Am I million miles from what we need?

I don't want to write tests if I'm still going the wrong direction!

manager/src/main/java/com/redhat/service/smartevents/manager/BridgesServiceImpl.java
```diff
@@ -230,12 +235,19 @@ public List<Bridge> findByShardIdWithReadyDependencies(String shardId) {
...
     @Override
-    public Bridge updateBridge(BridgeDTO bridgeDTO) {
+    public Bridge updateBridgeStatus(BridgeDTO bridgeDTO) {
         Bridge bridge = getBridge(bridgeDTO.getId(), bridgeDTO.getCustomerId());
         bridge.setStatus(bridgeDTO.getStatus());
         bridge.setEndpoint(bridgeDTO.getEndpoint());
         bridge.setModifiedAt(ZonedDateTime.now(ZoneOffset.UTC));
 
+        // If an exception happened; make sure to record it.
+        BridgeErrorInstance bridgeErrorInstance = bridgeDTO.getBridgeErrorInstance();
```
manstis, line 245: Store any error information returned for the Operator.

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
```diff
@@ -304,34 +309,41 @@ public List<Processor> findByShardIdWithReadyDependencies(String shardId) {
...
         Bridge bridge = bridgesService.getBridge(processorDTO.getBridgeId());
-        Processor p = processorDAO.findById(processorDTO.getId());
-        if (p == null) {
+        Processor processor = processorDAO.findById(processorDTO.getId());
+        if (processor == null) {
             throw new ItemNotFoundException(String.format("Processor with id '%s' does not exist for Bridge '%s' for customer '%s'", bridge.getId(), bridge.getCustomerId(),
                     processorDTO.getCustomerId()));
         }
 
+        // If an exception happened; make sure to record it.
```
manstis, line 318: Store any error information returned for the Operator.

r00ta:
> **Response to User**
> 
> ```
> [OPENBRIDGE-9] There was an internal exception. This is supposed not to be fixable by the user, so you should open a bug with all the information you have. Please contact support quoting UUID: 3678fc66-94a3-4c23-9500-40646476e7f2
> ```

@manstis I'm not 100% sure about this. Originally the idea was to have 2 different kind of exceptions: `ExternalUserException` and `InternalPlatformException`. If a `ExternalUserException` is raised then it was "a user fault" -> something that the user should be able to fix without any support (for example, a wrong configuration). An `InternalPlatformException` is always exposed to the user with a very generic description (we don't expose the real cause of the issue - for example that we could not create the Kafka topic for our internal implementation). 

So I'm wondering if we should keep this logic instead of always reporting to the user a generic description with a UUID code to ask for help

manstis:
@r00ta if our code throws a `ExternalUserException` we'll show whatever is defined accordingly. If an unexpected exception is thrown (i.e. we cannot map it to a `BridgeError`) we'll show a generic message. My example of `OPENBRIDGE-9` isn't great. Let me write some better examples.

infra/src/main/resources/exception/exceptionInfo.json
```diff
@@ -196,6 +196,35 @@
...
+    "exception": "com.redhat.service.smartevents.infra.exceptions.definitions.user.UnspecifiedProvisioningFailureException",
+    "id": 36,
+    "reason": "An unspecified error occurred handling your request. Please contact the Help Desk for support quoting the date and time of this error.",
+    "type": "USER"
+  },
+  {
+    "exception": "com.redhat.service.smartevents.infra.exceptions.definitions.user.ProvisioningTimeOutException",
+    "id": 37,
+    "reason": "Your request was not complete within the configured time-out. Please contact the Help Desk for support quoting the date and time of this error.",
+    "type": "USER"
```
robpblake, line 216: @manstis I think we need some clarity on the usage of `type` here. `type` was (I believe) intended to be used as a marker for an error directly related to something the user has done, and returned synchronously from the call the API of SmartEvents.  Good examples being: You're missing a field, you have a bad template definition.

On that understanding of `type`, it doesn't feel appropriate to use it here as these are asynchronous errors and most of them may not be related to what the user has done. For example: Being unable to provision a `ReplicaSet` may be zero fault of the user, and I'm not totally convinced we want to expose that much data to them anyway.

I would personally say that at this point, all errors from the Shard are `platform` errors and the role of the Fleet Manager is to map the specific error from the Data Plane to a useable `status_message`

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorInstance.java
```diff
@@ -0,0 +1,25 @@
+package com.redhat.service.smartevents.infra.exceptions;
+
+import java.util.UUID;
+
+public class BridgeErrorInstance extends BridgeError {
+
+    private final String uuid;
```
robpblake, line 7: @manstis This doesn't feel specific to errors from the Shard IMHO. If we think there is value in something like then I think we should apply it equally to all Errors, perhaps as something like a `request_id` or `operation_id` field on the Error.

I like the idea, but I wonder if we should consider it within the scope of tracing and perhaps outside this PR so that we can apply consistently all scenarios.

Your thoughts?

cheers,
Rob

infra/src/main/resources/exception/exceptionInfo.json
```diff
@@ -196,6 +196,35 @@
...
+    "exception": "com.redhat.service.smartevents.infra.exceptions.definitions.user.ProvisioningMaxRetriesExceededException",
+    "id": 38,
+    "reason": "Your request failed after the maximum number of configured re-try attempts. Please contact the Help Desk for support quoting the date and time of this error.",
+    "type": "USER"
+  },
+  {
+    "exception": "com.redhat.service.smartevents.infra.exceptions.definitions.user.ProvisioningReplicaFailureException",
+    "id": 39,
+    "reason": "A ReplicaSet failed to be provisioned for your request. Please contact the Help Desk for support quoting the date and time of this error.",
+    "type": "USER"
```
robpblake, line 228: @manstis Could you add a task to the backlog for us to review the contents of `exceptionInfo.json`. It's grown organically and I think we need to put in place better definitions on when to add things to it.

cheers,

Rob

manager/src/main/java/com/redhat/service/smartevents/manager/models/Bridge.java
```diff
@@ -71,6 +72,12 @@ public class Bridge extends ManagedDefinedResource<BridgeDefinition> {
     @Column(name = "region", nullable = false, updatable = false)
     private String region;
 
+    @Column(name = "bridge_error_id")
```
robpblake, line 75: @manstis I personally would just store the `status_message` directly on the object here.

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/AbstractWorker.java
```diff
@@ -104,6 +130,10 @@ protected T persist(T managedResource) {
         return getDao().getEntityManager().merge(managedResource);
     }
 
+    protected void recordError(Work work, Exception e) {
+        // Do nothing by default
```
robpblake, line 134: @manstis Is this is a left-over?

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/AbstractWorker.java
```diff
@@ -57,9 +63,25 @@ public T handleWork(Work work) {
...
-        if (areRetriesExceeded(work, managedResource) || isTimeoutExceeded(work, managedResource)) {
+        boolean areRetriesExceeded = areRetriesExceeded(work, managedResource);
+        boolean isTimeoutExceeded = isTimeoutExceeded(work, managedResource);
+        if (areRetriesExceeded || isTimeoutExceeded) {
             managedResource.setStatus(FAILED);
             persist(managedResource);
+
+            ProvisioningFailureException failure;
+            if (areRetriesExceeded) {
+                failure = new ProvisioningMaxRetriesExceededException(String.format(RETRIES_FAILURE_MESSAGE,
```
robpblake, line 74: @manstis If the `status_message` was just a string on the resource, we could just `r.setStatusMessage` here.

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/BridgeWorker.java
```diff
@@ -200,4 +213,29 @@ protected boolean isDeprovisioningComplete(Bridge managedResource) {
...
+        BridgeErrorInstance bridgeErrorInstance = bridgeErrorHelper.getBridgeErrorInstance(e);
+        bridge.setBridgeErrorId(bridgeErrorInstance.getId());
+        bridge.setBridgeErrorUUID(bridgeErrorInstance.getUUID());
+        persist(bridge);
+    }
+
+    // The ErrorHandler Processor is a special type. It is not visible to Users and hence
+    // should it fail to be provisioned (or de-provisioned) the User is unaware of the reason.
+    // Therefore, propagate the error associated with the Error Handler Processor to the Bridge.
+    protected void propagateProcessorError(Bridge bridge) {
```
robpblake, line 230: @manstis If we think about what originally started this discussion e.g. "user doesn't know that their error handler has failed to provision", do we feel this is actually helping?

Is it easier just to have a `bridge.setStatusMessage("Failed to provision the Error Handler");`?

robpblake (commented):
@manstis I think this is broadly the right direction, so thank you for getting us started.

I've added some comments to the PR that capture my thoughts as I was reviewing it. I'd love to hear your thoughts.

cheers,

Rob

infra/src/main/resources/exception/exceptionInfo.json
```diff
@@ -196,6 +196,35 @@
     "id": 34,
     "reason": "The selected Region is not valid.",
     "type": "USER"
+  },
+  {
+    "exception": "com.redhat.service.smartevents.infra.exceptions.definitions.user.ProvisioningFailureException",
```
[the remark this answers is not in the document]
manstis: The problem splitting RHOSE's global error catalog is the increased risk of duplicating codes (unless we agree to prefix Shard based ones with something like `OPENBRIDGE-OP-xxx` and Fleet Manager ones `OPENBRIDGE-FM-xx` (or simply leave as `OPENBRIDGE-xx`). However we'd then also need to maintain the Shard codes in the Manager to be able to map from one to the other. IMO, it all becomes a maintenance nightmare.

This PR is more trying to convey a _mechanism_ more than _specific examples_. The `ProvisioningXXX` errors I have created merely reflect "failed" scenarios we already have in the Operator and, in order for them to propagate back to the Manager, I have made them "fit" the mechanism. We should identify any specific failed scenarios we want to communicate back to the User and surround the applicable code with `try {...} catch(e)` blocks that notify the Manager explicitly. Anything not explicitly caught would be handled as a more generic error notification to the Manager.

infra/src/main/resources/exception/exceptionInfo.json
manstis: I had originally had them all extend `InternalPlatformException` as I agree; it makes more sense.

When error codes are expanded to messages I lookup the `reason` from the applicable `BridgeError`. `ErrorService` only supports looking up `USER` errors hence my change (to extend `ExternalUserException`). I can change back to `InternalPlatformException` and add a new method.

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorInstance.java
manstis: It's not used specifically for errors from the Shard; it's also used for errors from our [`Worker`'s](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/968/files#diff-1319ec61fecd04bc61a3e44b2bffba77efe39a0885cf8d585aa9b01b97736859R218-R225) in the Manager. It is used to help identify a specific occurrence of an error in the server logs. User A may perform an operation that causes Error X; so may User B or C. The _instance_ class contains a UUID to identify a specific occurrence of Error X in the logs. 

Other meta-data _could_ be useful (for other scenarios.. what IDK). 

Yes, it does overlap with Tracing but Tracing is a long way away (and we'd need to include the Trace context in the message to Users too; to be sure SysOps/DevOps/whoever looks at the right bit occurrence in the logs.. would we not?)

manager/src/main/java/com/redhat/service/smartevents/manager/models/Bridge.java
manstis: I am tempted not to; as it does not lend itself to i18n.

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/AbstractWorker.java
manstis: No, but I will refactor to an abstract method and override it accordingly in each `Worker` (`ConnectorWorker` does not need to record any errors as it is handled by its parent `ProcessorWorker`).

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/AbstractWorker.java
manstis: I don't favour that due to i18n ramifications.

IMO the use of Java exceptions is _intended_ to support programmatic try/catch/throw hiding `BridgeError`'s from developers. A lot of RHOSE's codebase has developers throw Java exceptions only to be caught and converted into `BridgeError`. The scenario here, for timeout/retries, bastardizes this concept - simply because an exception is needed for the recording mechanism I've implemented. This could lookup a `BridgeError` from the exception class and use that instead of instantiating an `Exception` only for it to be converted to a `BridgeError` in the `recordError(...)` method.

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/BridgeWorker.java
manstis: Yes, I do believe this is helping (propagate the `Processor` failure to the `Bridge`).

I propagate the error code whereas you're proposing a generic message should the `Processor` be `FAILED`. 

I am not keen on storing or propagating messages as I believe error codes better support i18n.

manstis:
@robpblake @r00ta @Srihari1192 This is ready for review.

Think _mechanism_ not specifically _what_ we log. At this point I merely ensure a single failure state in either the Operator or Manager is stored against the `ManagedResource`. Supported points of failure are exactly where the existing code handles exception states. _What_ and _where_ we may choose to capture and log can still be up for discussion (and the subject of a different JIRA). For example I don't have a `try { .. } catch(..}` block specifically around [RHOAS topic creation](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/196b7c1062725ec1cef4e26502c3a2d00538648c/manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/BridgeWorker.java#L80-L81); so a failure there would be [mapped](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/968/files#diff-9b7048c8bf03785afc5764aaf36abde124dc40721fb041d8db30d727d8ce0b5eR74) to a generic `ProvisioningMaxRetriesExceededException` errorin `AbstractWorker`.

jomarko (commented):
Thank you for the PR @manstis .

From UI team point of view, it is important for us to be notified when PR is merged, so we can request app-services team for new SDK release.

infra/src/test/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorHelperTest.java
```diff
@@ -0,0 +1,99 @@
...
+    private BridgeErrorService service;
+
+    private BridgeErrorHelper helper;
+
+    @BeforeEach
+    public void setup() {
+        this.helper = new BridgeErrorHelper();
+        this.helper.bridgeErrorService = service;
+
+        when(service.getError(UnspecifiedProvisioningFailureException.class)).thenReturn(Optional.of(new BridgeError(1, "code", "reason", BridgeErrorType.USER)));
```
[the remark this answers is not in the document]
manstis: Good catch. Changed.

It didn't really matter as the test is mocking the function of `BridgeErrorService` but it's nice to be consistent.

infra/src/main/resources/exception/exceptionInfo.json
```diff
@@ -202,5 +202,35 @@
     "id": 35,
     "reason": "Shard Operator received an invalid URL for the BridgeIngress endpoint.",
     "type": "PLATFORM"
+  },
+  {
+    "exception": "com.redhat.service.smartevents.infra.exceptions.definitions.platform.ProvisioningFailureException",
+    "id": 36,
+    "reason": "An error occurred handling your request. Please contact the Help Desk for support quoting the date and time of this error.",
```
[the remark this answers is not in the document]
manstis: Good catch.

I've changed `ProvisioningFailureException` to be an `abstract` class removed this entry from `exceptionInfo.json`.

`ProvisioningFailureException` was never instantiated and just acts as the base class for more specific failures.

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorHelper.java
```diff
@@ -0,0 +1,72 @@
...
+        return bridgeErrorService.getError(e.getClass())
+                .map(b -> {
+                    BridgeErrorInstance bei = new BridgeErrorInstance(b);
+                    LOGGER.info("Mapped '{}' to '{}'", e.getClass().getName(), bei);
+                    return bei;
+                })
+                .orElseGet(() -> {
+                    LOGGER.info("'{}' not found in error catalog. Falling back to generic UnspecifiedProvisioningFailureException.", e.getClass().getName());
+                    BridgeErrorInstance bei = new BridgeErrorInstance(deploymentFailedException);
+                    LOGGER.info("Mapped '{}' to '{}'", e.getClass().getName(), bei);
```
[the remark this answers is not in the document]
manstis: I believe it is correct; as the log entry is showing the mapping.

Can you please explain your concern?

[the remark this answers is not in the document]
manstis: This is an example of what we log:
```
2022-08-26 08:19:12,445 ERROR [com.red.ser.sma.man.wor.res.AbstractWorker] (QuarkusQuartzScheduler_Worker-12) Max retry attempts exceeded trying to create dependencies for 'ManagedConnectorProcessor' [54b05059-0d0e-4dd0-b47f-4dca1b3d9461].
2022-08-26 08:19:12,451 INFO  [com.red.ser.sma.inf.exc.BridgeErrorHelper] (QuarkusQuartzScheduler_Worker-12) Mapping 'com.redhat.service.smartevents.infra.exceptions.definitions.platform.ProvisioningMaxRetriesExceededException' to BridgeError: ProvisioningFailureException{message='The maximum number of re-tries for Resource of type 'com.redhat.service.smartevents.manager.models.Processor' with Id '54b05059-0d0e-4dd0-b47f-4dca1b3d9461' was exceeded.'}
	at com.redhat.service.smartevents.manager.workers.resources.AbstractWorker.handleWork(AbstractWorker.java:74)
	at com.redhat.service.smartevents.manager.workers.resources.ProcessorWorker_ClientProxy.handleWork(Unknown Source)
	at com.redhat.service.smartevents.manager.workers.quartz.WorkJob.execute(WorkJob.java:34)
	at org.quartz.core.JobRunShell.run(JobRunShell.java:202)
	at org.quartz.simpl.SimpleThreadPool$WorkerThread.run(SimpleThreadPool.java:573)
```

Srihari1192:
@manstis Please find the below observations 

i) Status message is showing for successful bridge creation as well with error handler scenarios with below status message
` "status_message": "[OPENBRIDGE-1] You tried to create an object which already exist in our repository. Please contact support quoting UUID: e8363833-0271-4394-9558-574e9274f75e"`

ii) Status message is not listing in case of Processor creation failed, intermittently listing only for Slack processor. 

iii) In case of slack processor creation, upon failed state while trying to update or delete throwing 500 Internal server response error with below error

`ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (executor-thread-7) HTTP Request to /api/smartevents_mgmt/v1/shard/bridges/processors failed, error id: 1310ab45-4144-49be-99bd-9a0833dbb829-32: java.lang.NullPointerException
        at com.redhat.service.smartevents.infra.exceptions.mappers.BaseExceptionMapper.mapError(BaseExceptionMapper.java:49)
        at com.redhat.service.smartevents.manager.api.ManagerJsonMappingExceptionMapper_Subclass.mapError$$superforward1(Unknown Source)
        at com.redhat.service.smartevents.manager.api.ManagerJsonMappingExceptionMapper_Subclass$$function$$5.apply(Unknown Source)
        at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:54)
        at io.quarkus.arc.runtime.devconsole.InvocationInterceptor.proceed(InvocationInterceptor.java:62)
        at io.quarkus.arc.runtime.devconsole.InvocationInterceptor.monitor(InvocationInterceptor.java:49)
        at io.quarkus.arc.runtime.devconsole.InvocationInterceptor_Bean.intercept(Unknown Source)
        at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:41)
        at io.quarkus.arc.impl.AroundInvokeInvocationContext.perform(AroundInvokeInvocationContext.java:41)
        at io.quarkus.arc.impl.InvocationContexts.performAroundInvoke(InvocationContexts.java:32)
        at com.redhat.service.smartevents.manager.api.ManagerJsonMappingExceptionMapper_Subclass.mapError(Unknown Source)
        at com.redhat.service.smartevents.infra.exceptions.mappers.JsonMappingExceptionMapper.toResponse(JsonMappingExceptionMapper.java:34)
        at com.redhat.service.smartevents.manager.api.ManagerJsonMappingExceptionMapper_Subclass.toResponse$$superforward1(Unknown Source)
        at com.redhat.service.smartevents.manager.api.ManagerJsonMappingExceptionMapper_Subclass$$function$$3.apply(Unknown Source)
        at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:54)
        at io.quarkus.arc.runtime.devconsole.InvocationInterceptor.proceed(InvocationInterceptor.java:62)
        at io.quarkus.arc.runtime.devconsole.InvocationInterceptor.monitor(InvocationInterceptor.java:49)
        at io.quarkus.arc.runtime.devconsole.InvocationInterceptor_Bean.intercept(Unknown Source)
        at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:41)
        at io.quarkus.arc.impl.AroundInvokeInvocationContext.perform(AroundInvokeInvocationContext.java:41)
        at io.quarkus.arc.impl.InvocationContexts.performAroundInvoke(InvocationContexts.java:32)
        at com.redhat.service.smartevents.manager.api.ManagerJsonMappingExceptionMapper_Subclass.toResponse(Unknown Source)
        at com.redhat.service.smartevents.infra.exceptions.mappers.JsonMappingExceptionMapper.toResponse(JsonMappingExceptionMapper.java:18)
        at com.redhat.service.smartevents.manager.api.ManagerJsonMappingExceptionMapper_Subclass.toResponse$$superforward1(Unknown Source)
        at com.redhat.service.smartevents.manager.api.ManagerJsonMappingExceptionMapper_Subclass$$function$$4.apply(Unknown Source)
        at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:54)
        at io.quarkus.arc.runtime.devconsole.InvocationInterceptor.proceed(InvocationInterceptor.java:62)
        at io.quarkus.arc.runtime.devconsole.InvocationInterceptor.monitor(InvocationInterceptor.java:49)
        at io.quarkus.arc.runtime.devconsole.InvocationInterceptor_Bean.intercept(Unknown Source)
        at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:41)
        at io.quarkus.arc.impl.AroundInvokeInvocationContext.perform(AroundInvokeInvocationContext.java:41)
        at io.quarkus.arc.impl.InvocationContexts.performAroundInvoke(InvocationContexts.java:32)
        at com.redhat.service.smartevents.manager.api.ManagerJsonMappingExceptionMapper_Subclass.toResponse(Unknown Source)
        at com.redhat.service.smartevents.manager.api.ManagerJsonMappingExceptionMapper_ClientProxy.toResponse(Unknown Source)
        at org.jboss.resteasy.core.ExceptionHandler.executeExceptionMapper(ExceptionHandler.java:136)
        at org.jboss.resteasy.core.ExceptionHandler.unwrapException(ExceptionHandler.java:182)
        at org.jboss.resteasy.core.ExceptionHandler.handleReaderException(ExceptionHandler.java:278)
        at org.jboss.resteasy.core.ExceptionHandler.handleException(ExceptionHandler.java:346)
        at org.jboss.resteasy.core.SynchronousDispatcher.writeException(SynchronousDispatcher.java:218)
        at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:519)
        at org.jboss.resteasy.core.SynchronousDispatcher.lambda$invoke$4(SynchronousDispatcher.java:261)
        at org.jboss.resteasy.core.SynchronousDispatcher.lambda$preprocess$0(SynchronousDispatcher.java:161)
        at org.jboss.resteasy.core.interception.jaxrs.PreMatchContainerRequestContext.filter(PreMatchContainerRequestContext.java:364)
        at org.jboss.resteasy.core.SynchronousDispatcher.preprocess(SynchronousDispatcher.java:164)
        at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:247)
        at io.quarkus.resteasy.runtime.standalone.RequestDispatcher.service(RequestDispatcher.java:73)
        at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatch(VertxRequestHandler.java:151)
        at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler$1.run(VertxRequestHandler.java:91)
        at io.quarkus.vertx.core.runtime.VertxCoreRecorder$13.runWith(VertxCoreRecorder.java:543)
        at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2449)
        at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1478)
        at org.jboss.threads.DelegatingRunnable.run(DelegatingRunnable.java:29)
        at org.jboss.threads.ThreadLocalResettingRunnable.run(ThreadLocalResettingRunnable.java:29)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:829)`

Note: Still review in Progress

jomarko (approved):
No more comments form my site.

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/BridgeWorker.java
robpblake: @manstis To be a bit clearer on what I meant. If the error propagated to the `Bridge` is something like "Failed to provision a ReplicaSet", how does the user know whether that relates to their Error Handler or the Bridge?

Is it better for us to have a ErrorHandler-specific code so that if the `FAILURE` status for a bridge is caused by the error handler, its much clearer for the user as to which component is actually causing the failure?

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/AbstractWorker.java
```diff
@@ -68,20 +90,26 @@ public T handleWork(Work work) {
...
         } else if (DEPROVISIONING_STARTED.contains(managedResource.getStatus())) {
             try {
                 updated = deleteDependencies(work, managedResource);
-                complete = isDeprovisioningComplete(updated);
             } catch (Exception e) {
                 LOGGER.info(String.format("Failed to delete dependencies for resource of type '%s' with id '%s'.", work.getType(), id), e);
                 // Something has gone wrong. We need to retry.
                 workManager.rescheduleAfterFailure(work);
+                recordError(work, e);
+                updated = load(id);
```
robpblake, line 110: @manstis I'm not sure the logic here feels right. If I read the code correctly, we could end up calling both `rescheduleAfterFailure` and `reschedule` if the `complete` boolean is set to `false` in the `finally` block.

I'm also not sure we need to record the intermediate failure status as it would only be visible on the resource in between retries. As a suggestion, I think it would be absolutely fine to record the final failure status_code e.g timed out or max retries exceeded.

manager/src/main/resources/db/migration/0.1/V0.1.23__add_bridge_error_info.sql
```diff
@@ -0,0 +1,8 @@
+-- add new "error_code" column
+alter table BRIDGE
+    add column "bridge_error_id" integer,
```
robpblake, line 3: I think the comment in the `.sql` is a bit misleading. I also personally think `error_id` and `error_uuid` would be OK as column names.

infra/src/main/resources/exception/exceptionInfo.json
```diff
@@ -202,5 +202,29 @@
...
+  {
+    "exception": "com.redhat.service.smartevents.infra.exceptions.definitions.platform.ProvisioningMaxRetriesExceededException",
+    "id": 38,
+    "reason": "Your request failed after the maximum number of configured re-try attempts. Please contact the Help Desk for support quoting the date and time of this error.",
+    "type": "PLATFORM"
+  },
+  {
+    "exception": "com.redhat.service.smartevents.infra.exceptions.definitions.platform.ProvisioningReplicaFailureException",
+    "id": 39,
+    "reason": "A ReplicaSet failed to be provisioned for your request. Please contact the Help Desk for support quoting the date and time of this error.",
```
robpblake, line 227: I'm personally -1 on the "contact the Help Desk" part of these errors. I think the first part is all that the user needs.

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorHelper.java
```diff
@@ -0,0 +1,72 @@
...
+    public BridgeErrorInstance getBridgeErrorInstance(Exception e) {
+        LOGGER.info("Mapping '{}' to BridgeError", e.getClass().getName(), e);
+        return bridgeErrorService.getError(e.getClass())
+                .map(b -> {
+                    BridgeErrorInstance bei = new BridgeErrorInstance(b);
+                    LOGGER.info("Mapped '{}' to '{}'", e.getClass().getName(), bei);
+                    return bei;
+                })
+                .orElseGet(() -> {
+                    LOGGER.info("'{}' not found in error catalog. Falling back to generic UnspecifiedProvisioningFailureException.", e.getClass().getName());
```
robpblake, line 51: Personally, I'd replace the `UnspecifiedProvisioningFailureException` with the `ProvisioningFailureException`. I think it can act as the catch-all and saves us an Error code.

manager/src/main/java/com/redhat/service/smartevents/manager/ProcessorServiceImpl.java
robpblake: @manstis minor, but this could be moved after the `if (ManagedResourceStatus.DELETED == updateDTO.getStatus())` check

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
```diff
@@ -87,10 +87,20 @@ public void createBridgeExecutor(ProcessorDTO processorDTO) {
 
             // create or update the secrets for the bridgeExecutor
             createOrUpdateBridgeExecutorSecret(bridgeExecutor, processorDTO);
-        } else if (existing.getStatus().isReady()) {
-            LOGGER.info("BridgeExecutor '{}' already exists and is ready. Notifying manager that it is ready.", processorDTO.getId());
-            ProcessorManagedResourceStatusUpdateDTO updateDTO =
-                    new ProcessorManagedResourceStatusUpdateDTO(processorDTO.getId(), processorDTO.getCustomerId(), processorDTO.getBridgeId(), ManagedResourceStatus.READY);
+        } else {
+            ManagedResourceStatus inferredStatus = existing.getStatus().inferManagedResourceStatus();
+            // The Controller would have notified the Manager with PROVISIONING before it first started.
```
robpblake, line 92: @manstis I'm a little unsure if this change is related to the `status_message` feature? Could you please clarify.

cheers,
Rob

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeIngressController.java
```diff
@@ -150,6 +162,18 @@ public DeleteControl cleanup(BridgeIngress bridgeIngress, Context context) {
         return DeleteControl.defaultDelete();
     }
 
+    @Override
+    public Optional<BridgeIngress> updateErrorStatus(BridgeIngress bridgeIngress, RetryInfo retryInfo, RuntimeException e) {
```
robpblake, line 166: @manstis For my understanding,  I assume the `RetryInfo` here relates to the configuration of the `KubernetesClient`?

manager/src/main/resources/db/migration/0.1/V0.1.23__add_bridge_error_info.sql
manstis: Fixed.

infra/src/main/resources/exception/exceptionInfo.json
manstis: Removed.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeIngressController.java
manstis: Correct. This method is called by k8s should it fail to reconcile a resource within it's "timeout/retry" parameters.

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/definitions/platform/UnspecifiedProvisioningFailureException.java
```diff
@@ -0,0 +1,9 @@
+package com.redhat.service.smartevents.infra.exceptions.definitions.platform;
+
+public class UnspecifiedProvisioningFailureException extends ProvisioningFailureException {
```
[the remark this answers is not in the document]
manstis: Fair point. Refactored.

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorHelper.java
manstis: Done.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
manstis: We already supported the scenario where the Controller had already started `PROVISIONING` a resource.

This change makes it work similarly for when a Controller has already started `DELETING` a resource.

It was found debugging the work for `status_message`.

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorHelper.java
```diff
@@ -0,0 +1,72 @@
...
+        if (error.isPresent()) {
+            deploymentFailedException = error.get();
+        } else {
+            LOGGER.error("{} error is not defined in the ErrorsService.", UnspecifiedProvisioningFailureException.class.getSimpleName());
+            Quarkus.asyncExit(1);
+        }
+    }
+
+    public BridgeErrorInstance getBridgeErrorInstance(Exception e) {
+        LOGGER.info("Mapping '{}' to BridgeError", e.getClass().getName(), e);
```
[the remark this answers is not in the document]
robpblake: Cheers!

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/AbstractWorker.java
manstis: I agree. Good spot. Refactored.

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorHelper.java
```diff
@@ -0,0 +1,78 @@
...
+        LOGGER.debug("Mapping '{}' to BridgeError", e.getClass().getName(), e);
+        return bridgeErrorService.getError(e.getClass())
+                .map(b -> {
+                    BridgeErrorInstance bei = new BridgeErrorInstance(b);
+                    LOGGER.info("Mapped '{}' to '{}'", e.getClass().getName(), bei);
+                    return bei;
+                })
+                .orElseGet(() -> {
+                    LOGGER.debug("'{}' not found in error catalog. Falling back to generic ProvisioningFailureException.", e.getClass().getName());
+                    BridgeErrorInstance bei = new BridgeErrorInstance(deploymentFailedException);
```
r00ta, line 52: Nitpick: Not sure if `ProvisioningFailureException` is generic enough. The `reason` of that exception looks good `An unspecified error occurred handling your request` but `Provisioning` sounds like something that can occur only during the provisioning. I'm wondering if `InternalPlatformException` is better in this case. In any case up to you, just brainstorming

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorInstance.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public BridgeErrorInstance(BridgeError bridgeError) {
+        super(bridgeError.getId(), bridgeError.getCode(), bridgeError.getReason(), bridgeError.getType());
+        this.uuid = UUID.randomUUID().toString();
+    }
+
+    @JsonIgnore
+    public String getUUID() {
+        return uuid;
+    }
```
r00ta, line 27: I can't remember by heart so I might be wrong. We are going to include the `uuid` in the serialized object right? If so, could you use `@JsonProperty` on the `uuid`? I don't remember exactly the magic behavior of Jackson in this case, but if you remove `@JsonIgnore` the result should be the same right?

infra/src/main/resources/exception/exceptionInfo.json
```diff
@@ -202,5 +202,35 @@
...
+    "id": 36,
+    "reason": "An unspecified error occurred handling your request.",
+    "type": "PLATFORM"
+  },
+  {
+    "exception": "com.redhat.service.smartevents.infra.exceptions.definitions.platform.ProvisioningTimeOutException",
+    "id": 37,
+    "reason": "Your request was not complete within the configured time-out.",
+    "type": "PLATFORM"
+  },
```
r00ta, line 217: Is this a `PLATFORM` exception? Should it be `USER`? Same for the others

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/AbstractWorker.java
```diff
@@ -30,6 +33,9 @@
 
 public abstract class AbstractWorker<T extends ManagedResource> implements Worker<T> {
 
+    private static final String RETRIES_FAILURE_MESSAGE = "The maximum number of re-tries for Resource of type '%s' with Id '%s' was exceeded.";
+    private static final String TIMEOUT_FAILURE_MESSAGE = "The timeout to process Work for Resource of type '%s' with Id '%s' was exceeded.";
+
```
r00ta, line 38: What about moving these constants to the classes `ProvisioningMaxRetriesExceededException` and `ProvisioningTimeOutException`?

r00ta (commented):
Looks good to me, just a couple of small questions for my understanding

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorHelper.java
manstis: Fair point. Removed class and used `InternalPlatformException` in its place.

infra/src/main/resources/exception/exceptionInfo.json
manstis: Well.. these exceptions come about from the PLATFORM failing to do something the USER requested.

I thought USER [exceptions](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/definitions/user/ExternalUserException.java#L6) were limited to errors relating to the User's input?

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorInstance.java
manstis: Done (and now consistent with our other DTOs).

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/AbstractWorker.java
manstis: Done.

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorInstance.java
manstis: I had to add back the `@JsonIgnore` otherwise I was getting two properties in the Open API definition:
```
uuid:
  type: string
uUID:
  type: string
```

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/AbstractWorker.java
```diff
@@ -141,4 +161,7 @@ protected boolean isTimeoutExceeded(Work work, ManagedResource managedResource)
     protected abstract boolean isProvisioningComplete(T managedResource);
 
     protected abstract boolean isDeprovisioningComplete(T managedResource);
+
+    protected abstract T recordError(Work work, Exception e);
```
robpblake, line 165: @manstis With you now having the `error_id` and `error_uuid` on the `ManagedResource`, I think we can provide a generic implementation of this method rather than having to provide the implementation in each sub-class:

```
    private T recordError(Work work, Exception e) {
        String managedResourceId = getId(work);
        T managedResource = load(managedResourceId);
        BridgeErrorInstance bridgeErrorInstance = bridgeErrorHelper.getBridgeErrorInstance(e);
        managedResource.setErrorId(bridgeErrorInstance.getId());
        managedResource.setErrorUUID(bridgeErrorInstance.getUuid());
        return persist(managedResource);
    }
```

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/BridgeWorker.java
robpblake: @manstis Gentle nudge for your opinion on this one.

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/HasErrorInformation.java
```diff
@@ -0,0 +1,8 @@
+package com.redhat.service.smartevents.infra.exceptions;
+
+public interface HasErrorInformation {
```
robpblake, line 3: @manstis Do we need this now you've moved the fields onto the `ManagedResource`?

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorHelper.java
```diff
@@ -0,0 +1,78 @@
...
+                    LOGGER.info("Mapped '{}' to '{}'", e.getClass().getName(), bei);
+                    return bei;
+                });
+    }
+
+    public String makeUserMessage(HasErrorInformation hasErrorInformation) {
```
robpblake, line 58: @manstis `ManagedResource` as the parameter now?

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
robpblake: @manstis OK. Thank you for being proactive on the fixes. 

In general, I think it would help reviewers in the future if fixes not related specifically to the feature in development in the PR are done in smaller PRs. It just helps to reduce the amount we have to consider when reviewing.

cheers,
Rob

robpblake (commented):
Some minor feedback @manstis . Thank you.

Are you comfortable with the size of this PR? There are a lot of changes and an option is to split the changes into Manager side and Shard side to reduce the complexity of each PR. I'll let you decide how you want to proceed.

cheers,
Rob

manager/src/main/java/com/redhat/service/smartevents/manager/workers/resources/AbstractWorker.java
manstis: Good spot. Changed.

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/HasErrorInformation.java
manstis: Technically you are correct; however it'd mean moving the error handling code from `infra` to `manager`. This, however, cannot be done as the error handling code used by both the `manager` and `operator.` The interface allows decoupling of these modules.

infra/src/main/java/com/redhat/service/smartevents/infra/exceptions/BridgeErrorHelper.java
manstis: Would be nice to; but cannot do due to the same reason given above.

robpblake (approved):

r00ta (approved):

Srihari1192 (approved):
LGTM

PR #972, opened 2022-07-25
[434cd] Update kustomization images

r00ta:


vaibhavjainwiz:
Jira Id is mandatory to process MR in CPaaS. Could you please assign a Jira to it.

PR #977, opened 2022-07-26
[MGDOBR-968] - Create Logger action for development

vaibhavjainwiz:
https://issues.redhat.com/browse/MGDOBR-968

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>

### DIFF
--- a/processor/configuration/src/main/resources/schemas/actions/catalog.json
+++ b/processor/configuration/src/main/resources/schemas/actions/catalog.json
@@ -40,5 +40,11 @@
     "name": "Webhook",
     "description": "Send the event to a webhook.",
     "isConnector": false
+  },
+  {
+    "id": "logger_sink_0.1",
+    "name": "logger",
+    "description": "Log event.",
+    "isConnector": false
   }
 ]
--- a/processor/configuration/src/main/resources/schemas/actions/logger_sink_0.1.json
+++ b/processor/configuration/src/main/resources/schemas/actions/logger_sink_0.1.json
@@ -1,0 +1,15 @@
+{
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "test": {
+      "type": "string",
+      "title": "Test",
+      "description": "Test.",
+      "example": "test"
+    }
+  },
+  "required": [
+    "test"
+  ]
+}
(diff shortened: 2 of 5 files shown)


processor/configuration/src/main/resources/schemas/actions/catalog.json
```diff
@@ -40,5 +40,11 @@
     "name": "Webhook",
     "description": "Send the event to a webhook.",
     "isConnector": false
+  },
+  {
+    "id": "logger_sink_0.1",
+    "name": "logger",
+    "description": "Log event.",
+    "isConnector": false
```
r00ta, line 48: The user of the service has no access to the executor logs, I'd not expose it in the catalog as this is just an internal debugging detail

vaibhavjainwiz: Current implementation enforce us to define it in catalog else it give error.

manstis: I agree; you cannot expose this _development_ Action in the usual way.

If it is not in the catalogue you may find how it is validated also needs refactoring.

tl;dr; adding a Logger Action is not as straight forward as you think.

manstis (changes requested):
You cannot add a _development_ Action the way you have.

The whole mechanism of catalogue, lookup and validation needs to be changed to support this.

processor/configuration/src/main/resources/schemas/actions/catalog.json
vaibhavjainwiz: Our product not going to customer any sooner. This logger action could help developer and tester both to speed up development. 

There could be multiple ways to hide it from customer like we could create a `development_catalog.json` where we could define these type of Action. We could think of more ways.

manstis: I agree with your reasoning; but this PR, IMO, should also include whatever "...way[s] to hide it from customer..." otherwise it will be forgotten and never addressed.

vaibhavjainwiz:
> You cannot add a _development_ Action the way you have.
> 
> The whole mechanism of catalogue, lookup and validation needs to be changed to support this.

sorry.. I didn't get you. What need to be change?
It you are looking for test parameter then it because current implementation enforce it. It a bug in current implementation that need to be fixed.

PR #978, opened 2022-07-26
MGDOBR-829: Add Data Plane deployment scripts

winklerm:
[https://issues.redhat.com/browse/MGDOBR-829](https://issues.redhat.com/browse/MGDOBR-829)

Added the scripts for deploying the Data Plane (shard-operator). 

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>


winklerm:
Still need to properly set the parameters to the operator deployment.

kustomize/data-plane/README.md
```diff
@@ -0,0 +1,37 @@
...
+    EVENT_BRIDGE_SHARD_ID
+    EVENT_BRIDGE_SSO_CLIENT_ID
+    EVENT_BRIDGE_SSO_SECRET
+    WEBHOOK_CLIENT_ID
+    WEBHOOK_CLIENT_SECRET
+    WEBHOOK_TECHNICAL_ACCOUNT_ID
+  ```
+  No values are specified in the properties file, they are expected to be set in the environment variables.
+3. Log in to target cluster (oc login)
+4. If no external SSO is used, deploy Keycloak by running [deployKeycloak.sh](deployKeycloak.sh)
```
sutaakar, line 31: Does it have sense to use Keycloak deployment? IMHO using SSO would be easier, with less code to maintain.

winklerm: Ah, good point, we can definitely switch to SSO, I did not know what we expect for the tests.

It should be only a matter of not calling the Keycloak scripts (and providing the expected values), do you think it makes sense to keep them or they can be deleted?

sutaakar: I think we can delete them, so we don't have to maintain it

sutaakar (approved):

kustomize/data-plane/prerequisites/service-mesh/resources/service-mesh-control-plane.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: maistra.io/v2
+kind: ServiceMeshControlPlane
+metadata:
+  name: basic
+  namespace: istio-system
+spec:
+  addons:
+    grafana:
```
jiripetrlik, line 8: Just out of curiosity. Do we need these addons for data plane? These are focused on monitoring. If we need to speed up, I would suggest to remove it.

jiripetrlik (commented):
Looks good to me. I have one question.

kustomize/data-plane/README.md
winklerm: Thanks, done!

kustomize/data-plane/prerequisites/service-mesh/resources/service-mesh-control-plane.yaml
winklerm: Good question! I guess we can remove the grafana addon, not sure about jaeger (if it is required by istio or not). I will try to remove grafana and see if the deployment is successful.

The current time of the data plane deployment is ~ 1 minute 40 seconds which seems not to be critical.

kustomize/data-plane/deployDataPlane.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+source common_tools.sh
+
+oc get project istio-system
+status=$?
+if [ $status -ne 0 ]; then
+  echo "The istio-system project does not exist, creating"
+  oc new-project istio-system
```
robpblake, line 9: @winklerm Is there any reason why the creation of the namespace cannot be in the `kustomize` resources so that's created declaratively rather than imperatively?

Longer-term, what should be creating this `namespace`? The actual installation of the Openshift ServiceMesh operator?

kustomize/data-plane/shard/operator-csv-patch.json
```diff
@@ -0,0 +1,15 @@
+[
+  { "op":  "add",
```
robpblake, line 2: @winklerm Why do we need this patch?

kustomize/data-plane/README.md
```diff
@@ -0,0 +1,37 @@
...
+  ```
+    EVENT_BRIDGE_FLEET_SHARD_INDEX_IMAGE
+    EVENT_BRIDGE_SSO_URL=http://keycloak:8180/auth/realms/event-bridge-fm
+    EVENT_BRIDGE_MANAGER_URL=http://event-bridge:8080
+  ```
+  Please note that not providing a value in the properties file (such as for `EVENT_BRIDGE_FLEET_SHARD_INDEX_IMAGE`) 
+  makes kustomize use a value from the environment variable with the same name.
+
+  **Important:** The values set in the properties file take precedence over environment variables.
+- The secrets for shard-operator deployment are specified in [secret-properties.env](secret-properties.env):
```
robpblake, line 21: @winklerm What's the reason we need another approach to secret handling? Does Sealed Secrets not work for you?

kustomize/data-plane/environment-properties.env
```diff
@@ -0,0 +1,3 @@
+EVENT_BRIDGE_FLEET_SHARD_INDEX_IMAGE
+EVENT_BRIDGE_SSO_URL
```
robpblake, line 2: Apart from local development, I don't think this will ever be anything other than sso.redhat.com if that helps to simply this.

kustomize/data-plane/prerequisites/service-mesh/resources/service-mesh-control-plane.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: maistra.io/v2
```
robpblake, line 1: @winklerm We seem to be duplicating resources here from the `kustomize/base-openshift/shard` directory. What's the reason we need to do that please?

kustomize/data-plane/shard/resources/operator-config.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: smart-events-shard-operator-config
+  namespace: openshift-operators
+data:
+  EVENT_BRIDGE_ISTIO_GATEWAY_NAME: istio-ingressgateway
```
robpblake, line 7: @winklerm You're just re-using the default values of `EVENT_BRIDGE_ISTIO_GATEWAY_NAME` , `EVENT_BRIDGE_ISTIO_GATEWAY_NAMESPACE` and `EVENT_BRIDGE_K8S_PLATFORM`.

kustomize/data-plane/deployDataPlane.sh
```diff
@@ -0,0 +1,32 @@
...
+fi
+
+echo "Deploying all resources"
+oc kustomize . | oc apply -f -
+
+status=$?
+if [ $status -ne 0 ]; then
+  echo "Some resources fail to deploy (concurrency issues), redeploying"
```
robpblake, line 17: Are you able to provide a list of the resources that fail to deploy? Ideally we should provide a permanent fix to it

robpblake (commented):
@winklerm This seems to be slightly mixing two concepts into a single prerequisites directory:

- Install our prerequisites e.g Openshift Serverless, Openshift ServiceMesh
- Install the RHOSE Fleet Shard Operator in a manner very specific to the work you're doing 

I like the prerequisite idea, but I think we should make it re-usable across all environments to reduce the duplication here. So for example I could image a layout like this:

- prerequisites
- prerequisites/data-plane/base/openshift-serverless
- prerequisites/data-plane/base/openshift-servicemesh
- prerequisites/data-plane/overlays/dev 
- prerequisites/data-plane/overlays/stage
- prerequisites/data-plane/overlays/<env>

That follows the same pattern that we have now and would allow us to remove some of the duplicated files that currently existing in the `openshift-base` directory. Our worfklow for installing an instance of the data-plane would then be (using dev environment as an example):

1. `kustomize build prerequisites/data-plane/overlays/dev | oc apply -f -` (or using your wrapper script to waiting for dependencies)
2. `kustomize build overlays/dev | oc apply -f -`

I'm also concerned that we're introducing yet another approach to secret management. Can you explain why we need to do that?

cheers,

Rob

kustomize/data-plane/deployDataPlane.sh
winklerm: This is a difference in the upstream istio operator and the servicemesh operator that I have run into: The ServiceMesh operator documentation [1] says the namespace should be created manually (while istioctl will create it automatically).

Good point with creating the namespace declaratively, I just followed the docs and did not realize that, I will change this, thanks!

[1] https://docs.openshift.com/container-platform/4.10/service_mesh/v2x/ossm-create-smcp.html#ossm-control-plane-deploy-cli_ossm-create-smcp

kustomize/data-plane/README.md
winklerm: I wanted to avoid setting the sealed secrets private key on the QE cluster (and other clusters); in addition, this approach allows you to inject any secrets you need in your Jenkins job which seemed more flexible and secure.

I agree it is yet another approach, thinking about it again..

kustomize/data-plane/environment-properties.env
winklerm: Thanks! This could be set as a default, then.

kustomize/data-plane/deployDataPlane.sh
winklerm: This was taken from [1] - but I can see now that this script has been removed, I will remove that condition and see if it actually fails now.

[1] https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/ee48af577a981f6fd33c983147d1fc69a6ac29cf/kustomize/startMinikubeDeployLocalDev.sh#L60

kustomize/data-plane/shard/operator-csv-patch.json
winklerm: The patch was needed in order to set parameters for the shard operator - since we use subscription to install it, we cannot specify the parameters directly in the deployment. The operator needs to be installed, which creates Operator CSV with a deployment template, which can be patched then.

When I tried to learn about possible approach to this, I was suggested also to provide the configuration for the operator as a custom resource, that way it could work even when the operator is updated to a newer version (as opposed to setting the operator parameters in the CSV deployment template which would probably get overwritten with an operator update). However, that would mean additional work, not sure if that would make sense fro your perspective, though?

Right now I have a feeling that this could be replaced by configuring an addon for the shard-operator (afaik, addons support parameters), not sure if that could be used outside of the managed environment?

kustomize/data-plane/README.md
robpblake: @winklerm I'd really advocate for keeping a single, consistent approach to secret management across everything. We're a very small team and having one way that everyone understands for secret management will help us all out.

We could for example install the Sealed Secrets operator as part of the prerequisistes and then you simply configure the key required for your environment as part of the CI pipeline? So pattern would be:

1. prerequisites installs SealedSecrets
2. your key for ^ is installed via the setup part of your CI pipeline, with the value being maintained in Jenkins (or whatever else is secure)
3. We store the secrets for your environment here, using the same consistent approach as we do for all our other environments.

kustomize/data-plane/prerequisites/service-mesh/resources/service-mesh-control-plane.yaml
robpblake: @winklerm Let's take it step-by-step, but it's important that we have the Jira task in our backlog to remove the duplicates so that we keep our setup clean and that the rest of the team isn't confused as to why there is so much duplication.

So I'd break it down into two/three steps, something like:

1. Create the prerequisites directory structure here with overlay for your environment only and verify that it works
2. Follow up PR to create overlay for `dev` environment, removing the pre-requisites stuff from the `base-openshift` directory with email to team on `kustomize` structure changes

kustomize/data-plane/deployDataPlane.sh
robpblake: Thanks 👍

kustomize/data-plane/deployDataPlane.sh
robpblake: @winklerm Yeah let's see if it will work declaratively in our `kustomize` resources. Thanks for investigating.

cheers,
Rob

kustomize/data-plane/prerequisites/knative/resources/knative-eventing.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: operator.knative.dev/v1alpha1
+kind: KnativeEventing
+metadata:
+  name: knative-eventing
+  namespace: knative-eventing
+spec: {}
```
robpblake, line 6: @winklerm Is there a way to constrain which version of Knative evening is installed using this method?

kustomize/data-plane/shard/operator-csv-patch.json
robpblake: @winklerm Thank you for the extra context.

I think if this gets you up and running, let's continue with it. But I think we should probably fix it longer-term. The order of fix preference I would say is:

1. Make the CVP pipeline use OSD and AddOn model of deployment. This would have the additional benefit of aligning the CVP pipeline more closely with our `dev`, `stage` and `prod` environments.
2. We could do something with the Quarkus Kubernetes Config extension https://quarkus.io/guides/kubernetes-config to ensure that the Operator always reads a particular `ConfigMap` or `Secret` from the `namespace` in which it is installed.

WDYT?

kustomize/data-plane/README.md
winklerm: Agreed, this pattern could work, will take a look.

winklerm: Now I realized what I like about the properties files with the env. vars and secrets - I like that they specify the "interface" - you do not need to go through several kustomize resource files to find all the places you need to adjust something. For a person who is not familiar with the deployment scripts, it seems to be more straightforward (could be just my personal opinion).

kustomize/data-plane/shard/resources/operator-config.yaml
winklerm: Yes, I was not sure if they were required to be in the configmap, probably not, thanks!

robpblake:
@winklerm I think we can now close this PR. Do you agree?
cheers,
Rob

r00ta:
Closing this. Feel free to reopen it in case

PR #956, opened 2022-07-18
MGDOBR-306 - Implement ittest for KafkaTopicAction

jiripetrlik:
[MGDOBR-306](https://issues.redhat.com/browse/MGDOBR-306)

Implement integration test which will send a message to Kafka topic. Test should verify that the message was successfully published. Managed Kafka could be used for this test.

Please make sure that your PR meets the following requirements:

- [*] Your code is properly formatted according to [this configuration](https://github.com/kiegroup/kogito-runtimes/tree/main/kogito-build/kogito-ide-config).  
  *Run `mvn clean verify -DskipTests` so that imports are correctly set.*
- [*] Your commit messages are clear and reference the JIRA issue e.g: "[MGDOBR-1] - $clear_explanation_of_what_you_did"
- [*] The layers in the `kustomize` folder have been updated accordingly.
- [*] All new functionality is tested
- [*] Pull Request title is properly formatted: `MGDOBR-XYZ Subject`
- [*] Pull Request contains link to the JIRA issue
- [*] Pull Request contains link to any dependent or related Pull Request
- [*] Pull Request contains description of the issue
- [*] Pull Request does not include fixes for issues other than the main ticket

## PR reviewer(s) checklist

- [ ] Check if code and Github action workflows have been modified and if the modification is safe
- [ ] If the modification is safe, add the `safe to test` label

<details>
<summary>
How to trigger pipelines and use the bots:
</summary>

- <b>Run the end to end pipeline</b>  
  Annotate the pull request with the label: `safe to test`. If you want to run the pipeline again, remove and add it again.

- <b>Rebase the pull request</b>  
  Comment with: `/rebase`.
- <b>Deploy BOT</b>
  Comment with `/deploy <dev,stable> when the PR has been merged to deploy to a target environment.

</details>

PR #985, opened 2022-07-26
MGDOBR-306 - Implement ittest for KafkaTopicAction

jiripetrlik:
Replacement of PR: https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/956 which had to be reverted by connection timeouts of Kafka client.

### DIFF
--- a/integration-tests/src/test/java/com/redhat/service/smartevents/integration/tests/resources/kafka/KafkaResource.java
+++ b/integration-tests/src/test/java/com/redhat/service/smartevents/integration/tests/resources/kafka/KafkaResource.java
@@ -1,0 +1,175 @@
+package com.redhat.service.smartevents.integration.tests.resources.kafka;
+
+import java.time.Duration;
+import java.util.Collections;
+import java.util.List;
+import java.util.Properties;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.common.serialization.StringDeserializer;
+import org.keycloak.representations.AccessTokenResponse;
+
+import com.openshift.cloud.api.kas.auth.AclsApi;
+import com.openshift.cloud.api.kas.auth.TopicsApi;
+import com.openshift.cloud.api.kas.auth.invoker.ApiClient;
+import com.openshift.cloud.api.kas.auth.invoker.ApiException;
+import com.openshift.cloud.api.kas.auth.invoker.Configuration;
+import com.openshift.cloud.api.kas.auth.invoker.auth.OAuth;
+import com.openshift.cloud.api.kas.auth.models.AclBinding;
+import com.openshift.cloud.api.kas.auth.models.AclOperation;
+import com.openshift.cloud.api.kas.auth.models.AclOperationFilter;
+import com.openshift.cloud.api.kas.auth.models.AclPatternType;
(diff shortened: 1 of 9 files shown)


lucamolteni (approved):
This is fine by me, let's wait for the system to be stable to merge this to avoid having even more variables

integration-tests/src/test/java/com/redhat/service/smartevents/integration/tests/resources/kafka/KafkaResource.java
```diff
@@ -0,0 +1,155 @@
...
+                .patternType(AclPatternType.LITERAL)
+                .principal("User:" + kafkaConnectionParameters.getOpsClientID())
+                .operation(AclOperation.WRITE)
+                .permission(AclPermissionType.ALLOW);
+        try {
+            aclsApi.createAcl(aclBindingRead);
+            aclsApi.createAcl(aclBindingWrite);
+        } catch (ApiException e) {
+            throw new RuntimeException("Error assigning ACL to topic", e);
+        }
```
danielezonca, line 74: I would make each call in a different `try/catch` because we should try to revert the part that has been already created (i.e. the topic and the first ACL).
What I mean is a `cleanup` method that tries to delete the other resources before stopping the execution

integration-tests/src/test/java/com/redhat/service/smartevents/integration/tests/resources/kafka/KafkaResource.java
```diff
@@ -0,0 +1,155 @@
...
+                    AclOperationFilter.READ,
+                    AclPermissionTypeFilter.ALLOW);
+            aclsApi.deleteAcls(AclResourceTypeFilter.TOPIC,
+                    topicName,
+                    AclPatternTypeFilter.LITERAL,
+                    "User:" + kafkaConnectionParameters.getOpsClientID(),
+                    AclOperationFilter.WRITE,
+                    AclPermissionTypeFilter.ALLOW);
+        } catch (Exception e) {
+            throw new RuntimeException("Error removing ACL for topic", e);
```
danielezonca, line 95: Same here

integration-tests/src/test/resources/features/action-kafka.feature
```diff
@@ -0,0 +1,43 @@
+Feature: Kafka Action tests
+
+  Background:
+    Given authenticate against Manager
+
+  @kafkaaction
+  Scenario: Send message to Kafka topic using the Kafka action
+    And create a new Kafka topic "topic1"
```
danielezonca, line 8: Given that we use the same kafka cluster, can this test produce errors if multiple jobs are running in parallel?
They both can create/delete the same topic concurrently.

integration-tests/src/test/java/com/redhat/service/smartevents/integration/tests/resources/kafka/KafkaResource.java
jiripetrlik: @danielezonca Thank you. Good point. Now all previously allocated resources should be deleted in case of exception.

integration-tests/src/test/java/com/redhat/service/smartevents/integration/tests/resources/kafka/KafkaResource.java
jiripetrlik: I'm not sure here. It is more about deleting of ACLS which were already created. Basically all items need to be destroyed.

integration-tests/src/test/resources/features/action-kafka.feature
jiripetrlik: Yes, I understand. There is already some logic for it. It will append prefix with random hash to the name. https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/985/files#diff-f6186aa2d4db523463b2612cf5b1f828ec29566cf37aa7a90287e95c1c1a55e7R85

danielezonca (approved):

PR #1065, opened 2022-08-24
WIP: Update knative 1.2 to 1.3

matzew:
**NOTE:** this should be rebased once https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/917 is merged

Updating to the new Knative 1.3 version (backed by the Serverless Operator 1.24)

I guess I have no JIRA, but I can create on.

* Updates to Eventing Core, from 1.2 to 1.3 from [here](https://github.com/openshift/knative-eventing/blob/release-v1.3/openshift/release/artifacts/eventing-core.yaml)
* Updates to KnativeKafka from 1.2 to 1.3, sources from [here](https://github.com/openshift-knative/eventing-kafka-broker/blob/release-v1.3/openshift/release/artifacts/eventing-kafka.yaml)

PR #1070, opened 2022-08-25
[DRAFT] MGDOBR-1017: Namespace stuck in terminating state

manstis:
**=== DRAFT ===**

[https://issues.redhat.com/browse/MGDOBR-1017](https://issues.redhat.com/browse/MGDOBR-1017)

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>


shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeExecutorServiceImpl.java
```diff
@@ -53,7 +53,7 @@ public class BridgeExecutorServiceImpl implements BridgeExecutorService {
     KubernetesClient kubernetesClient;
 
     @Inject
-    CustomerNamespaceProvider customerNamespaceProvider;
+    CustomerBridgeNamespaceProvider customerBridgeNamespaceProvider;
```
manstis, line 56: tl;dr; I added `BridgeId` to the namespace provider (in addition to `CustomerId`).

Did we want to retain the use of `CustomerId` (or just use `BridgeId`)?

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeIngressServiceImpl.java
```diff
@@ -83,13 +85,14 @@ public void createBridgeIngress(BridgeDTO bridgeDTO) {
...
-                kubernetesClient
-                        .resources(BridgeIngress.class)
-                        .inNamespace(namespace)
-                        .delete(BridgeIngress.fromDTO(bridgeDTO, namespace));
-        if (!bridgeDeleted) {
+        String customerId = bridgeDTO.getCustomerId();
+        String bridgeId = bridgeDTO.getId();
+
+        // Deletion of the Namespace will cascade deletion of the BridgeIngress
+        final Namespace namespace = customerBridgeNamespaceProvider.fetchOrCreateCustomerBridgeNamespace(customerId, bridgeId);
```
manstis, line 92: Since we have a `Namespace` per `Bridge` we can delete the whole `Namespace`.

I discussed cascade deletion with the Java Operator SDK team and they confirmed it's behaviour.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/providers/CustomerBridgeNamespaceProviderImpl.java
```diff
@@ -62,8 +63,11 @@ public void deleteNamespaceIfEmpty(Namespace namespace) {
             // deletion can be tricky with finalizers, pending objects, etc.
             // let's start simple and build on top of the constraints and scenarios as we evolve.
             if (!kubernetesClient.namespaces().withName(namespace.getMetadata().getName()).delete()) {
-                LOGGER.warn("Namespace '{}' hasn't been deleted", namespace.getMetadata().getName());
+                LOGGER.warn("Namespace '{}' has not been deleted.", namespace.getMetadata().getName());
             }
+        } else {
+            LOGGER.warn("Namespace '{}' has not been deleted as it contained other resources.", namespace.getMetadata().getName());
```
manstis, line 69: I _should_ be able to delete the [scheduled](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1070/files#diff-59cb844d16bf6e217094b6f0b48152d3ca0cf3fc2c5e258e05ab8f3f011afad9R48) method and the `deleteNamespaceIfEmpty(..)` [method](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1070/files#diff-59cb844d16bf6e217094b6f0b48152d3ca0cf3fc2c5e258e05ab8f3f011afad9R61). 

Would you agree?

They both serve as "house keeping" on `main` to delete a `Namespace` when all of it's `BridgeIngress` Custom Resources are removed (that should not be required now we have a `Namespace` per `Bridge`).

manstis:
@r00ta I fear this may have unforeseen consequences on KNative.

This will now provision a KNative `Broker` for each `Bridge` whereas previously we had a `Broker` per `Customer` (shared across multiple `Bridge`'s). That screams at me but I'm not the expert here. Is this likely to kill our Cluster capacity with a potential explosion in KNative Broker CR instances? A shared `Broker` (on a shared Customer `Namespace`) feeding into many (`Bridge`) `Namespace`'s sounds like a problem in the making too.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeIngressServiceImpl.java
```diff
@@ -83,13 +85,22 @@ public void createBridgeIngress(BridgeDTO bridgeDTO) {
...
-                        .resources(BridgeIngress.class)
-                        .inNamespace(namespace)
-                        .delete(BridgeIngress.fromDTO(bridgeDTO, namespace));
-        if (!bridgeDeleted) {
+        String customerId = bridgeDTO.getCustomerId();
+        String bridgeId = bridgeDTO.getId();
+
+        // Deletion of the Namespace will cascade deletion of the BridgeIngress...
+        // In a *real* k8s environment the Namespace deletion should cascade into all children.
+        final String namespace = customerBridgeNamespaceProvider.resolveName(customerId, bridgeId);
```
robpblake, line 93: @manstis The alternative being that we move the delete of the `namespace` into the `BridgeIngressController.cleanUp()` method. That leaves us with a single, consistent code-path regardless of environment.

If we're going to rely on the cascade, then we definitely need to verify that it works as expected. In particular I'm looking at https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/shard-operator-integration-tests/src/test/java/com/redhat/service/smartevents/shard/operator/cucumber/steps/BridgeIngressSteps.java#L62-L69 

I suspect this will need updating anyway as it relies on the namespace being available.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/providers/CustomerBridgeNamespaceProvider.java
```diff
@@ -8,19 +8,21 @@
...
  * Service Interface for customer namespace handling.
  */
-public interface CustomerNamespaceProvider {
+public interface CustomerBridgeNamespaceProvider {
 
     /**
      * String format for the owned Bridge Namespace name.
      */
-    String NS_PREFIX_FORMAT = "ob-%s";
+    String NS_PREFIX_FORMAT = "ob-%s-%s";
```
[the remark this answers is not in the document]
robpblake: @manstis Could you also add the bridge Id as a `label` on the `namespace`, in the same way as we do for the `customer_id`. See the `createLabels` method in this class.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/BridgeIngressServiceImpl.java
robpblake: @manstis We also need to be careful here in that in our existing environments we have namespaces with multiple bridges in them. So this will delete all those bridges rather than just the single bridge. So we need to consider a strategy for the situation where there are multiple bridges in a `namespace`.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/providers/CustomerBridgeNamespaceProviderImpl.java
```diff
@@ -20,16 +21,16 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(CustomerBridgeNamespaceProviderImpl.class);
 
     @Inject
     KubernetesClient kubernetesClient;
 
     @Override
-    public Namespace fetchOrCreateCustomerNamespace(String customerId) {
-        final String name = this.resolveName(customerId);
+    public Namespace fetchOrCreateCustomerBridgeNamespace(String customerId, String bridgeId) {
+        final String name = this.resolveName(customerId, bridgeId);
```
robpblake, line 33: @manstis We need to consider existing bridges. Based upon this change we won't be able to update/delete existing bridges in our `dev` or `stable` environment as the `namespace` lookup will fail.

What I'd suggest is that we have a namespace look-up order of preference in here, something like:

- Check to see if a namespace exists in the new format, if yes, return it
- Check to see if a namespace exists in the old format, if yes, return it
- If no namespace exists, create it in the new format.

By my thinking that should mean that existing bridges continue to function as expected, with any new bridge being created with the new namespace format.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/providers/CustomerBridgeNamespaceProviderImpl.java
robpblake: @manstis We need to consider `namespaces` with existing bridges in them. Ideally we'd clean them up over time as well. 

Just walking through the code, one proposal that came to mind:

1. Rename this method `deleteNamespace`
2. Call it from the `BridgeIngressController.cleanup` method when a `BridgeIngress` is deleted, taking the `namespace` to delete from the `BridgeIngress` resource
3. Replace the logic in this method to ensure that the `namespace` is marked for deletion _if_:
   1. There are no `BridgeIngress` in the `namespace`, or 
   2. There are `BridgeIngress` in the `namespace` but they all have the `HasMetaData.isMarkedForDeletion` set to true
4. Delete the method annotated with `@Scheduled` in this class.

That should mean that over time we eventually clear-up the namespaces as we delete Bridges using the old format and it should work equally as well for `namespaces` created in the new format.

robpblake (changes requested):
@manstis We need to consider how we're going to handle existing bridges. At the minute this change will mean that we can no longer update or delete existing bridges.

robpblake:
> 

@manstis It's a good question, but we already create a `Broker` resource for each Bridge. If you take a look at this part of the code in the `BridgeIngressServiceImpl.fetchOrCreateBridgeIngressBroker` method:

```
KnativeBroker existing = kubernetesClient.resources(KnativeBroker.class)
                .inNamespace(bridgeIngress.getMetadata().getNamespace())
                .withName(bridgeIngress.getMetadata().getName())
                .get();
```

You'll see that it looks for a `Broker` resource in the `namespace` with the same `name` as the `Bridge` and if it doesn't exist, creates one.

So I don't think we're changing our deployment footprint at the minute, only the `namespace` in which the `Broker` resources are created. Behind the scenes, our `Broker` resource is still fulfilled by a single deployment running in the `knative-eventing` namespace.

cheers,

Rob

r00ta (commented):
This issue is going to be fixed by https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1065 because it contains the fix for https://github.com/knative-sandbox/eventing-kafka-broker/issues/2211

manstis:
Closing in favour of @r00ta [comment](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1070#pullrequestreview-1089759298).

PR #1091, opened 2022-09-02
MGDOBR-1037: Shard should verify hostname and SSL configuration of the Route exposed for a Bridge as part of the reconcile loop

manstis:
[https://issues.redhat.com/browse/MGDOBR-1037](https://issues.redhat.com/browse/MGDOBR-1037)

It's not clear what `RouteSpec` properties we should be considering in the comparison.

However, since `RouteSpec` overrides `equals(..)` delegate complete comparison to it.

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>

### DIFF
--- a/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/networking/OpenshiftNetworkingService.java
+++ b/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/networking/OpenshiftNetworkingService.java
@@ -52,7 +52,7 @@ public class OpenshiftNetworkingService implements NetworkingService {
                 .withName(bridgeIngress.getMetadata().getName())
                 .get();
 
-        if (existing == null || !expected.getSpec().getTo().getName().equals(existing.getSpec().getTo().getName())) {
+        if (existing == null || !existing.getSpec().equals(expected.getSpec())) {
             client.routes()
                     .inNamespace(service.getMetadata().getNamespace())
                     .withName(bridgeIngress.getMetadata().getName())
--- a/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/networking/OpenshiftNetworkingService.java
+++ b/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/networking/OpenshiftNetworkingService.java
@@ -21,6 +21,8 @@ import io.fabric8.openshift.api.model.RouteTargetReferenceBuilder;
 import io.fabric8.openshift.client.OpenShiftClient;
 import io.javaoperatorsdk.operator.processing.event.source.EventSource;
 
+import static com.redhat.service.smartevents.shard.operator.networking.OpenshiftRouteSpecMatchesHelper.matches;
+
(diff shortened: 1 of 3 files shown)


shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/networking/OpenshiftNetworkingService.java
```diff
@@ -52,7 +52,7 @@ public NetworkResource fetchOrCreateBrokerNetworkIngress(BridgeIngress bridgeIng
                 .withName(bridgeIngress.getMetadata().getName())
                 .get();
 
-        if (existing == null || !expected.getSpec().getTo().getName().equals(existing.getSpec().getTo().getName())) {
+        if (existing == null || !existing.getSpec().equals(expected.getSpec())) {
```
sutaakar, line 55: OpenShift adjusts Route spec definition with default parameters if not defied (i.e. host, to.weight, wildcardPolicy).
Such change will IMHO cause infinite reconciliation.
I think we need to check specific parameters.

manstis: @sutaakar Is there a definitive set of fields we should check?

We used to compare `To` as a whole (and not properties excluding `to.weight` so we have a problem already.

I really could not find anything to suggest what specific properties we should check.

Even `To` (as defined in our template) is, as you write, not entirely comparable.

sutaakar: In previous version I see that we compare just `to.name`.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/networking/OpenshiftRouteSpecMatchesHelper.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private OpenshiftRouteSpecMatchesHelper() {
+        //Static utility class
+    }
+
+    public static boolean matches(RouteSpec existing, RouteSpec expected) {
+        Function<RouteSpec, RouteSpec> spec = (RouteSpec source) -> source;
+        if (!Objects.equals(spec.apply(existing), spec.apply(expected))) {
+            return false;
+        }
```
sutaakar, line 19: This check returns false even when existing and expected don't match in a parameter which we don't check (i.e. `to.weight`).

manstis: @sutaakar Fixed.. I made a right mess of this change. Hopefully it's OK now.

sutaakar (approved):
Looks good to me

r00ta (approved):

Srihari1192 (approved):

PR #1126, opened 2022-09-08
MGDOBR-1035: Shard does not report Bridge as Failed if Knative Broker resource does not deploy

manstis:
[https://issues.redhat.com/browse/MGDOBR-1035](https://issues.redhat.com/browse/MGDOBR-1035)

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>

### DIFF
--- a/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeExecutorController.java
+++ b/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeExecutorController.java
@@ -31,7 +31,6 @@ import com.redhat.service.smartevents.shard.operator.resources.BridgeExecutor;
 import com.redhat.service.smartevents.shard.operator.resources.BridgeExecutorStatus;
 import com.redhat.service.smartevents.shard.operator.resources.Condition;
 import com.redhat.service.smartevents.shard.operator.resources.ConditionReasonConstants;
-import com.redhat.service.smartevents.shard.operator.resources.ConditionStatus;
 import com.redhat.service.smartevents.shard.operator.resources.ConditionTypeConstants;
 import com.redhat.service.smartevents.shard.operator.utils.DeploymentStatusUtils;
 import com.redhat.service.smartevents.shard.operator.utils.EventSourceFactory;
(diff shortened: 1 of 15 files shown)


shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeExecutorController.java
```diff
@@ -82,27 +98,66 @@ public List<EventSource> prepareEventSources(EventSourceContext<BridgeExecutor>
...
             bridgeExecutor.getStatus().markConditionTrue(ConditionTypeConstants.AUGMENTATION, ConditionReasonConstants.SECRETS_NOT_FOUND);
-            return UpdateControl.updateStatus(bridgeExecutor);
+            return UpdateControl.updateStatus(bridgeExecutor).rescheduleAfter(executorPollIntervalSeconds);
+        } else {
+            // Update "Last Transition Time", effectively resetting time-out.
+            status.getConditionByType(ConditionTypeConstants.AUGMENTATION).ifPresent(c -> {
+                if (Objects.equals(c.getReason(), ConditionReasonConstants.SECRETS_NOT_FOUND)) {
+                    status.markConditionTrue(ConditionTypeConstants.PROGRESSING, ConditionReasonConstants.SECRETS_FOUND);
+                }
+            });
```
r00ta, line 138: Unfortunately this snippet is not going to be used as the flow would be the following: 

- If `secret == null` -> update status in CRD and reschedule (in this case the changes are written on the CRD when you call `return UpdateControl.updateStatus(bridgeExecutor).rescheduleAfter(executorPollIntervalSeconds);`)
- if `secret != null` -> **in theory** update the status in the CRD and continue (the status is not really updated on the CRD, it is kept in memory)
 - if `!image.equals(bridgeExecutor.getSpec().getImage())` -> Overwrite the status in the CRD and reschedule (the changes are written on the CRD)
 - if `image.equals(bridgeExecutor.getSpec().getImage())` -> **in theory** Overwrite the status in the CRD and continue (the changes are NOT written on the CRD)
 
In short, all the snippets in the `else` branch of the steps have no effect at all. 

I think we should create specific `conditions` for each step, an example from knative is 
```
    - lastTransitionTime: '2022-06-23T13:34:35Z'
      status: 'True'
      type: Addressable
    - lastTransitionTime: '2022-06-23T13:34:32Z'
      reason: Config map knative-eventing/kafka-broker-brokers-triggers updated
      status: 'True'
      type: ConfigMapUpdated
    - lastTransitionTime: '2022-06-23T13:34:32Z'
      status: 'True'
      type: ConfigParsed
    - lastTransitionTime: '2022-08-28T00:36:46Z'
      status: 'True'
      type: DataPlaneAvailable
    - lastTransitionTime: '2022-09-09T04:50:02Z'
      status: 'True'
      type: ProbeSucceeded
    - lastTransitionTime: '2022-09-09T04:50:02Z'
      status: 'True'
      type: Ready
    - lastTransitionTime: '2022-06-23T13:34:32Z'
      reason: Topic ob-dev-brdg-0aabd1ed-20cd-4710-baef-4d93afbf98b3 created
      status: 'True'
      type: TopicReady
```

So instead of using `Augmenting` for everything, I think we should keep the status of each step in a specific condition to solve the problem I described above. Please correct me if I'm wrong :) wdyt?

r00ta (commented):
Just writing here some thoughts, we can have a call to design it together

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeIngressController.java
```diff
@@ -120,26 +173,45 @@ public UpdateControl<BridgeIngress> reconcile(BridgeIngress bridgeIngress, Conte
...
-        if (!bridgeIngress.getStatus().isReady()) {
-            bridgeIngress.getStatus().markConditionTrue(ConditionTypeConstants.READY);
-            bridgeIngress.getStatus().markConditionFalse(ConditionTypeConstants.AUGMENTATION);
+        if (!status.isReady()) {
+            status.markConditionTrue(ConditionTypeConstants.READY);
+            status.markConditionFalse(ConditionTypeConstants.AUGMENTATION);
+            status.markConditionFalse(ConditionTypeConstants.PROGRESSING);
             notifyManager(bridgeIngress, ManagedResourceStatus.READY);
-            return UpdateControl.updateStatus(bridgeIngress);
+            return UpdateControl.updateStatus(bridgeIngress).rescheduleAfter(ingressPollIntervalSeconds);
```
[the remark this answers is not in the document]
manstis: Removed.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeExecutorController.java
manstis: @r00ta I've refactored `BridgeIngressController` and removed use of the generic `AUGMENTING` condition to more specific counterparts. Could you have another look and let me know what you think? Two points you should note: (1) I realise the blocks setting `Condition`'s to `True` are only "in memory" but they are written to the CR when the following step first fails (or the reconciliation is complete). This way I avoid delaying reconciliation (by returning a Status Update early) just to set a `Condition` to `True`, (2) I only update a `Condition` value if it has changed (I could move this to the `CustomResourceStatus` when I make similar changes to `BridgeExecutorController`). This is to avoid updating the `lastTransitionTime` every reconcile loop due to the idempotency of k8s.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeIngressController.java
```diff
@@ -85,61 +100,116 @@ public List<EventSource> prepareEventSources(EventSourceContext<BridgeIngress> e
...
+        } else if (!status.isConditionTypeTrue(BridgeIngressStatus.SECRET_AVAILABLE)) {
+            status.markConditionTrue(BridgeIngressStatus.SECRET_AVAILABLE);
         }
 
-        ConfigMap configMap = bridgeIngressService.fetchOrCreateBridgeIngressConfigMap(bridgeIngress, secret);
-
         // Nothing to check for ConfigMap
+        ConfigMap configMap = bridgeIngressService.fetchOrCreateBridgeIngressConfigMap(bridgeIngress, secret);
+        if (!status.isConditionTypeTrue(BridgeIngressStatus.CONFIG_MAP_AVAILABLE)) {
```
manstis, line 140: I added this for "Nothing to check..." dependencies since the underlying k8s calls _could_ throw a `RuntimeException` and (for example) the `ConfigMap` could not be returned. With this block the Condition status for the step remains "Unknown". I think this makes it more consistent with the other dependencies.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeIngressController.java
```diff
@@ -85,61 +100,116 @@ public List<EventSource> prepareEventSources(EventSourceContext<BridgeIngress> e
...
+            status.markConditionTrue(ConditionTypeConstants.READY);
             notifyManager(bridgeIngress, ManagedResourceStatus.READY);
             return UpdateControl.updateStatus(bridgeIngress);
         }
 
         return UpdateControl.noUpdate();
     }
 
+    private boolean isTimedOut(BridgeIngressStatus status) {
+        Optional<Date> lastTransitionDate = status.getConditions()
```
manstis, line 198: Each `Condition` will have it's own "last transitioned date". I therefore get the _newest_ value when considering if provisioning has timed out as a whole.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeIngressController.java
```diff
@@ -85,61 +100,116 @@ public List<EventSource> prepareEventSources(EventSourceContext<BridgeIngress> e
...
+        }
 
         Secret secret = bridgeIngressService.fetchBridgeIngressSecret(bridgeIngress);
 
         if (secret == null) {
             LOGGER.info("Secrets for the BridgeIngress '{}' have been not created yet.", bridgeIngress.getMetadata().getName());
-            bridgeIngress.getStatus().markConditionFalse(ConditionTypeConstants.READY);
-            bridgeIngress.getStatus().markConditionTrue(ConditionTypeConstants.AUGMENTATION, ConditionReasonConstants.SECRETS_NOT_FOUND);
-            return UpdateControl.updateStatus(bridgeIngress);
+            if (!status.isConditionTypeFalse(ConditionTypeConstants.READY)) {
```
manstis, line 127: I only update the `Condition` if it has not already been done so otherwise the "last transitioned date" is updated each time the reconcile loop iterates.

manstis:
@robpblake @Srihari1192 This is ready for your reviews; following integration of @r00ta feedback.

We follow the same approach as taken by k8s `Deployment` (see JIRA for details).

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeExecutorController.java
```diff
@@ -53,6 +62,12 @@ public class BridgeExecutorController implements Reconciler<BridgeExecutor>,
 
     private static final Logger LOGGER = LoggerFactory.getLogger(BridgeExecutorController.class);
 
+    @ConfigProperty(name = "event-bridge.executor.deployment.timeout-seconds")
+    int executorTimeoutSeconds;
+
+    @ConfigProperty(name = "event-bridge.executor.poll-interval.seconds")
+    int executorPollIntervalSeconds;
+
```
r00ta, line 70: We might have to move these to the manager and pass them to the shard with the DTO for every resource. This is because in general the guideline is that the manager has to decide everything. But I suggest to move on with your current implementation now, let's create a ticket and move forward

r00ta (approved):
Looks good to me, good job :+1:

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeExecutorController.java
r00ta: Looks great :+1:

manstis:
@lucamolteni Since @robpblake is sunning himself I wonder if you'd mind taking a look.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeExecutorController.java
manstis: Crikey, I can understand that when the properties are _dynamic_ and vary for different instances (even `image` definition of the executor could be different when we roll-out newer versions).. but these properties are static system configuration. Surely we don't want to pass those from the `manager` to the `operator`?!?!

Srihari1192 (approved):

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeExecutorController.java
r00ta: Sorry I had the k8s `Deployment` in mind: that definition keeps the timeout in the spec. Your current implementation does not keep that in every CRD so you're right: it is a static config. 
Let's move forward then, in case we will need to move that parameter inside the CRD definition we will consider to move it to the DTO/manager

PR #1137, opened 2022-09-13
MGDOBR-1025 - shard-operator should be able to configure Istio

vaibhavjainwiz:
https://issues.redhat.com/browse/MGDOBR-1025

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>


r00ta (approved):

r00ta:
Looks like some pipeline have to be updated as well `error: the path "kustomize/overlays/ci/istio/gateway.yaml" does not exist`

r00ta (commented):
I think you are also missing to update `ClusterRole` for the new resources to be deployed by the shard

sutaakar (commented):
Looks good to me, the authorizations need to be addressed as Jacopo mentioned.
The CI may be working, needs to be checked by running a PR check from a dedicated upstream branch as the CI changes in this PR are not reflected.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/IstioSetupService.java
```diff
@@ -0,0 +1,121 @@
...
+
+        if (existing == null) {
+            Gateway expected = templateProvider.loadIstioGatewayTemplate();
+            expected.getMetadata().setName(ISTIO_GATEWAY_NAME);
+            expected.getMetadata().setNamespace(ISTIO_GATEWAY_NAMESPACE);
+
+            try {
+                kubernetesClient.resources(Gateway.class)
+                        .inNamespace(ISTIO_GATEWAY_NAMESPACE)
+                        .create(expected);
```
manstis, line 67: IDK whether we should improve the API use here (and elsewhere).

The `create(..)` method throws an `Exception` **if** the **request** encountered problems.

There is no guarantee the resource has actually been **created** by this method invocation; only that it was **requested**.

We could either poll here until a call to `get()` returns an instance of enhance `BridgeIngressController` to check its Istio depenencies are available... but simply calling `create(...)` in a "fire and forget" approach probably isn't best.

manstis (commented):
A comment.. but others more knowledgeable/influential can correct me.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/IstioSetupService.java
manstis: That said we never explicitly checked their existence when creating them with `yaml`..

sutaakar: AFAIK this is a synchronous operation, so if no exception is thrown then the resource should be created.
Though I am not sure how Istio actually process this resource, whether some service is deployed, or just Istio use it internally.

vaibhavjainwiz: These are Istio internal resource only.

vaibhavjainwiz:
I had created a new branch in upstream repo **MGDOBR-1025** and trigger the CI pipeline there to verify my changes.
https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1139

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/IstioSetupService.java
manstis: I asked the Java Operator SDK team about the `delete(...)` method a while ago; and they categorically advised it only _requests_ deletion and returns if the _request_ was successful. It is then for the applicable Controller to perform the deletion. I'd assume the same applies for `create(..)` i.e. the _request_ is successful and then the applicable Controller performs the provisioning; but if you know otherwise then fair point (although in our tests we call `create(..)` and then poll for things to become available).

sutaakar: `delete` is IMHO a different case due to finalizers. So delete request should just set the delete timestamp, the actual deletion is done by k8s based on finalizers content.

manstis: Great, thanks for this discussion @sutaakar.

I checked the source code and `create(OrReplace)(..)` does indeed return once the item is actually created.

manstis (approved):

sutaakar:
@vaibhavjainwiz Can you please update also https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/kustomize/overlays/ci/shard/cluster-role.yaml and https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/kustomize/base-openshift/shard/cluster-role.yaml ?

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/IstioSetupService.java
manstis: Hmmm. I'm still not convinced... but I won't argue my point (much more).

The `create(..)` method asks k8s to provision a resource with the proposed `spec`. A HTTP PUT is invoked to the k8s API Server, that persists the request and returns the requested object. At this point there has been no Controller execute to actually create anything. It is only when the usual CRD reconciliation loop runs is the object created. This is explained well [here](https://medium.com/nerd-for-tech/k8s-deployment-creation-what-happens-behind-the-scenes-4e7362aee295)... although I am extrapolating the article to cover all (Custom) Resources.

manstis: [This](https://stackoverflow.com/questions/61267334/kubectl-create-apply-synchronously-until-each-object-is-created) seems to confirm my understanding too.

vaibhavjainwiz:
> @vaibhavjainwiz Can you please update also https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/kustomize/overlays/ci/shard/cluster-role.yaml and https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/kustomize/base-openshift/shard/cluster-role.yaml ?

done

vaibhavjainwiz:
As tested in https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1142, CI is now green.

sutaakar (approved):

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/IstioSetupService.java
```diff
@@ -0,0 +1,121 @@
...
+            } catch (RuntimeException e) {
+                LOGGER.error(
+                        "Failed to create Istio Gateway resource. Please make sure it was properly deployed. The application keeps running due to https://issues.redhat.com/browse/MGDOBR-940 but the functionalitis are compromised.");
```
danielezonca, line 70: We should be able to use some config to make this failure fatal in the "normal" scenario, in general the operator should stop in this case, it is just with CVP that we have a corner case.
@vaibhavjainwiz 
Wdyt?
I am thinking at some config param to pass to the operator

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/IstioSetupService.java
```diff
@@ -0,0 +1,121 @@
...
+            expected.getMetadata().setName(ISTIO_VIRTUAL_SERVICE_NAME);
+            expected.getMetadata().setNamespace(ISTIO_VIRTUAL_SERVICE_NAMESPACE);
+            expected.getSpec().getGateways().set(0, ISTIO_GATEWAY_NAME);
+            try {
+                kubernetesClient.resources(VirtualService.class)
+                        .inNamespace(ISTIO_VIRTUAL_SERVICE_NAMESPACE)
+                        .create(expected);
+            } catch (RuntimeException e) {
+                LOGGER.error(
+                        "Failed to create Istio Virtual Service resource. Please make sure it was properly deployed. The application keeps running due to https://issues.redhat.com/browse/MGDOBR-940 but the functionalitis are compromised.");
```
danielezonca, line 93: Same here

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/IstioSetupService.java
```diff
@@ -0,0 +1,121 @@
...
+            expected.getMetadata().setNamespace(JWT_REQUEST_AUTHENTICATION_NAMESPACE);
+            expected.getSpec().getJwtRules().get(0).setIssuer(jwtIssuer);
+            expected.getSpec().getJwtRules().get(0).setJwksUri(jwksUri);
+            try {
+                kubernetesClient.resources(RequestAuthentication.class)
+                        .inNamespace(JWT_REQUEST_AUTHENTICATION_NAMESPACE)
+                        .create(expected);
+            } catch (RuntimeException e) {
+                LOGGER.error(
+                        "Failed to create JWT Request Authentication resource. Please make sure it was properly deployed. The application keeps running due to https://issues.redhat.com/browse/MGDOBR-940 but the functionalitis are compromised.");
```
danielezonca, line 117: And here

shard-operator/src/main/resources/application.properties
```diff
@@ -42,5 +42,5 @@ quarkus.log.console.json.fields.sequence.enabled=false
 event-bridge.secret.name=addon-smart-events-operator-parameters
 # Name of the secret that needs to be imported on operator deployment.
 quarkus.kubernetes.env.secrets=${event-bridge.secret.name}
-
-
+event-bridge.istio.jwt.issuer=https://sso.redhat.com/auth/realms/redhat-external
+event-bridge.istio.jwt.jwksUri=https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/certs
```
danielezonca, line 46: I would use a param instead of hardcoded value like we do for
`event-bridge.sso.auth-server-url=${EVENT_BRIDGE_SSO_URL:http://${minikubeip}:30007/auth/realms/event-bridge-fm}`

Wdyt?

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/IstioSetupService.java
vaibhavjainwiz: There is no easy way to provide special configuration in CVP pipeline.

I am already in discussion with CVP team to check if it is possible to create a secret in CVP pipeline before start our operator. If it works then I could add this extra parameter in secret which would be mounted on our operator deployment.

shard-operator/src/main/resources/application.properties
vaibhavjainwiz: done

PR #1201, opened 2022-10-05
MGDOBR-746 Test performance of Bridge with error handler creation a…

sutaakar:
…nd deletion

[MGDOBR-746](https://issues.redhat.com/browse/MGDOBR-746)

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>

### DIFF
--- a/integration-tests/src/test/resources/performance-features/continuously-create-bridges.feature
+++ b/integration-tests/src/test/resources/performance-features/continuously-create-bridges.feature
@@ -127,3 +127,137 @@ Feature: Continuous create Bridges performance tests
     Examples:
       | users | shared-connections |
       | 1     | 1                  |
+
+  @continuously-create-bridges-with-error-handler
+  Scenario Outline: Continuously create Bridges with error handler by <users> users
+    When run benchmark with content:
+      """text/vnd.yaml
+      name: rhose-continuously-create-bridges-with-error-handler
+      http:
+      - host: ${env.event-bridge.manager.url}
+        name: manager
+        sharedConnections: <shared-connections>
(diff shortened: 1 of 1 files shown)


sutaakar:
It is mostly copy pasted previous scenario of creating Bridges, the difference is basically only the Bridge creation request content.
Copy pasting is far from ideal. Was thinking about alternative and the only other option would be to provide the Bridge request body as a parameter. However as Gherkin doesn't support multiline parameters the Bridge body would have to be provided on single line, making it hard to read.
So it is a tradeoff. I don't have much preference on which approach may be better.
@afalhambra what do you think?

afalhambra:
> It is mostly copy pasted previous scenario of creating Bridges, the difference is basically only the Bridge creation request content. Copy pasting is far from ideal. Was thinking about alternative and the only other option would be to provide the Bridge request body as a parameter. However as Gherkin doesn't support multiline parameters the Bridge body would have to be provided on single line, making it hard to read. So it is a tradeoff. I don't have much preference on which approach may be better. @afalhambra what do you think?

Well, I don't have a strong opinion either. 
At the moment, this scenario might be shared and reused by 2 tests, so it's not that much, I would be much worried if we had to duplicate the same for several test cases... so in that case, then a change would require to change it all of them... but at the moment, I'd say it's not that much, so maybe we can leave it as is, and if later the amount of cases changes, we could go for the path of sending the Bridge request body as parameter in a single line.

afalhambra (approved):
nice job!

jiripetrlik (approved):

PR #1262, opened 2022-11-01
MGDOBR-1180 - Always export junit artifacts

jiripetrlik:
Forgot to export junits also for failed runs.

### DIFF
--- a/.github/workflows/dev-cluster-check.yml
+++ b/.github/workflows/dev-cluster-check.yml
@@ -83,6 +83,7 @@ jobs:
           working-directory: integration-tests
       - name: Archive junit test results
         uses: actions/upload-artifact@v3
+        if: ${{ always() }}
         with:
           name: junit-test-results
           path: |
--- a/.github/workflows/production-cluster-check.yml
+++ b/.github/workflows/production-cluster-check.yml
@@ -83,6 +83,7 @@ jobs:
           working-directory: integration-tests
       - name: Archive junit test results
         uses: actions/upload-artifact@v3
+        if: ${{ always() }}
         with:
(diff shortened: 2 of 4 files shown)


.github/workflows/dev-cluster-check.yml
```diff
@@ -83,6 +83,7 @@ jobs:
           working-directory: integration-tests
       - name: Archive junit test results
         uses: actions/upload-artifact@v3
+        if: ${{ always() }}
```
afalhambra, line 86: I think we don't need curly braces here, wdyt?

sutaakar: Seems, so, though it is used in examples
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif

sutaakar (approved):

.github/workflows/dev-cluster-check.yml
afalhambra: we have a mix in our workflows, some with curly [braces](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/.github/workflows/dev-cluster-check.yml#L109) and some without [them](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/.github/workflows/dev-cluster-check.yml#L117)

afalhambra (approved):

.github/workflows/dev-cluster-check.yml
jiripetrlik: Good point. I can look at it and try to unify this in the separate PR.

PR #1273, opened 2022-11-07
MGDOBR-1180 - Script for test results import to Report Portal

jiripetrlik:
Import test results from Github workflows to Report Portal.

### DIFF
--- a/dev/bin/report-portal/report-portal-import-from-workflows.sh
+++ b/dev/bin/report-portal/report-portal-import-from-workflows.sh
@@ -1,0 +1,118 @@
+#!/bin/bash
+set -e
+
+# Import test result from Github workflow to Report Portal.
+#
+# Usage: report-portal-import-from-workflows.sh repo_owner, repository, workflow, launch_name
+# Expected environment variables
+# GITHUB_TOKEN ... token with access to workflow run artifacts
+# REPORT_PORTAL_URL ... Report Portal URL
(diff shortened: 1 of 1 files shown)


dev/bin/report-portal/report-portal-import-from-workflows.sh
```diff
@@ -0,0 +1,118 @@
...
+
+check_env_variable_exists() {
+    env_variable_name=$1
+    if [[ -z "${!env_variable_name}" ]]; then
+        echo "Error: $env_variable_name environmental variable must be set" 1>&2;
+        exit 1
+    fi
+}
+
+import_junit() {
```
winklerm, line 21: Is there any specific reason why not to use Data Router for importing the junit results? 

I think it is totally fine to start what is implemented here, but the DR is supposed to help us to avoid implementing the import logic by ourselves, so it might be a good practice in future?

winklerm (approved):
Thanks, looks good to me - just wondering about the Data Router usage for future - please see my comment, thanks!

dev/bin/report-portal/report-portal-import-from-workflows.sh
jiripetrlik: Yes, I agree with improving to Data Router in the future. It should not be hard if it can import junits.

sutaakar (approved):

Srihari1192 (approved):

PR #1334, opened 2022-11-21
Mgdobr 1144

vaibhavjainwiz:

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>


manstis:
I'll paste my comments here too:

1) Status updates

When/where do you signal to the Manager that a Bridge/Processor is being "provisioned"? 

I see you prematurely[*] [record](https://github.com/vaibhavjainwiz/sandbox/blob/MGDOBR-1144/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/reconcilers/BridgeIngressReconciler.java#L47) the status of individual components; but not the Bridge/Processor as a whole.

[*] I say it is premature as you're setting "success" when all you've really done is ask k8s to create/update/delete the resource in [DeltaProcessorServiceImpl](https://github.com/vaibhavjainwiz/sandbox/blob/MGDOBR-1144/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/DeltaProcessorServiceImpl.java#L32-L34) with no guarantee that it was either successful or failed.

It is issues with this mechanism that we are trying to improve on for v2 so I think it is very important you PoC demonstrates this.

Am I missing something?

2) Parallelism

I see you've refactored the Bridge/Processor Controllers to handle each sub-resource in parallel. Previously these were handled [sequentially](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/shard-operator-parent/shard-operator-v1/src/main/java/com/redhat/service/smartevents/shard/operator/v1/controllers/BridgeIngressController.java#L145-L161).

e.g. Your code will attempt to reconcile Kafka Secrets, ConfigMap, Broker, Path, Route all at the same time (each sub-reconciler calls into the delta processor asking k8s to provision the CR).

Are you confident this new approach is OK? I foresee issues with use of [path](https://github.com/vaibhavjainwiz/sandbox/blob/MGDOBR-1144/shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/controllers/BridgeIngressController.java#L98-L100) before it is available but perhaps I am missing something. 

3) Timeout

We added the ability to "timeout" provisioning of a Bridge/Processor if the operation took more than X seconds.

I do not see this feature being implemented in your refactoring.

Is it dropped from v2?

4) Failed k8s reconciliation

I see you have removed support for signalling failure to reconcile a Bridge/Processor by k8s to the Manager.

I assume this is simply to be added back in your final refactoring.

5) Deletion

I see you have removed support for deletion of Bridge/Processor from your PoC.

I assume this is simply to be added back in your final refactoring.

shard-operator/src/main/java/com/redhat/service/smartevents/shard/operator/reconcilers/BridgeExecutorDeploymentReconciler.java
```diff
@@ -0,0 +1,74 @@
...
+    public void reconcile(BridgeExecutor bridgeExecutor){
+
+        try{
+            List<Deployment> requestResource = createRequiredResources(bridgeExecutor);
+
+            List<Deployment> deployedResources = fetchDeployedResources(bridgeExecutor);
+
+            processDelta(requestResource, deployedResources);
+
+            validateDeploymentIsReady(bridgeExecutor);
```
manstis, line 43: @vaibhavjainwiz I have been considering this _solution_ for premature success notifications.

Unfortunately this still has the potential to incorrectly notify intermittent failure statuses to the `manager` when the User updates a failed resource. Consider the following: 

1. A `Deployment` has its Conditions in a _failed_ state for whatever reason; and the Processor is marked as failed to the User.
2. The User updates the Processor definition and the Operator's reconcile loop is eventually re-entered.
3. When L#41 is invoked you _signal_ to k8s to update the `Deployment`. This update will be handled by another [Operator](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/deployment_controller.go) somewhere at some point in time.
4. You immediately retrieve the `Deployment`'s status on L#43 which _could be before_ the Operator has had sufficient time to attempt to re-provision the revised `Deployment`.
5. This could lead to the `validateDeploymentIsReady(..)` throwing `ReconciliationException` and signalling to the `manager` that the Update has failed.
6. When the reconcile loop runs again the `Deployment` Operator may have re-provisioned the resource and L#43 retives the updated `Deployment` status; no `ReconciliationException` is thrown and the `manager` signalled that everything was successful.

This behaviour is no different to v1 but something we was hoping to improve on with v2.

I think you ideally do not want to retrieve the `Deployment` status until the next reconcile loop iteration triggered by the `Deployment`'s `EventSource`. Reading the "[Patterns and Best Practices](https://javaoperatorsdk.io/docs/patterns-best-practices#eventsources-and-caching)" suggests both v1 and v2 implementations are flawed in that they should not be "...Reading the actual state of a resource from the Kubernetes API Server directly all the time..." but should respond to the `EventSource` triggers.

With the code split you have proposed; it should be reasonably simple to have _your_ reconcilers respond only to `EventSource` triggers and not on every reconcile loop. It may be something worth investigating and considering.

r00ta:
Hi @vaibhavjainwiz , I'm closing this for the time being. Your work is going to be splitted into multiple PRs so I guess this one is not necessary anymore

PR #1344, opened 2022-11-23
[no Jira] Add run permissions to manager-run script

jiripetrlik:
For some reason run permissions for manager-run script were removed. Adding it back.


manstis (approved):
Brilliant.. I thought I'd committed that in a PR somewhere.

But great stuff for calling it our explicitly :1st_place_medal:

manstis:
> Brilliant.. I thought I'd committed that in a PR somewhere.

This is not an admission of guilt. Git Blame is for that.

sutaakar (approved):

PR #1350, opened 2022-11-24
MGDOBR-1270: add v2 database and POJOs

r00ta:
[MGDOBR-1270](https://issues.redhat.com/browse/MGDOBR-1270)

## PR owner checklist

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>


manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
```diff
@@ -0,0 +1,75 @@
...
+    region             varchar(255) NOT NULL,
+
+    endpoint           varchar(255),
+    shard_id           varchar(255) NOT NULL,
+
+    submitted_at       timestamp NOT NULL,
+    published_at       timestamp,
+
+    error_id           integer,
+    error_uuid         varchar(255),
```
manstis, line 20: Personally, now we have `Condition`'s I'd be tempted to kill this off.

It was used to propagate "(failure) reasons" to the User and is largely constructed from `Condition` messages.

@robpblake @lucamolteni WDYT?

manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
```diff
@@ -0,0 +1,75 @@
...
+    owner              varchar(255) NOT NULL,
+
+    submitted_at       timestamp NOT NULL,
+    published_at       timestamp,
+
+    version            integer NOT NULL default 0,
+    generation         integer NOT NULL default 0,
+
+    error_id           integer,
+    error_uuid         varchar(255),
```
manstis, line 44: Same comment.

manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
```diff
@@ -0,0 +1,75 @@
...
+    type                 varchar(255) NOT NULL,
+    requested_at         timestamp NOT NULL,
+
+    managed_resource_id  varchar(255)
+);
+
+CREATE INDEX managed_resource_id_index ON OPERATION (managed_resource_id);
+
+CREATE TABLE CONDITION (
+    id                     varchar(255) NOT NULL PRIMARY KEY,
```
manstis, line 63: Where do you get the field list from?

MGDOBR-1186 defines: `id`, `type`, `reason`, `message`, `lastUpdateTime`, `lastTransitionTime` and `status`.

You have `errorCode` and `component` too.

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/persistence/models/Condition.java
```diff
@@ -0,0 +1,113 @@
...
+import javax.persistence.Entity;
+import javax.persistence.FetchType;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "CONDITION")
+public class Condition {
```
manstis, line 16: Should we have `equals(..)` and `hashCode()` methods?

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/persistence/models/Operation.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.UUID;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "OPERATION")
+public class Operation {
```
manstis, line 13: Should we have `equals(..)` and `hashCode()` methods?

manstis (commented):
Generally LGTM :+1:

manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
```diff
@@ -0,0 +1,75 @@
...
+    error_id           integer,
+    error_uuid         varchar(255),
+
+    bridge_id          varchar(255) NOT NULL,
+
+    unique (bridge_id, name),
+    constraint fk_bridge foreign key (bridge_id) references BRIDGE_V2 (id)
+);
+
+CREATE TABLE OPERATION (
```
danielezonca, line 52: I don't understand the goal of this `OPERATION` table 🤔 
As far as I see this is tied to `CONDITION` but a condition can be created directly by the shard operator so when this operation is created/linked?

manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
```diff
@@ -0,0 +1,75 @@
...
+    id                   varchar(255)   NOT NULL PRIMARY KEY,
+    type                 varchar(255)   NOT NULL,
+    requested_at         timestamp      NOT NULL,
+
+    managed_resource_id  varchar(255)
+);
+
+CREATE INDEX managed_resource_id_index ON OPERATION (managed_resource_id);
+
+CREATE TABLE CONDITION (
```
danielezonca, line 62: How does `CONDITION` is linked with the resources? As far as I see there are no foreign keys

manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
r00ta: See 
Screenshot from 2022-11-24 11-01-11

for the DB schema

r00ta: The `Operation` aims to keep track of the user's request (like `create`, `update`, `delete`) and the set of conditions is linked to this `Operation`

manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
r00ta: There is the `managed_resource_id` that is going to be used to lookup the Bridge or the Processor

manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
danielezonca: Does that mean that every condition returned by the operator is then saved in the DB spitting the status section in the different conditions "linked" by the `Operation`?
What if the operator returns a "failed" condition that is not triggered by any user input (i.e. a system error or some external conditions that have changed and now the resource is not working anymore)?

danielezonca: Thinking at what you wrote I have also another questions, do we plan to cleanup the `Operation` over time?
I.e. a processor is:
- created
- updated
- updated
- deleted

Do we plan to keep the whole "history"? I expect we only need the current/latest

manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
r00ta: Hey @manstis , the `Condition` in the CRD defines the `errorCode` so given that we said to have the same model I've reported it here (see https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/628d0eff3b3a28e64bec7351c8cefe4877660404/shard-operator-parent/shard-operator-core/src/main/java/com/redhat/service/smartevents/shard/operator/core/resources/Condition.java#L50). 

Regarding the `component`, I thought it was needed to store which is the `component`  (`controlPlane` or `dataPlane` - actually I'm going to refactor it with an ENUM in the java pojo, much better I think). wdyt?

manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
r00ta: The idea is that we store a new `Condition` for every dependency deployed by the control plane (as well as if it fails), and we store all the conditions returned by the data plane. The top-level status of the resource is then calculated at runtime when the user interacts with the REST API (for example, if all the conditions are `Ready`, then the resource is `Ready`. If there is at least a `Failed` condition, then the resource is in the `Failed` state. @manstis prepared a document to describe the logic that we are going to implement for this). 

Currently we keep a 1:1 mapping between the managed resource and the operation - meaning we keep just the latest request of the user. But we thought this data model could be easy to be extended to keep the entire history of the user's requests and their results

manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
r00ta: Gone :rocket:

manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
r00ta: Gone :rocket:

manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
manstis: Thanks for the explanation @r00ta.

I forgot about `errorCode` (being `BridgeError` code.. IDK if renaming it for v2 would make it more obvious).

An enumeration sounds much better for `component` :+1:

manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
manstis: See https://docs.google.com/document/d/1TM_ev8RV_Vft9YusIVDRe0zm7aHE2G-VHszFxYOpBRs/edit

manstis: We were not keen on the notion of keeping an _operation_ i.e. something done to a resource; on the resource itself.

An `Operation` is applied to a `Resource` by `Work`. The outcome of the `Operation` is a status (`Condition`'s).

manager-parent/manager-core/src/main/resources/db/migration/0.2/V0.2.1__create_v2_db.sql
r00ta: wopppssss

r00ta: Gone now :rocket:

shard-operator-parent/shard-operator-core/src/main/resources/application.properties
```diff
@@ -28,9 +28,9 @@ quarkus.log.console.json.fields.sequence.enabled=false
 event-bridge.secret.name=addon-smart-events-operator-parameters
 
 # Metric names
-rhose.metrics-name.operator.operation-total-count=managed_services_api_rhose_operator_operation_count
-rhose.metrics-name.operator.operation-success-total-count=managed_services_api_rhose_operator_operation_success_count
-rhose.metrics-name.operator.operation-failure-total-count=managed_services_api_rhose_operator_operation_failure_count
+rhose.metrics-name.operator.operationType-total-count=managed_services_api_rhose_operator_operation_count
```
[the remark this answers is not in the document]
r00ta: intellij renamed these properties by mistake. Reverted

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/persistence/models/Operation.java
```diff
@@ -0,0 +1,95 @@
...
+    @Enumerated(EnumType.STRING)
+    private OperationType type;
+
+    @Column(name = "requested_at", columnDefinition = "TIMESTAMP", nullable = false)
+    private ZonedDateTime requestedAt;
+
+    @Column(name = "managed_resource_id", nullable = false)
+    private String managedResourceId;
+
+    @OneToMany(fetch = FetchType.LAZY, mappedBy = "operation")
```
danielezonca, line 36: I'm not sure if `LAZY` is the correct here: if I have understood correctly the "final" `ManagedResourceStatus` is going to be dynamically computed based on the conditions so essentially every time we fetch a resource we will have this:
`Bridge`/`Processor` -> `Operation` ->* `Condition`

So we can consider `EAGER` here (even if it has other drawbacks) 🤔

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/persistence/models/Bridge.java
```diff
@@ -0,0 +1,131 @@
...
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Table;
+import javax.persistence.UniqueConstraint;
+
+import com.redhat.service.smartevents.manager.core.models.ManagedResource;
+
+@Entity(name = "Bridge_V2")
+@Table(name = "BRIDGE_V2", uniqueConstraints = { @UniqueConstraint(columnNames = { "name", "customer_id" }) })
+public class Bridge extends ManagedResource {
```
danielezonca, line 14: I would expect a `Operation` field here to easily reach `Condition`

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/persistence/models/Processor.java
```diff
@@ -0,0 +1,95 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.redhat.service.smartevents.manager.core.models.ManagedResource;
+
+import io.quarkiverse.hibernate.types.json.JsonBinaryType;
+import io.quarkiverse.hibernate.types.json.JsonTypes;
+
+@Entity(name = "Processor_V2")
+@TypeDef(name = JsonTypes.JSON_BIN, typeClass = JsonBinaryType.class)
+@Table(name = "PROCESSOR_V2", uniqueConstraints = { @UniqueConstraint(columnNames = { "name", "bridge_id" }) })
+public class Processor extends ManagedResource {
```
danielezonca, line 25: Similar comment here about `Operation`

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/persistence/models/Bridge.java
manstis: We explicitly decided against that and instead will have something like an `OperationService` to handle interaction with operations and conditions. Our reason being two fold: (1) We did not want both a `bridge_id` and `processor_id` FK on the `Operation` table as a single `managed_resource_id` and index is all that is technically needed. This means we cannot have Hibernate handle the relationship. (2) `Operation` is a standalone concept that is applied to a `ManagedResource` by `Work`. It is not a property of the `ManagedResource`.

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/persistence/models/Processor.java
manstis: Same as above.

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/persistence/models/Operation.java
manstis: @danielezonca @r00ta and myself discussed this. When we get the list of ALL resources to return to the User we will probably use a different HQL query to retrieve the batch resultset eagerly populated to place the burden of joins on the database.

manstis (changes requested):
@r00ta I did some basic testing of the `Operation` and `Condition` DTOs to check the bi-directional relationship. In short, at best, developers need to programmatically set properties on both ends of the relationship in order for it to be visible but at worse causes a Hibernate Circular Reference issue. I attach my code for you to see.

[DAOtests.zip](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/files/10091772/DAOtests.zip)

I tihnk we probably need to move to the single-direction relationship of `Operation.getConditions()` as IMO that's the most likely use case.

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/persistence/models/Operation.java
lucamolteni: If we're going to use the HQL query to eager fetch the data let's remove the `OneToMany` mapping altogether, it's not mandatory

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/persistence/models/Bridge.java
r00ta: @danielezonca @manstis I've implemented an alternative embedding the Operation in the Bridge/Processor. Check it out https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1354

danielezonca: @r00ta 
Thanks for the PR, I personally prefer #1354 so that we avoid at least the additional join. We still have the "fake" join between `Condition` and the managed resource but I don't have any suggestion in particular. If this is going to be a problem we can always have multiple foreign keys so I'm personally in favour to proceed with #1354

r00ta: agreed

r00ta:
Closing in favor of https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1354

PR #1408, opened 2022-12-08
[No JIRA] Update run-local-tests script for shard operator

jiripetrlik:


sjamboro (approved):

sutaakar:
@jiripetrlik What is this PR needed for?

jiripetrlik:
@sutaakar 
It is to fix the script when you are running tests locally. I've found out that without this it doesn't work.

sutaakar (approved):

PR #1438, opened 2022-12-16
[MGDOBR-1267] - Update Control Plane with ManagedBridge status

vaibhavjainwiz:
https://issues.redhat.com/browse/MGDOBR-1267

## PR owner checklist

- [ ] Your commit messages are clear and reference the JIRA issue e.g: "[MGDOBR-XXX] - $clear_explanation_of_what_you_did"

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>


shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
```diff
@@ -72,100 +65,72 @@ public class ManagedBridgeController implements Reconciler<ManagedBridge>,
...
-        if (secret == null) {
-            LOGGER.info("The Secret for the ManagedBridge with id '{}' has been not created yet.",
-                    managedBridge.getSpec().getId());
-            if (!status.isConditionTypeFalse(ConditionTypeConstants.READY)) {
-                status.markConditionFalse(ConditionTypeConstants.READY);
-            }
-            if (!status.isConditionTypeFalse(ManagedBridgeStatus.SECRET_AVAILABLE)) {
+            Secret secret = managedBridgeService.fetchBridgeSecret(managedBridge);
+            if (secret == null) {
                 status.markConditionFalse(ManagedBridgeStatus.SECRET_AVAILABLE);
```
manstis, line 83: A general comment (before I review any more).

We had proposed to use `String` constants in [`DefaultConditions`](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/infra-parent/infra-v2/src/main/java/com/redhat/service/smartevents/infra/v2/api/models/DefaultConditions.java) in `infra-v2` so that we can reference the same constants throughout the entire codebase. WDYT? (Feel free to rename constants prefixed `CP_` and their values as you see fit).

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
```diff
@@ -72,100 +65,72 @@ public class ManagedBridgeController implements Reconciler<ManagedBridge>,
     @Inject
     OperatorMetricsService metricsService;
 
+    @Inject
+    ManagerClient managerClient;
```
danielezonca, line 69: I think with v2 we should avoid the usage of `ManagerClient` inside the controllers: what if the manager is not reachable? Do we loose the update? K8s update succeed so it is going to have locally the resource updated but the status on the Manager remains wrong.
I would move also the shard -> manager logic in a scheduled method that periodically send back ~everything (aka all the statuses) to the manager. We can do it smarter with some delta but I would start simple with a full update.
In this way we also preserve the single write principle that makes the system reliable.

@robpblake 
Wdyt?

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/ManagedBridgeStatus.java
```diff
@@ -22,19 +23,21 @@ public class ManagedBridgeStatus extends CustomResourceStatus {
...
-        {
-            add(new Condition(ConditionTypeConstants.READY, ConditionStatus.Unknown));
-            add(new Condition(SECRET_AVAILABLE, ConditionStatus.Unknown));
-            add(new Condition(CONFIG_MAP_AVAILABLE, ConditionStatus.Unknown));
-            add(new Condition(KNATIVE_BROKER_AVAILABLE, ConditionStatus.Unknown));
-            add(new Condition(AUTHORISATION_POLICY_AVAILABLE, ConditionStatus.Unknown));
-            add(new Condition(NETWORK_RESOURCE_AVAILABLE, ConditionStatus.Unknown));
-        }
-    };
+    private static Set<Condition> getInitialConditions() {
```
danielezonca, line 26: As far as I see the existing code uses a constant while now every instance of `ManagedBridgeStatus` has a different instance of `Set` of conditions. Can you please clarify the reason of this change?

danielezonca (commented):
Overall looks good to me, just some comments

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
```diff
@@ -72,100 +65,72 @@ public class ManagedBridgeController implements Reconciler<ManagedBridge>,
...
-            LOGGER.info("The Secret for the ManagedBridge with id '{}' has been not created yet.",
-                    managedBridge.getSpec().getId());
-            if (!status.isConditionTypeFalse(ConditionTypeConstants.READY)) {
-                status.markConditionFalse(ConditionTypeConstants.READY);
-            }
-            if (!status.isConditionTypeFalse(ManagedBridgeStatus.SECRET_AVAILABLE)) {
+            Secret secret = managedBridgeService.fetchBridgeSecret(managedBridge);
+            if (secret == null) {
                 status.markConditionFalse(ManagedBridgeStatus.SECRET_AVAILABLE);
+                throw new ReconciliationException();
```
r00ta, line 84: Not sure what you want to do with this `throw new ReconciliationException()`. I have few comments: 
1) The fact that the a Deployment or a Broker is not ready is not an exception: it is actually expected that these resources are not ready immediately like a ConfigMap. As this is not an Exception, our code should handle it properly with no Exceptions - unless you want to use Exception to control the flow, which I would discourage as well. 
2) more importantly, you are updating the conditions and then you `throw` the exception: the CRD on k8s is not going to be updated as you would expect. 
3) The operator is going to raise tons of `Error during event processing ExecutionScope...` exceptions
4) After a given number of retries (configurable in the operator sdk), the operator sdk will raise a `Exhausted retries for ExecutionScope` exception and will stop retrying.

r00ta (dismissed):

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
danielezonca: As far as I see the design choice of that exception is like a "goto" to stop execution and reschedule the loop in a single `catch`. The exception is [thrown inside a `try/catch`](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/62a73d022687997341dc3a0b9d990b3fe2611757/shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java#L124) and not "leaked" outside that method

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/ManagedBridgeStatus.java
vaibhavjainwiz: The issue I realize while writing testcase is that we only have single Condition Set instance which is bounded to every ManagedBridgeStatus instance so any change in condition of one ManagedBridgeStatus reflects to all other ManagedBridgeStatus within memory.

danielezonca: Ok thanks for the clarification, so essentially that Set is going to be modified so it cannot be the same instance 👍

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
r00ta: Thanks @danielezonca , I missed that. I'm personally very against using Exceptions for flow control: they are expensive and the GOTO-like effect is something that should be avoided in general.

vaibhavjainwiz: With exception I could write handling of failure in single place inside catch block.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
```diff
@@ -219,4 +168,12 @@ private String extractBrokerPath(KnativeBroker broker) {
             return null;
         }
     }
+
+    private void notifyManager(ManagedBridge managedBridge) {
```
robpblake, line 172: @vaibhavjainwiz See discussion about removing callback to Control Plane from Controllers and simply doing it on a timed loop in the `ManagerBridgeSyncService`

shard-operator-parent/shard-operator-v2/src/test/java/com/redhat/service/smartevents/shard/operator/v2/ManagerClientTest.java
```diff
@@ -55,4 +66,21 @@ public void TestFetchProcessorsToDeployOrDelete() throws JsonProcessingException
 
         assertThat(managerClient.fetchProcessorsToDeployOrDelete().await().atMost(Duration.ofSeconds(10)).size()).isEqualTo(1);
     }
+
+    @Test
+    public void TestNotifyBridgeStatus() {
```
robpblake, line 71: @vaibhavjainwiz nitpick, but could you please using camel case for method names. Cheers!

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
vaibhavjainwiz: done

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/ManagedBridgeStatus.java
```diff
@@ -22,19 +23,21 @@ public class ManagedBridgeStatus extends CustomResourceStatus {
...
+        return new HashSet<>() {
+            {
+                add(new Condition(ConditionTypeConstants.READY, ConditionStatus.Unknown));
+                add(new Condition(SECRET_AVAILABLE, ConditionStatus.Unknown));
+                add(new Condition(CONFIG_MAP_AVAILABLE, ConditionStatus.Unknown));
+                add(new Condition(KNATIVE_BROKER_AVAILABLE, ConditionStatus.Unknown));
+                add(new Condition(AUTHORISATION_POLICY_AVAILABLE, ConditionStatus.Unknown));
+                add(new Condition(NETWORK_RESOURCE_AVAILABLE, ConditionStatus.Unknown));
+            }
+        };
```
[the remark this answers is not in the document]
vaibhavjainwiz: done

shard-operator-parent/shard-operator-v2/src/test/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeServiceImplTest.java
```diff
@@ -135,6 +138,20 @@ public void managedBridgeCreationTriggersController() {
                 });
     }
 
+    @Test
+    public void IsBridgeStatusChange() {
```
robpblake, line 142: @vaibhavjainwiz Can we have the opposite test please that verifies that if all conditions on a `ManagedBridge` are the same, then this method returns `false` as it's a key part of ensuring we only send updates to the Control Plane when we need to.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
danielezonca: If we do perform the refactor [mentioned here](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1438#discussion_r1050500887) to avoid manager notify directly in the loop we don't need an exception just to centralise this logic, we can still keep logic in a single method like
```java
private UpdateControl<ManagedBridge> handleFailure(ManagedBridge managedBridge) {
    return UpdateControl.updateStatus(managedBridge).rescheduleAfter(reconcileIntervalMillis);
}
```
and replace all `throw new ReconciliationException()` with
```java
if (path == null) {
    LOGGER.info("The Knative Broker Resource for ManagedBridge '{}' in namespace '{}' is not ready",
            managedBridge.getMetadata().getName(),
            managedBridge.getMetadata().getNamespace());
    status.markConditionFalse(ManagedBridgeStatus.KNATIVE_BROKER_AVAILABLE);
    return handleFailure(managedBridge);
}
```

It is the same and we avoid to throw an exception. This makes the solution also a bit less error prone because `ReconciliationException` must be catch while now we can add new failure/stop condition without the need to enforce the try/catch

@vaibhavjainwiz Wdyt? I agree that we should not use exceptions as "goto"

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
```diff
@@ -72,100 +65,72 @@ public class ManagedBridgeController implements Reconciler<ManagedBridge>,
...
-        } else if (!status.isConditionTypeTrue(ManagedBridgeStatus.NETWORK_RESOURCE_AVAILABLE)) {
-            status.markConditionTrue(ManagedBridgeStatus.NETWORK_RESOURCE_AVAILABLE);
         }
 
-        LOGGER.info("Ingress networking resource for ManagedBridge with id '{}' in namespace '{}' is ready.", managedBridge.getMetadata().getName(), managedBridge.getMetadata().getNamespace());
-
-        if (!status.isReady()) {
+        if (managedBridgeService.isBridgeStatusChange(managedBridge)) {
+            notifyManager(managedBridge);
             metricsService.onOperationComplete(managedBridge, MetricsOperation.CONTROLLER_RESOURCE_PROVISION);
```
robpblake, line 131: @vaibhavjainwiz This is changing our metric recording logic as you record the provisioning operation as complete if there are any changes in conditions on the `ManagedBridge`. In the previous implementation, we only recorded the metric if the `ManagedBridge` had the `ready=true` condition e.g all provisioning was done.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/ManagedBridgeStatus.java
```diff
@@ -22,19 +23,21 @@ public class ManagedBridgeStatus extends CustomResourceStatus {
...
-            add(new Condition(CONFIG_MAP_AVAILABLE, ConditionStatus.Unknown));
-            add(new Condition(KNATIVE_BROKER_AVAILABLE, ConditionStatus.Unknown));
-            add(new Condition(AUTHORISATION_POLICY_AVAILABLE, ConditionStatus.Unknown));
-            add(new Condition(NETWORK_RESOURCE_AVAILABLE, ConditionStatus.Unknown));
-        }
-    };
+    private static Set<Condition> getInitialConditions() {
+        return new HashSet<>() {
+            {
+                add(new Condition(ConditionTypeConstants.READY, ConditionStatus.Unknown));
```
manstis, line 29: @vaibhavjainwiz I thought you was removing the top-level `Condition`?

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/ManagedBridgeStatus.java
```diff
@@ -22,19 +23,21 @@ public class ManagedBridgeStatus extends CustomResourceStatus {
...
+                add(new Condition(SECRET_AVAILABLE, ConditionStatus.Unknown));
```
manstis, line 30: I'd prefer you to use the following to avoid code duplication:

https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/infra-parent/infra-v2/src/main/java/com/redhat/service/smartevents/infra/v2/api/models/DefaultConditions.java#L15-L19

https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/infra-parent/infra-v2/src/main/java/com/redhat/service/smartevents/infra/v2/api/models/ConditionStatus.java

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
robpblake: +1 to @danielezonca comment.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
robpblake: > @danielezonca @vaibhavjainwiz
> 
> > We can do it smarter with some delta but I would start simple with a full update
> 
> The `manager` needs and depends on the full set at the moment.

@manstis Just to clarify, I think the question is can we be smarter about which `ManagedBridge` we send the `conditions` for, not which `conditions` do we send for a particular `ManagedBridge`. It's 100% the intention at this point that the data plane reports all `conditions` on the `ManagedBridge`.

robpblake: > 

@manstis I'm not sure about this. I think the Controller should be the one who would have the knowledge as to why/where the timeout had occurred and should report it via the `Conditions` it adds to the resource. This keeps all provisioning logic for a particular resource encapsulated within the Controller and reported via the Conditions e.g we could decide that we have different timeouts for certain resources.

It also keeps the role of the synchronizer simple: send and receive updates from the Control Plane

We also have https://issues.redhat.com/browse/MGDOBR-1308 to discuss at some point in which we move the general "timeout" back to the control plane.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
```diff
@@ -57,9 +53,6 @@ public class ManagedBridgeController implements Reconciler<ManagedBridge>,
     @ConfigProperty(name = "event-bridge.managed-bridge.poll.interval-milliseconds")
     int reconcileIntervalMillis;
 
-    @ConfigProperty(name = "event-bridge.managed-bridge.deployment.timeout-seconds")
```
robpblake, line 60: @vaibhavjainwiz Do you intend to re-add this at some point? I'm fine for it to be in another PR, but we just need to make sure we don't forget to do it

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
vaibhavjainwiz: done..
As suggested I had introduced a new `Scheduler` inside `ManagerBridgeSyncService` to send status back.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
vaibhavjainwiz: done.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
vaibhavjainwiz: done

shard-operator-parent/shard-operator-v2/src/test/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeServiceImplTest.java
vaibhavjainwiz: done

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/ManagedBridgeStatus.java
vaibhavjainwiz: done.. leftover by mistake.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/ManagedBridgeStatus.java
vaibhavjainwiz: done

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
vaibhavjainwiz: timeout logic would moved to the control plane. It would not required in data-plane from now. 
I had created below jira for it
https://issues.redhat.com/browse/MGDOBR-1332

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
vaibhavjainwiz: done

shard-operator-parent/shard-operator-v2/src/test/java/com/redhat/service/smartevents/shard/operator/v2/ManagerClientTest.java
vaibhavjainwiz: done

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+                .onItem()
+                .invoke(this::processDelta)
+                .subscribe().with(
+                        success -> LOGGER.debug("Successfully processed ManagedBridge deltas"),
+                        error -> LOGGER.error("Failed to process ManagedBridge deltas", error));
+    }
+
+    @Override
+    public void syncManagedBridgeStatusBackToManager() {
+        managerClient.fetchBridgesToDeployOrDelete()
```
danielezonca, line 53: Why does it fetch the resources from the manager?
I think we can even do a simple "send all statuses" reading all resources available in kubernetes. In this way we should avoid inconsistencies.
For example what if a resource switches to failed? This resource is not in the list of "toDeployOrDelete".

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeService.java
```diff
@@ -21,4 +23,8 @@ public interface ManagedBridgeService {
     KnativeBroker fetchOrCreateKnativeBroker(ManagedBridge managedBridge, ConfigMap configMap);
 
     AuthorizationPolicy fetchOrCreateBridgeAuthorizationPolicy(ManagedBridge managedBridge, String path);
+
+    boolean isBridgeStatusChange(ManagedBridge updatedBridge);
```
danielezonca, line 27: Minor note, this method makes the assumption that it is invoked _before_ the actual resource is updated on k8s side.
We might make it "stateless" moving the fetch outside and transform the method as `compareStatus(old, new)`
Btw it is safe in a reconcile loop so I'm fine also with this code, just a minor comment

infra-parent/infra-v2/src/main/java/com/redhat/service/smartevents/infra/v2/api/models/ConditionStatus.java
```diff
@@ -8,7 +8,8 @@ public enum ConditionStatus implements BaseEnumeration {
     TRUE("True"),
     FALSE("False"),
     UNKNOWN("Unknown"),
-    FAILED("Failed");
+    FAILED("Failed"),
+    DELETED("Deleted");
```
r00ta, line 12: We don't have such logic for a `DELETED` status in the CP. In case the resources are deleted, there should be a condition like `ResourceDeleted` with status `TRUE`.

shard-operator-parent/shard-operator-core/src/main/java/com/redhat/service/smartevents/shard/operator/core/resources/Condition.java
```diff
@@ -53,6 +53,7 @@ public class Condition {
     public Condition(@JsonProperty("type") final String type, @JsonProperty("status") final ConditionStatus status) {
         this.type = type;
         this.status = status;
+        this.lastTransitionTime = new Date();
```
r00ta, line 56: Better to use `ZonedDateTime.now(ZoneOffset.UTC);`

r00ta (commented):
As per chat, I'm personally -1 with the choice of sending back the entire list of resources from the Data plane to the Control Plane every X seconds. The main reason why I'm against this is because this is going to force the manager to run N update queries (where N is the total number of resources deployed on a shard) every X seconds: the user is going to perform very few operations during the lifecycle of the resources, so it's very likely that the 99% of the updates/queries are absolutely redundant/not needed at all. 
The scenarios that you aim to cover with this could be implemented keeping in memory the failed calls and implement a retry mechanism (no need to use a persistent storage because if the application is restarted it is going to process - and send an update - all the resources by design). 

Having said that for the sake of transparency, you have the green light from @robpblake . So feel free to move forward with this.

infra-parent/infra-v2/src/main/java/com/redhat/service/smartevents/infra/v2/api/models/ConditionStatus.java
robpblake: Agreed with @r00ta . IIRC the proposal was to have a `Condition` in the form `Deleted=true` once the resource has been deleted.

manstis: I prefer `XXXDeleted=Unknown|True|False|Failed` too.

It is symmetrical with the _creation_ `Condition`'s of `XXXReady=Unknown|True|False|Failed`.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeServiceImpl.java
```diff
@@ -244,4 +248,31 @@ public AuthorizationPolicy fetchOrCreateBridgeAuthorizationPolicy(ManagedBridge
 
         return existing;
     }
+
+    @Override
+    public boolean isBridgeStatusChange(ManagedBridge updatedBridge) {
```
manstis, line 253: I believe we should probably also include whether there are _new_ `Condition`'s or _removed_ `Condition`'s. This is not strictly necessary when the `Set` of `Condition`'s remains unchanged however given [this](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1438/files#r1051955973) discussion that can no longer be assumed to be true.

robpblake:
> As per chat, I'm personally -1 with the choice of sending back the entire list of resources from the Data plane to the Control Plane every X seconds. The main reason why I'm against this is because this is going to force the manager to run N update queries (where N is the total number of resources deployed on a shard) every X seconds: the user is going to perform very few operations during the lifecycle of the resources, so it's very likely that the 99% of the updates/queries are absolutely redundant/not needed at all. The scenarios that you aim to cover with this could be implemented keeping in memory the failed calls and implement a retry mechanism (no need to use a persistent storage because if the application is restarted it is going to process - and send an update - all the resources by design).
> 
> Having said that for the sake of transparency, you have the green light from @robpblake . So feel free to move forward with this.

Thanks @r00ta 👍  I have captured all the points raised in our discussion into follow on JIRAs so that we actively review our decisions here and continue to iterate in this area. 

As discussed on Friday, I hear and recognise the concerns you are raising. The goal with this work was to get a logical split inside the data plane of "synchronizer" e.g. the thing that reads/writes from the control plane, and then the operator itself e.g the thing that reconciles the k8s resources.

I consider this the very first basic implementation of the synchronizer and there are many improvements we can make, some of which you have called out in the discussion. I'm comfortable with this as the first incredibly basic implementation so that we get the end-to-end working with the new API contract between CP and DP, and then once we have the base line, start to iterate and improve.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceImpl.java
manstis: I agree with @danielezonca.

IMO k8s should be queried for the resources' status conditions.

The `operator` is the "source of truth" of `Condition`'s during provisioning/deprovisioning; not the `manager`.

Furthermore the `manager` holds a [temporary](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/468203f50e91160441a47cda31801438e84d7787/infra-parent/infra-v2/src/main/java/com/redhat/service/smartevents/infra/v2/api/models/DefaultConditions.java#L6) `Condition` as a place-holder until the first update from the `operator`. Then it overwrites the temporary one with the set from the `operator`. We took this approach so that the `manager` could be blissfully unaware of what `Condition`'s are required to provision/deprovision a resource. If, the set of `Condition`'s changes the `manager` works without change.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+        Map<String, ManagedBridge> deployedManagedBridges = managedBridgeService.fetchAllManagedBridges()
+                .stream().collect(Collectors.toMap(m -> m.getSpec().getId(), m -> m));
+
+        List<BridgeStatusDTO> bridgeStatusDTOs = new ArrayList<>(bridgeDTOList.size());
+        for (BridgeDTO bridgeDTO : bridgeDTOList) {
+            ManagedBridge deployedManagedBridge = deployedManagedBridges.get(bridgeDTO.getId());
+            if (deployedManagedBridge != null) {
+                bridgeStatusDTOs.add(BridgeStatusConverter.fromManagedBridgeToBrideStatusDTO(deployedManagedBridge));
+            } else {
+                if (bridgeDTO.getOperationType() == OperationType.DELETE) {
```
manstis, line 82: Given the discussion [above](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1438/files#r1051915340) this code currently serves no purpose and in fact is broken. If you query the `manager` for the list of resources for status updates and then use the `Operation==DELETED` to know whether to send a set of _deleted_ `Condition`'s back you're basically saying "Send me a list of resources; oh, X is deleted I best send you an update saying it has been deleted".

Furthermore, we don't store _deleted_ resources in the database. Once they are deleted they are removed from the tables. The `manager` is therefore not even possible to advise what resources have **actually been deleted**; only those that have been **requested to be deleted**.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagerClientImpl.java
```diff
@@ -53,6 +55,15 @@ public Uni<List<BridgeDTO>> fetchBridgesToDeployOrDelete() {
                 .onItem().transform(this::getBridges);
     }
 
+    @Override
+    public Uni<HttpResponse<Buffer>> notifyBridgeStatus(List<BridgeStatusDTO> bridgeStatusDTOs) {
+        LOGGER.debug("Notifying manager about the new status of the Bridge");
+        return getAuthenticatedRequest(webClientManager.put(V2APIConstants.V2_SHARD_API_BASE_PATH), request -> request.sendJson(bridgeStatusDTOs))
+                .onItem().invoke(success -> updateManagerRequestMetricsOnSuccess(MetricsOperation.OPERATOR_MANAGER_UPDATE, success))
```
manstis, line 62: This changes the semantics of metrics handling.

We used to store a metric count of for each resource updated (because we sync'ed each individually). If we are sending status updates for multiple resources in a batch we need to either accept that metrics now represent the number of HTTP updates and not the number of resource updates; or change how metrics are recorded in this new code.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
manstis: Erm, [L#130](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/468203f50e91160441a47cda31801438e84d7787/shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java#L130) is an `@Override` statement.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
```diff
@@ -219,4 +161,8 @@ private String extractBrokerPath(KnativeBroker broker) {
             return null;
         }
     }
+
+    private UpdateControl<ManagedBridge> handleFailure(ManagedBridge managedBridge) {
+        return UpdateControl.updateStatus(managedBridge).rescheduleAfter(reconcileIntervalMillis);
```
manstis, line 166: If you're stating (and we've all accepted) "time-out" is to be handled in the `manager` you can remove this method and just use the regular `UpdateControl.updateStatus(...)` - without the `reschedule(..)`. We added that specific method invocation when the "time-out" was handled in the `operator` to ensure our reconcile loop ran at determined intervals. Otherwise, if we had our loop invoked by k8s when events triggered it our "time-out" could have expired long ago.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/converters/ConditionConverter.java
```diff
@@ -0,0 +1,27 @@
...
+        return conditions.stream().map(ConditionConverter::fromConditionToConditionDTO).collect(Collectors.toSet());
+    }
+
+    private static ConditionDTO fromConditionToConditionDTO(Condition condition) {
+        return new ConditionDTO(condition.getType(),
+                ConditionStatus.fromString(condition.getStatus().name()),
+                condition.getReason(),
+                condition.getMessage(),
+                condition.getErrorCode(),
+                condition.getLastTransitionTime().toInstant().atZone(ZoneOffset.UTC));
```
manstis, line 25: I assume this is redundant if you remove the of use `Date` as requested by @r00ta elsewhere.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/ManagedBridgeStatus.java
```diff
@@ -16,25 +22,18 @@
...
+        conditions.add(new Condition(DP_CONFIG_MAP_READY_NAME, ConditionStatus.Unknown));
+        conditions.add(new Condition(DP_KNATIVE_BROKER_READY_NAME, ConditionStatus.Unknown));
+        conditions.add(new Condition(DP_AUTHORISATION_POLICY_READY_NAME, ConditionStatus.Unknown));
+        conditions.add(new Condition(DP_NETWORK_RESOURCE_READY_NAME, ConditionStatus.Unknown));
+        return conditions;
+    }
 
     public ManagedBridgeStatus() {
-        super(CONDITIONS);
+        super(getInitialConditions());
```
manstis, line 36: `getCreationConditions()` will probably be best and an equivalent `getDeletionConditions()` for the removal.

shard-operator-parent/shard-operator-v2/src/main/resources/application.properties
```diff
@@ -1,3 +1,2 @@
 # Configuration of polling and timeout operations for resources relating to a ManagedBridge instance
-event-bridge.managed-bridge.poll.interval-milliseconds=5000
-event-bridge.managed-bridge.deployment.timeout-seconds=60
\ No newline at end of file
+event-bridge.managed-bridge.poll.interval-milliseconds=5000
```
manstis, line 2: You can remove this if "time-out" is to be moved to the `manager`.

shard-operator-parent/shard-operator-v2/src/test/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceTest.java
```diff
@@ -0,0 +1,107 @@
...
+    @Test
+    public void TestSyncManagedBridgeStatusBackToManager() {
+        // setup
+        List<BridgeDTO> bridgeDTOList = new ArrayList<>();
+        BridgeDTO bridgeDTO1 = Fixtures.createBridge(OperationType.CREATE);
+        bridgeDTO1.setId("1");
+        bridgeDTO1.setGeneration(1);
+        bridgeDTOList.add(bridgeDTO1);
+
+        BridgeDTO bridgeDTO2 = Fixtures.createBridge(OperationType.DELETE);
```
manstis, line 77: This test demonstrates my comment above.

The mocked `manager` has a resource set to `DELETED` and then the test asserts that the `manager` receives notification that the resource has been deleted.... erm....

infra-parent/infra-v2/src/main/java/com/redhat/service/smartevents/infra/v2/api/models/ConditionStatus.java
vaibhavjainwiz: done.. I create a new Condition with type `BridgeDeleted` and set it to `True`

shard-operator-parent/shard-operator-core/src/main/java/com/redhat/service/smartevents/shard/operator/core/resources/Condition.java
vaibhavjainwiz: done

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeServiceImpl.java
vaibhavjainwiz: sorry, I didn't get you. Could you please explain

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceImpl.java
vaibhavjainwiz: I am fetching the bridges from manager so that we could also send back status for those resources which are deleted. 
If I just send back status of only those resources whose CRD exists then status for deleted resources would never sent.

@danielezonca @manstis I would remove this logic if you guys think it would not required for now.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeService.java
vaibhavjainwiz: done.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceImpl.java
vaibhavjainwiz: yes, I mean `OperationType.DELETE` would be **requested to be deleted** so that operator could confirm you back that it has successfully delete the resource.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagerClientImpl.java
vaibhavjainwiz: How this metric is different from metric recored in `fetchBridgesToDeployOrDelete`. In `fetchBridgesToDeployOrDelete` as well we are recording metric as a single event for a group of BridgeRequest. I am just trying to keep thing consistent.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
vaibhavjainwiz: @robpblake @danielezonca @r00ta 
Are you all agree to remove `reconcileIntervalMillis` from `handleFailure`.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/converters/ConditionConverter.java
vaibhavjainwiz: `lastTransitionTime`  is `Date` inside `shard-operator-core`. If I make it `ZonedDateTime` them it would require changes in v1 codebase. That why I keep this conversion.

shard-operator-parent/shard-operator-core/src/main/java/com/redhat/service/smartevents/shard/operator/core/resources/Condition.java
vaibhavjainwiz: `lastTransitionTime`  is `Date` inside `shard-operator-core`. If I make it `ZonedDateTime` them it would require changes in v1 codebase.

Maximum I could do it add conversion here `this.lastTransitionTime = Date.from(ZonedDateTime.now(ZoneOffset.UTC).toInstant());`

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceImpl.java
danielezonca: > I am fetching the bridges from manager so that we could also send back status for those resources which are deleted.
> If I just send back status of only those resources whose CRD exists then status for deleted resources would never sent.

Good point, but at the same time the other scenario that I mentioned about a failed resource is not covered by current code.
If we agree that the logic is "send all", the manager should be able to do the delta: if a resource in deleting state is not included in the returned list, it means that it is deleted.

We can even consider a mixed approach where we do both but if we take this statement

> The operator is the "source of truth" of Condition's during provisioning/deprovisioning; not the manager.

it is not wrong to assume that a `deleting` that is not returned by the shard implies that is deleted

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/ManagedBridgeStatus.java
vaibhavjainwiz: done

shard-operator-parent/shard-operator-v2/src/main/resources/application.properties
vaibhavjainwiz: sure, As team decide on to remove retry interval. I will update it accordingly.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceImpl.java
manstis: But you're not querying k8s to know if it's been deleted you're querying the `manager` that already knows its been requested to be deleted. The best your code does is convert an `Operation` into a set of `Condition`'s.

shard-operator-parent/shard-operator-v2/src/test/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceTest.java
vaibhavjainwiz: please refer to my comment above.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagerClientImpl.java
manstis: But inconsistent with how `v1` worked. We should at least agree `v1` was wrong and what you've done is correct.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeServiceImpl.java
manstis: OK, sure. 

So if the original set of `Condition`'s for _create_ contains [`Create1=Unknown`, `Create2=True`] which then becomes [`Create1=True` `Create2=True`] your logic detects this has changed. Great. When we then delete the resource the set of `Conditon`'s for _delete_ contains [`Delete1=Unknown`, `Delete2=Unknown`] your logic would fail to detect this change. In fact I suspect your [code](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1438/files/468203f50e91160441a47cda31801438e84d7787#diff-c352b575f3a8abde0ace3a82d0f8e0e87a5398d47aea81233053bf7b6d961775R271) would throw a `NullPointerExcepton` as the `updatedCondition` would be `null` (as the _old name_ `Create1` cannot be found in the new set of [`Delete1`, `Delete2`]).

infra-parent/infra-v2/src/main/java/com/redhat/service/smartevents/infra/v2/api/models/ConditionStatus.java
manstis: Nooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo.

Your whole logic remains a bit off. The `manager` wants to know as each dependency is removed; much the same as it was advised as they were created. The `manager` tracks "delete requested" and "deleting" before the final deletion. "delete request" is when the User has requested it. "deleting" is when activities are in progress for the deletion (e.g. delete DNS entry, delete Isitio configuration, delete knative broker etc - whatever) and "deleted" is the final update needed to signal the `manager` to actually delete database records.

We need conditions symmetrical to _create_ for _delete_. For example:-
```
    public static final String DP_SECRET_DELETED_NAME = "SecretDeleted";
    public static final String DP_CONFIG_MAP_DELETED_NAME = "ConfigMapDeleted";
    public static final String DP_KNATIVE_BROKER_DELETED_NAME = "KNativeBrokerDeleted";
    public static final String DP_AUTHORISATION_POLICY_DELETED_NAME = "AuthorisationPolicyDeleted";
    public static final String DP_NETWORK_RESOURCE_DELETED_NAME = "NetworkResourceDeleted";
```
IDK if they are all needed (as k8s will delete some automagically) but for the resources we need to delete manually I'd have thought it preferential to have an explicit `Condition` for them (so that we'd know what had been deleted successfully if the `operator` pod died before deletion had fully completed).

I guess you don't handle deletion [yet](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/468203f50e91160441a47cda31801438e84d7787/shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeServiceImpl.java#L123).

shard-operator-parent/shard-operator-v2/src/main/resources/application.properties
robpblake: @manstis @vaibhavjainwiz Let's leave it in for now to reduce the number of changes in this PR. We have https://issues.redhat.com/browse/MGDOBR-1308 to discuss whether moving timeout back to the Manager is the right thing to do. Once we've had that discussion, we can make another (smaller PR) to update.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceImpl.java
robpblake: @vaibhavjainwiz I think all we're missing here is a check to make sure that the CR instance is not in the map of deployed managed bridges.

```
if (bridgeDTO.getOperationType() == OperationType.DELETE) {
  
  / * Add this: make sure that the CR has actually been deleted */
  if(deployedManagedBridge == null) {
      bridgeStatusDTOs.add(getDeletedBridgeStatus(bridgeDTO));
  }
}
```
At the minute you simply assume that if a delete has been requested by the control plane, then it's been deleted by the data plane. The simple `== null` check will verify that the `ManagedBridge` CR, has in fact been deleted.

cc @manstis 

cheers,

Rob

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
robpblake: Let's leave it in place for this PR to reduce the number of changes and then discuss https://issues.redhat.com/browse/MGDOBR-1308. We can remove it in a follow up PR pending what we decide.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceImpl.java
robpblake: @danielezonca Can you clarify what you mean when you say:

```
For example what if a resource switches to failed? This resource is not in the list of "toDeployOrDelete".
```
At the minute the Control Plane will provide all Bridges/Processors when the conditions added by the Control Plane are `true` e.g the Control Plane has finished the provisioning that it is responsible for e.g DNS, Kafka Topic. It's then up to the data plane to do it's provisioning which will either succeed or not.

danielezonca: @robpblake 
I might be wrong but as far as I understand the `fetchBridgesToDeployOrDelete` method is not going to return a resource that has been fully provisioned and now it is just running.
If that assumption is correct, I'm not sure if it is possible to manage a scenario where a resource that is currently running well, move to a failed status (for whatever reason, i.e. a broken update) because this resource is not in the list returned by the manager.

The second quote is just an alternative proposal but I agree that it is better (if possible) to have an explicit `deleted=true`

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
manstis: OK @robpblake.

In which case @vaibhavjainwiz you need to reinstate the time-out code in the [`Controller`](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/90de10229b8a09280db7a0a4724a97dda6cf3e0c/shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java), unless it is elsewhere?

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceImpl.java
manstis: @danielezonca 

> I might be wrong but as far as I understand the fetchBridgesToDeployOrDelete method is not going to return a resource that has been fully provisioned and now it is just running.

That understanding is wrong.

This is something @r00ta and I argued until we were blue in the face; but it was eventually conceded that the `manager` will furnish the `operator` with a list of ALL resources that the `manager` has finished preliminary work for. This includes all new resources, all updated resources, all provisioned resources, all failed resources. EVERYTHING.

danielezonca: Ok if so my comment doesn't apply and I only suggest to rename the method `fetchBridgesToDeployOrDelete` because it sounds to me a bit misleading :) (I think even a simple `fetchBridgesForDataPlane` or similar can be enough)
👍

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeServiceImpl.java
vaibhavjainwiz: - there would not be any `Delete1=Unknown, Delete2=Unknown`. There is only one delete `BridgeDeleted=True`.
- `BridgeDeleted` would not be maintain inside CRD status. Deleted CRDs would be identify by the `sync with manager` part of operator and directly send back to manager through DTO if some CRDs don't exists.
- `public boolean isBridgeStatusChange(ManagedBridge updatedBridge)` would always works with fixed set of conditions.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceImpl.java
vaibhavjainwiz: @robpblake @manstis 
I think there is a understanding gap, I am querying the k8s to fetch existing CRDs and then check for deleted CRD by comparing with bridge object received from Manager.

here
```
 Map<String, ManagedBridge> deployedManagedBridges = managedBridgeService.fetchAllManagedBridges()
                .stream().collect(Collectors.toMap(m -> m.getSpec().getId(), m -> m));
```

`fetchAllManagedBridges` fetch all deployed bridge CRDs from k8s.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagerClientImpl.java
vaibhavjainwiz: there is nothing wrong or right here. In `v1` we are independently sending status back to manager for each bridge whereas in `v2` we are sending them in batch.

I could do following :
 - remove `metric` logic completely from here and create a new Jira to discuss which `metric` are required here.
 - I could move record `mertic` functionality before making REST call. In that case I could iterate over `List<BridgeStatusDTO>` and generate metric foe each for them.
 - Or just keep it as it is.
 
 Please suggest what I should do here.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
vaibhavjainwiz: @manstis Jira for same already exists.
https://issues.redhat.com/browse/MGDOBR-1308

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/converters/BridgeStatusConverter.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.Set;
+
+import com.redhat.service.smartevents.infra.v2.api.models.dto.BridgeStatusDTO;
+import com.redhat.service.smartevents.infra.v2.api.models.dto.ConditionDTO;
+import com.redhat.service.smartevents.shard.operator.v2.resources.ManagedBridge;
+
+public class BridgeStatusConverter {
+
+    public static BridgeStatusDTO fromManagedBridgeToBrideStatusDTO(ManagedBridge managedBridge) {
```
Srihari1192, line 11:
```suggestion
    public static BridgeStatusDTO fromManagedBridgeToBridgeStatusDTO(ManagedBridge managedBridge) {
```

shard-operator-parent/shard-operator-core/src/main/java/com/redhat/service/smartevents/shard/operator/core/resources/Condition.java
r00ta: Until the conditions were not propagated to the manager, it was okaish to have a different representation of the a date in the control plane and the data plane (it was not ideal, but acceptable). Now that the Manager expects a ZonedDateTime for the Conditions, we have to align with that. Can you please check if it's possible to change the lastTransitionTime property to a `ZonedDateTime`?

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/converters/BridgeStatusConverter.java
vaibhavjainwiz: done

shard-operator-parent/shard-operator-v2/src/main/resources/application.properties
vaibhavjainwiz: @manstis Jira for same already exists.
https://issues.redhat.com/browse/MGDOBR-1308

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceImpl.java
r00ta: With the current implementation you are not sending any update for a resource that is on the shard but (for any reason) not in the fetched resources from the control plane

shard-operator-parent/shard-operator-core/src/main/java/com/redhat/service/smartevents/shard/operator/core/resources/Condition.java
vaibhavjainwiz: As I say `lastTransitionTime` is `Date` inside `shard-operator-core`. If I make it `ZonedDateTime` them it would require changes in v1 codebase.
Maybe we could comeback it once we removed v1 from our codebase. If its ok with you I could create Jira for it so that this work would not lost.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceImpl.java
vaibhavjainwiz: How is it possible for any resource to be on operator but not on control-plane. Even in 1/1000 case it is a case then it make no sense to send a status of resource to control-plane which doesn't exists in there.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagerClientImpl.java
robpblake: @vaibhavjainwiz I think we should just leave it as it is 👍 

This metric has always simply recorded the number of HTTP requests between data plane and control plane, recording the http status code in the event of failure. For `v1` it would record multiple metrics for each Bridge/Processor because we had the various lifecycle states e.g `PROVISIONING`, `READY`...

In `v2` it will simply record a metric each time we do the sync back. It's not being used to record the number of correctly provisioned resources.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/ManagedBridgeSyncServiceImpl.java
robpblake: > @robpblake @manstis I think there is a understanding gap, I am querying the k8s to fetch existing CRDs and then check for deleted CRD by comparing with bridge object received from Manager.
> 
> here
> 
> ```
>  Map<String, ManagedBridge> deployedManagedBridges = managedBridgeService.fetchAllManagedBridges()
>                 .stream().collect(Collectors.toMap(m -> m.getSpec().getId(), m -> m));
> ```
> 
> `fetchAllManagedBridges` fetch all deployed bridge CRDs from k8s.

@vaibhavjainwiz Ah yes, you're absolutely right. Thank you for pointing that out 👍

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
manstis: My point is; if you're not having time-out in the `operator` (and this PR removes it) then you don't need the use of `rescheduleAfter(..)` and this method along with the property can be removed. Time-out, the property and use of `rescheduleAfter(..)` all go hand-in-hand.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
```diff
@@ -121,51 +96,34 @@ public UpdateControl<ManagedBridge> reconcile(ManagedBridge managedBridge, Conte
...
+        } else {
+            LOGGER.info("Ingress networking resource for ManagedBridge with id '{}' in namespace '{}' is ready.", managedBridge.getMetadata().getName(),
+                    managedBridge.getMetadata().getNamespace());
+            status.markConditionTrue(DP_NETWORK_RESOURCE_READY_NAME);
         }
 
-        LOGGER.info("Ingress networking resource for ManagedBridge with id '{}' in namespace '{}' is ready.", managedBridge.getMetadata().getName(), managedBridge.getMetadata().getNamespace());
-
-        if (!status.isReady()) {
+        if (isBridgeStatusChange(managedBridge)) {
```
manstis, line 123: You are recording the operation was complete if there was no change in the `Condition`'s between subsequent iterations.

What happens if it takes more than 1 iteration of the reconsile loop before `Condition`'s change?

It **might not be a problem** if we only relied on k8s invoking our loop in response to changes to Custom Resources on which we depend with the k8s eventing mechanism however we force re-entry with use of `rescheduleAfter(..)` and hence the possibility of our reconcile loop running multiple times before `Condition`'s change increases.

It might be better to check if ALL conditions are `True` to signal completion vs there has not been a change.

robpblake: @manstis It only ever gets to this point in the code if all dependencies are ready, otherwise the reconcile loop is ended early via a call to the `handleFailure(ManagedBridge managedBridge)` method. So at this point, essentially all conditions on the `ManagedBridge` will have been be set to `true` earlier in the reconcile loop, or they are already `true`.  If the conditions were set to `true` in the reconcile loop, then the `isBridgeStatusChange` method detects that and implicitly marks the operation as complete.

@vaibhavjainwiz I think something like a `isReady()` method on the `ManagedBridgeStatus` might make it clearer so that the logic looks something like this:

```
if (isBridgeStatusChange(managedBridge)) {
  if(status.isReady()) {
    // update metrics
  }
}
```
However, I'm not going to block this PR for that change, so if you agree, please create a JIRA and we can follow up.

robpblake (approved):
@vaibhavjainwiz Thank you!

Please can you resolve all open discussions and where follow up JIRAs are required, create them. 

cheers,

Rob

manstis (approved):
Looks as good as it's going to get on this sprint.

I acknowledge more work is needed to complete the feature; but this PR is a reasonable start.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
vaibhavjainwiz: Create below Jira to incorporate this change
https://issues.redhat.com/browse/MGDOBR-1343

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java
vaibhavjainwiz: got your point @manstis. Lets incorporate this changes in https://issues.redhat.com/browse/MGDOBR-1308

Srihari1192 (approved):

PR #1445, opened 2022-12-20
MGDOBR-1340: ProcessorService: Restore metrics capture

manstis:
[https://issues.redhat.com/browse/MGDOBR-1340](https://issues.redhat.com/browse/MGDOBR-1340)

## PR owner checklist

- [x] Your commit messages are clear and reference the JIRA issue e.g: "[MGDOBR-XXX] - $clear_explanation_of_what_you_did"

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>


manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/services/ProcessorServiceImpl.java
```diff
@@ -288,15 +295,20 @@ public Processor updateProcessorStatus(ProcessorStatusDTO statusDTO) {
...
                 break;
 
             case UPDATE:
-                // TODO: record metrics with MetricsService
-                // metricsService.onOperationComplete(processor, MetricsOperation.MANAGER_RESOURCE_MODIFY);
+                if (isOperationComplete(updatedConditions)) {
+                    metricsService.onOperationComplete(processor, MetricsOperation.MANAGER_RESOURCE_MODIFY);
+                } else if (isOperationFailed(updatedConditions)) {
+                    metricsService.onOperationFailed(processor, MetricsOperation.MANAGER_RESOURCE_MODIFY);
+                }
```
r00ta, line 311: According to the current update strategy we are going to receive the updates every X seconds, so you would register the success/failure operations multiple times - the metrics would not have much sense anymore

r00ta (commented):
I think we have to revisit the strategy to register the metrics entirely :/

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/services/ProcessorServiceImpl.java
manstis: I hear you!! 

`CREATE` and `DELETE` will only be recorded once due to other checks; but `UPDATE` and `FAILED` are problematic.

I wonder what can be done to address `UPDATE` and `FAILED` (I don't really want to compare `Condition`'s).

Hmmm.....

manstis: @r00ta I've added some more commits.

Most notably is the last that prevents duplicate updates from being recorded.

I convert `Condition`'s to `ManagedResourceStatus` when checking for duplicate updates to prevent failures in different dependencies from being recorded as successive failed operations. I think it is an improvement on what happens in `v1` (where we potentially record duplicate metrics for failures).

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/metrics/MetricsServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+    String operationTotalCountMetricName;
+
+    @ConfigProperty(name = "rhose.metrics-name.operation-success-total-count")
+    String operationTotalSuccessCountMetricName;
+
+    @ConfigProperty(name = "rhose.metrics-name.operation-failure-total-count")
+    String operationTotalFailureCountMetricName;
+
+    @ConfigProperty(name = "rhose.metrics-name.operation-duration-seconds")
+    String operationDurationMetricName;
```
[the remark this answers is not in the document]
manstis: I asked the question about metrics and observability in yesterdays planning meeting.

@robpblake commented that, for now, we can consider them "application wide" and hence re-using the same names is not a problem...... The metric itself contains a tag for the FQCN so we can disambiguate between versions using that.. but it begs the question as to whether (1) we should have different metric names, (2) we should tag the version if using the same metric name.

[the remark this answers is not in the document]
manstis: I added a commit adding a version tag.

manstis:
@r00ta I know you're not a fan.... but this PR adds back (an _improved_ implementation of) metrics to `v2`.

I wonder if you'd be able to provide one final review an opinion as to whether you feel we should or should not proceed.

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/services/BridgeServiceImpl.java
```diff
@@ -232,53 +238,62 @@ public Bridge updateBridgeStatus(ResourceStatusDTO statusDTO) {
             return bridge;
         }
         Operation operation = bridge.getOperation();
+        if (Objects.isNull(operation)) {
```
robpblake, line 241: @manstis According to the DB schema, this can never happen as both columns of the embedded `Operation` object have a `not null` clause on them.

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/metrics/MetricsServiceImpl.java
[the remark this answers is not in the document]
robpblake: The metric names in use here were designed to provide the SLI for the original set of SLO that we had for `v1`. IIRC, there is nothing actively relying on those SLI right now, but it's worth checking with @msmagnanijr to see if any of our dashboards/alerting in app-interface would be impacted. Even so, I don't immediately see a problem with re-using the same metric name, but your `v2` tag seems fine as well to me @manstis 

Another thing for us to consider is that we didn't have any SLI or SLO relating to `Processor` lifecycle events for `v1` as there was no SLO offered from RHOC on connector deployment. Given that we now no longer use RHOC for `Processors` we can probably consider adding an SLO for our `Processors`. I'll create a JIRA for that as we'll need to have some agreement with PM/BU on what they want to offer there (if anything).

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/services/BridgeServiceImpl.java
```diff
@@ -232,53 +238,62 @@ public Bridge updateBridgeStatus(ResourceStatusDTO statusDTO) {
...
+                } else if (ConditionUtilities.isOperationFailed(updatedConditions)) {
+                    metricsService.onOperationFailed(bridge, MetricsOperation.MANAGER_RESOURCE_PROVISION);
                 }
                 break;
 
             case UPDATE:
-                // TODO: record metrics with MetricsService
-                // metricsService.onOperationComplete(bridge, MetricsOperation.MANAGER_RESOURCE_MODIFY);
+                if (ConditionUtilities.isOperationComplete(updatedConditions)) {
+                    metricsService.onOperationComplete(bridge, MetricsOperation.MANAGER_RESOURCE_MODIFY);
```
robpblake, line 275: @manstis For consistency, should we rename this to `MANAGER_RESOURCE_UPDATE`. I'm aware that this will impact `v1` as well, but if there is nothing relying on it, I don't think it's a problem

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/services/ProcessorServiceImpl.java
```diff
@@ -277,54 +283,64 @@ public Processor updateProcessorStatus(ResourceStatusDTO statusDTO) {
                     statusDTO.getGeneration());
             return processor;
         }
+
         Operation operation = processor.getOperation();
+        if (Objects.isNull(operation)) {
```
robpblake, line 288: @manstis Same comment here. I think this scenario is prevented by our DB schema.

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/metrics/MetricsServiceImpl.java
```diff
@@ -0,0 +1,107 @@
...
+        return List.of(instanceTag, versionTag, operation.getMetricTag());
+    }
+
+    private void recordOperationDuration(ManagedResourceV2 managedResource, MetricsOperation operation) {
+        Duration operationDuration = calculateOperationDuration(managedResource.getOperation());
+        meterRegistry.timer(operationDurationMetricName, buildTags(managedResource, operation)).record(operationDuration);
+    }
+
+    private Duration calculateOperationDuration(Operation operation) {
+        return Duration.between(operation.getRequestedAt(), ZonedDateTime.now(ZoneOffset.UTC));
```
robpblake, line 104: @manstis Shouldn't we be persisting the time the operation was completed on the `Operation` resource? We set it when the user first requests a create/update/delete, so shouldn't we additionally record the timestamp when we first determine the operation to be either failed or completed?

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/metrics/MetricsServiceImpl.java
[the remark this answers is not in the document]
robpblake: FYI added https://issues.redhat.com/browse/MGDOBR-1363 for us to discuss SLO around Processors with BU/PM.

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/services/ProcessorServiceImpl.java
manstis: Removed check.

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/services/BridgeServiceImpl.java
manstis: Removed check.

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/services/BridgeServiceImpl.java
manstis: Changed in a new commit. I also changed the _tag value_ to `update` too (as opposed to just changing the name of the enumeration entry). This means we have two _operations_ with the same _tag value_ but since they're used on different metrics I don't think it will cause a problem.

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/services/BridgeServiceImpl.java
```diff
@@ -232,53 +238,62 @@ public Bridge updateBridgeStatus(ResourceStatusDTO statusDTO) {
...
+                        metricsService.onOperationComplete(bridge, MetricsOperation.MANAGER_RESOURCE_PROVISION);
                     }
+                } else if (ConditionUtilities.isOperationFailed(updatedConditions)) {
+                    metricsService.onOperationFailed(bridge, MetricsOperation.MANAGER_RESOURCE_PROVISION);
                 }
                 break;
 
             case UPDATE:
-                // TODO: record metrics with MetricsService
-                // metricsService.onOperationComplete(bridge, MetricsOperation.MANAGER_RESOURCE_MODIFY);
```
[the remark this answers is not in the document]
manstis: Yes it was intentional. 

A `v1` resource had properties to reflect the timestamps when different _operations_ were performed on it; created (`submittedAt`), updated (`modifiedAt`) and deleted (`deletionRequestedAt`). `v2` introduces the `Operation` class that details _what_ is being done to a resource (created, updated or deleted). The `Operation` contains a single `requestedAt` property; which when examined in the context of the operation itself equates to the discrete properties in `v1`.

In `v1` the `modifiedAt` and `deletionRequestedAt` are only used to calculate the duration an operation took to complete to record the same as a metric. The properties are not used for any other purpose.

In `v2` we can calculate the duration from the `Operation`.

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/metrics/MetricsServiceImpl.java
manstis: The "completion" timestamp is never used other than to [calculate the duration](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/manager-parent/manager-v1/src/main/java/com/redhat/service/smartevents/manager/v1/metrics/MetricsServiceImpl.java#L91-L102) metric.

This is why we decided not to store it in `v2`.

ChughShilpa (approved):

Aakanksha0311 (approved):
LGTM

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/services/BridgeServiceImpl.java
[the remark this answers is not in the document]
robpblake: @manstis But in `v1` we did expose the `modifiedAt` timestamp to the user via the API, which I think feels reasonable as it allows a user to track when they requested the update of a resource.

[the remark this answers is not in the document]
robpblake: @manstis Ah, perfect. Thanks for the explanation 👍

robpblake (approved):

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/services/BridgeServiceImpl.java
```diff
@@ -231,54 +237,59 @@ public Bridge updateBridgeStatus(ResourceStatusDTO statusDTO) {
...
         statusDTO.getConditions().forEach(c -> updatedConditions.add(Condition.from(c, ComponentType.SHARD)));
+
+        // Don't do anything if Conditions are unchanged from our previous state.
+        ManagedResourceStatusV2 status = StatusUtilities.getManagedResourceStatus(operation, conditions);
+        ManagedResourceStatusV2 originalStatus = StatusUtilities.getManagedResourceStatus(operation, updatedConditions);
+        if (Objects.equals(status, originalStatus)) {
+            LOGGER.info("Update for Bridge with id '{}' was discarded. The ManagedResourceStatus reflected by the Conditions is unchanged.", bridge.getId());
+            return bridge;
+        }
+
```
r00ta, line 254: I see the point, but I'm wondering if this is conceptually wrong. Example: 
1) we deploy the bridge `mybridge` with the version `A` of the operator
2) it gets deployed properly by the operator, the manager gets updated, all good.
3) After some months we update the operator  with the version `B`, and we add some new conditions/whatever.
4) The manager gets the update with the new set of conditions, but the manager discards them because the resource is already `ready`

As well as we might discard updates on the `lastTransitionTime` if the resource gets updated but it does not change the status.

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/services/ProcessorServiceImpl.java
```diff
@@ -277,54 +283,59 @@ public Processor updateProcessorStatus(ResourceStatusDTO statusDTO) {
...
         statusDTO.getConditions().forEach(c -> updatedConditions.add(Condition.from(c, ComponentType.SHARD)));
+
+        // Don't do anything if Conditions are unchanged from our previous state.
+        ManagedResourceStatusV2 status = StatusUtilities.getManagedResourceStatus(operation, conditions);
+        ManagedResourceStatusV2 originalStatus = StatusUtilities.getManagedResourceStatus(operation, updatedConditions);
+        if (Objects.equals(status, originalStatus)) {
+            LOGGER.info("Update for Processor with id '{}' was discarded. The ManagedResourceStatus reflected by the Conditions is unchanged.", processor.getId());
+            return processor;
+        }
+
```
r00ta, line 300: same here

r00ta (approved):
Super :rocket:

manager-parent/manager-v2/src/main/java/com/redhat/service/smartevents/manager/v2/services/BridgeServiceImpl.java
r00ta: As per chat, we agreed to add a column in the DB to persist when we completed the user's request as per diagram 
metrics drawio

PR #1492, opened 2023-01-12
MGDOBR-1391: Support Failed conditions in Data Plane

manstis:
[https://issues.redhat.com/browse/MGDOBR-1391](https://issues.redhat.com/browse/MGDOBR-1391)

## PR owner checklist

- [x] Your commit messages are clear and reference the JIRA issue e.g: "[MGDOBR-XXX] - $clear_explanation_of_what_you_did"

## PR reviewer(s) checklist

<details>
<summary>
</summary>

  Comment with: `/rebase`.
- <b>Deploy BOT</b>

</details>

### DIFF
--- a/shard-operator-parent/shard-operator-core/src/main/java/com/redhat/service/smartevents/shard/operator/core/resources/CustomResourceStatus.java
+++ b/shard-operator-parent/shard-operator-core/src/main/java/com/redhat/service/smartevents/shard/operator/core/resources/CustomResourceStatus.java
@@ -9,7 +9,6 @@ import java.util.Set;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.redhat.service.smartevents.infra.core.exceptions.BridgeError;
 
 import io.javaoperatorsdk.operator.api.ObservedGenerationAwareStatus;
 
@@ -36,22 +35,6 @@ public abstract class CustomResourceStatus extends ObservedGenerationAwareStatus
     public final Optional<Condition> getConditionByType(final String conditionType) {
         // o(1) operation since we are fetching by our hash key
         return conditions.stream().filter(c -> conditionType.equals(c.getType())).findFirst();
-    }
-
-    @JsonIgnore
-    public final boolean isReady() { // Don't use this in V2 as READY constant doesn't exist https://issues.redhat.com/browse/MGDOBR-1369
-        return conditions.stream().anyMatch(c -> ConditionTypeConstants.READY.equals(c.getType()) && ConditionStatus.True.equals(c.getStatus()));
-    }
-
-    @JsonIgnore
-    public final boolean isReadyV2() {
-        return conditions.stream().allMatch(c -> ConditionStatus.True.equals(c.getStatus()));
-    }
-
-    @JsonIgnore
-    public final boolean isAugmentingTrueOrUnknown() {
-        return conditions.stream()
-                .anyMatch(c -> ConditionTypeConstants.AUGMENTING.equals(c.getType()) && (ConditionStatus.True.equals(c.getStatus()) || ConditionStatus.Unknown.equals(c.getStatus())));
     }
 
     @JsonIgnore
@@ -133,11 +116,4 @@ public abstract class CustomResourceStatus extends ObservedGenerationAwareStatus
         markConditionFailed(conditionType, null, "", null);
     }
 
-    @JsonIgnore
-    public final void setStatusFromBridgeError(BridgeError bridgeError) {
-        markConditionFalse(ConditionTypeConstants.READY,
-                bridgeError.getReason(),
-                bridgeError.getReason(),
-                bridgeError.getCode());
-    }
 }
--- a/shard-operator-parent/shard-operator-core/src/main/java/com/redhat/service/smartevents/shard/operator/core/resources/CustomResourceStatus.java
+++ b/shard-operator-parent/shard-operator-core/src/main/java/com/redhat/service/smartevents/shard/operator/core/resources/CustomResourceStatus.java
@@ -107,6 +107,33 @@ public abstract class CustomResourceStatus extends ObservedGenerationAwareStatus
     }
 
     @JsonIgnore
+    public final boolean isConditionTypeFailed(final String conditionType, final String reason) {
+        return getConditions().stream().anyMatch(c -> conditionType.equals(c.getType())
+                && Objects.equals(c.getReason(), reason)
+                && ConditionStatus.Failed.equals(c.getStatus()));
+    }
+
+    public void markConditionFailed(final String conditionType, final String reason, final String message, final String errorCode) {
+        final Optional<Condition> condition = this.getConditionByType(conditionType);
+        if (condition.isPresent()) {
+            condition.get().setMessage(message);
+            condition.get().setErrorCode(errorCode);
+            condition.get().setReason(reason);
+            condition.get().setStatus(ConditionStatus.Failed);
+            condition.get().setLastTransitionTime(new Date());
+            this.conditions.add(condition.get());
+        }
+    }
+
+    public void markConditionFailed(final String conditionType, final String reason, final String message) {
+        markConditionFailed(conditionType, reason, message, null);
+    }
+
+    public void markConditionFailed(final String conditionType) {
+        markConditionFailed(conditionType, null, "", null);
+    }
+
+    @JsonIgnore
     public final void setStatusFromBridgeError(BridgeError bridgeError) {
         markConditionFalse(ConditionTypeConstants.READY,
                 bridgeError.getReason(),
--- a/shard-operator-parent/shard-operator-v1/src/test/java/com/redhat/service/smartevents/shard/operator/v1/resources/CustomResourceStatusTest.java
+++ b/shard-operator-parent/shard-operator-v1/src/test/java/com/redhat/service/smartevents/shard/operator/v1/resources/CustomResourceStatusTest.java
@@ -1,6 +1,8 @@
-package com.redhat.service.smartevents.shard.operator.core.resources;
+package com.redhat.service.smartevents.shard.operator.v1.resources;
 
 import org.junit.jupiter.api.Test;
+
+import com.redhat.service.smartevents.shard.operator.core.resources.ConditionReasonConstants;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -9,7 +11,7 @@ public class CustomResourceStatusTest {
     @Test
     public void testMarkConditionReady() {
         // Given
-        final CustomResourceStatus resourceStatus = new FooResourceStatus();
+        final BaseResourceStatus resourceStatus = new FooResourceStatus();
 
         // When
         resourceStatus.markConditionTrue(ConditionTypeConstants.READY);
@@ -22,7 +24,7 @@ public class CustomResourceStatusTest {
     @Test
     public void testMarkConditionFalse() {
         // Given
-        final CustomResourceStatus resourceStatus = new FooResourceStatus();
+        final BaseResourceStatus resourceStatus = new FooResourceStatus();
 
         // When
         resourceStatus.markConditionFalse(ConditionTypeConstants.READY, ConditionReasonConstants.DEPLOYMENT_FAILED, "");
@@ -41,7 +43,7 @@ public class CustomResourceStatusTest {
     @Test
     public void testMarkConditionFailure() {
         // Given
-        final CustomResourceStatus resourceStatus = new FooResourceStatus();
+        final BaseResourceStatus resourceStatus = new FooResourceStatus();
 
         // When
         resourceStatus.markConditionFailed(ConditionTypeConstants.READY, ConditionReasonConstants.DEPLOYMENT_FAILED, "");
--- a/shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/BaseResourceStatus.java
+++ b/shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/BaseResourceStatus.java
@@ -1,0 +1,31 @@
+package com.redhat.service.smartevents.shard.operator.v2.resources;
+
+import java.util.Set;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.redhat.service.smartevents.infra.core.exceptions.BridgeError;
+import com.redhat.service.smartevents.infra.v2.api.models.DefaultConditions;
+import com.redhat.service.smartevents.shard.operator.core.resources.Condition;
+import com.redhat.service.smartevents.shard.operator.core.resources.ConditionStatus;
+import com.redhat.service.smartevents.shard.operator.core.resources.CustomResourceStatus;
+
+abstract class BaseResourceStatus extends CustomResourceStatus {
+
+    public BaseResourceStatus(Set<Condition> initialConditions) {
+        super(initialConditions);
+    }
+
+    @JsonIgnore
+    public final boolean isReady() {
+        return getConditions().stream().allMatch(c -> ConditionStatus.True.equals(c.getStatus()));
+    }
+
+    @JsonIgnore
+    public final void setStatusFromBridgeError(BridgeError bridgeError) {
+        markConditionFailed(DefaultConditions.CP_DATA_PLANE_READY_NAME,
+                bridgeError.getReason(),
+                bridgeError.getReason(),
+                bridgeError.getCode());
+    }
+
+}
--- a/shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/BaseResourceStatus.java
+++ b/shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/BaseResourceStatus.java
@@ -3,8 +3,6 @@ package com.redhat.service.smartevents.shard.operator.v2.resources;
 import java.util.Set;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.redhat.service.smartevents.infra.core.exceptions.BridgeError;
-import com.redhat.service.smartevents.infra.v2.api.models.DefaultConditions;
 import com.redhat.service.smartevents.shard.operator.core.resources.Condition;
 import com.redhat.service.smartevents.shard.operator.core.resources.ConditionStatus;
 import com.redhat.service.smartevents.shard.operator.core.resources.CustomResourceStatus;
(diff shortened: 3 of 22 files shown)


shard-operator-parent/shard-operator-core/src/main/java/com/redhat/service/smartevents/shard/operator/core/resources/CustomResourceStatus.java
```diff
@@ -40,35 +40,35 @@ public final Optional<Condition> getConditionByType(final String conditionType)
 
     @JsonIgnore
     public final boolean isReady() { // Don't use this in V2 as READY constant doesn't exist https://issues.redhat.com/browse/MGDOBR-1369
-        return conditions.stream().anyMatch(c -> ConditionTypeConstants.READY.equals(c.getType()) && ConditionStatus.True.equals(c.getStatus()));
+        return conditions.stream().anyMatch(c -> ConditionTypeConstants.READY.equals(c.getType()) && getTrue().equals(c.getStatus()));
```
manstis, line 43: This is an abstract class in `-core` so I cannot use any _FIXED_ values.

Access is delegated to the different implementations so the correct version-specific literal can be used.

lucamolteni (approved):

vaibhavjainwiz:
@manstis 
Could you please add more detail in Jira for why we need these changes?
https://issues.redhat.com/browse/MGDOBR-1391

vaibhavjainwiz:
@robpblake @danielezonca 
Data-plane has provision the multiple resources like Secret, ConfigMap, KnativeBroker, AuthorisationPolicy etc.
lets assumes If AuthorisationPolicy is failed to provision and timeout occurs then what happended to other resources?
Does they keep exists in DP and continue to consume resources even complete Bridge is Failed to provision because time-out in AuthorisationPolicy ?

Doesn't it make sence to delete all Bridges resource if its not capable to provision?

robpblake:
@manstis I'm not following why we need a `FAILED` condition status? What are we using it for over our usage of `FALSE`?

manstis:
Our Conditions are named like `XYZReady`.

- `Unknown` means we don't know if it is ready or not.
- `True` means it is ready.
- `False` means it is not ready, but we will keep trying to provision it, as it could become ready.
- `Failed` means provisioning has barfed. We consider provisioning to have failed. This could be a time-out; or k8s retry policy exceeded or a dependency's `Condition` meaning it has failed too (e.g. a k8s `Deployment` has a `Condition` called `ReplicaFailure` which (in `v1` and `v2`) leads us to mark the `[Managed]Bridge` as `FAILED`.

This was all discussed and communicated in my document (shared 23rd November) relating to [MGDOBR-1265](https://issues.redhat.com/browse/MGDOBR-1265) and email "Control Plane decisions" of 24th November.

robpblake:
> Our Conditions are named like `XYZReady`.
> 
> * `Unknown` means we don't know if it is ready or not.
> * `True` means it is ready.
> * `False` means it is not ready, but we will keep trying to provision it, as it could become ready.
> * `Failed` means provisioning has barfed. We consider provisioning to have failed. This could be a time-out; or k8s retry policy exceeded or a dependency's `Condition` meaning it has failed too (e.g. a k8s `Deployment` has a `Condition` called `ReplicaFailure` which (in `v1` and `v2`) leads us to mark the `[Managed]Bridge` as `FAILED`.
> 
> This was all discussed and communicated in my document (shared 23rd November) relating to [MGDOBR-1265](https://issues.redhat.com/browse/MGDOBR-1265) and email "Control Plane decisions" of 24th November.

In our discussions yesterday, we talked about using the `Progressing` condition to communicate a timeout, so `Progressing=True` if things are OK, and then `Progressing=False` if things have timed out. How does that fit in here?

manstis:
Hi @robpblake 

> In our discussions yesterday, we talked about using the `Progressing` condition to communicate a timeout, so `Progressing=True` if things are OK, and then `Progressing=False` if things have timed out. How does that fit in here?

We talked about not using `CamelIntegrationAvailable` to signal the timeout.

We've also talked previously about not wanting a top-level `Condition` too.

I'd prefer to use `Managed[Bridge|Processor]Ready` and set it to `Unknown`, `False`, `True` or `Failed` (which could be interpreted as a top level `Condition` but it is semantically different i.e. it has no different interpretation or priority to any other `Condition`). `Progressing` would be a "special" `Condition` and it's presence and value overriding the interpretation of others.

@lucamolteni @vaibhavjainwiz @r00ta What are you opinions?

r00ta:
> Hi @robpblake
> 
> > In our discussions yesterday, we talked about using the `Progressing` condition to communicate a timeout, so `Progressing=True` if things are OK, and then `Progressing=False` if things have timed out. How does that fit in here?
> 
> We talked about not using `CamelIntegrationAvailable` to signal the timeout.
> 
> We've also talked previously about not wanting a top-level `Condition` too.
> 
> I'd prefer to use `Managed[Bridge|Processor]Ready` and set it to `Unknown`, `False`, `True` or `Failed` (which could be interpreted as a top level `Condition` but it is semantically different i.e. it has no different interpretation or priority to any other `Condition`). `Progressing` would be a "special" `Condition` and it's presence and value overriding the interpretation of others.
> 
> @lucamolteni @vaibhavjainwiz @r00ta What are you opinions?

My opinion is that we have to decide to use a top-level condition or not. If we decide to step back on our original plan about not using a top-level condition, it's fine by me. I simply think it's important that we don't end up using something like a top-level condition but we don't call it "top-level" - it's just confusing and the contract between CP/DP would be even more confusing. 
For the timeout/failure thing, I don't see the problem in populating the `message`/`error` field in the specific condition (like the Camel's condition) with something like `Failed due to timeout` to notify **why** the condition was set to `failed` - no need for a kind of top-level condition to handle this case - IMO

robpblake:
> Hi @robpblake
> 
> > In our discussions yesterday, we talked about using the `Progressing` condition to communicate a timeout, so `Progressing=True` if things are OK, and then `Progressing=False` if things have timed out. How does that fit in here?
> 
> We talked about not using `CamelIntegrationAvailable` to signal the timeout.
> 
> We've also talked previously about not wanting a top-level `Condition` too.
> 
> I'd prefer to use `Managed[Bridge|Processor]Ready` and set it to `Unknown`, `False`, `True` or `Failed` (which could be interpreted as a top level `Condition` but it is semantically different i.e. it has no different interpretation or priority to any other `Condition`). `Progressing` would be a "special" `Condition` and it's presence and value overriding the interpretation of others.
> 
> @lucamolteni @vaibhavjainwiz @r00ta What are you opinions?

OK, thanks for the refresher.

I'd personally prefer at this time to keep the status specific to the `Condition` rather than introducing a top-level one, so I'm happy to go with `Failed` as well.

robpblake:
@manstis I'm -1 on the versioned approach to `Conditions` if I'm honest. I don't think the additional overhead of the class hierarchy is worth the fact that `v1` controllers shouldn't use the `Failed` Condition status. I think we can just add another value to the `ConditionStatus` enum and expose methods to set `Condition` values to `Failed` on the `CustomResourceStatus`. It will only be used by the `v2` controllers, but I really don't see that being a problem as modification of the `Conditions` on a resource are owned by the Controller.

Any reason not to take this simpler approach that you can see?

cheers,

Rob

robpblake (commented):
@manstis I'm -1 on the versioned approach to Conditions if I'm honest. I don't think the additional overhead of the class hierarchy is worth the fact that v1 controllers shouldn't use the Failed Condition status. I think we can just add another value to the ConditionStatus enum and expose methods to set Condition values to Failed on the CustomResourceStatus. It will only be used by the v2 controllers, but I really don't see that being a problem as modification of the Conditions on a resource are owned by the Controller.

Any reason not to take this simpler approach that you can see?

cheers,

Rob

manstis:
HI @robpblake 

> Any reason not to take this simpler approach that you can see?

Flexibility really (not necessarily for `v1` and `v2` but thinking longer term to `v3` and `v4` etc).

That said I guess the set should (now) be pretty static, so I can refactor back to a single `ConditionStatus`.

manstis:
Hi @r00ta @lucamolteni if you're in preference of @robpblake proposal I have a new PR locally.

It is more simple; all be it less technically correct. 

If you're all in favour I will push instead of this one and then refactor `TemplateProvider` again.

vaibhavjainwiz:
> Hi @r00ta @lucamolteni if you're in preference of @robpblake proposal I have a new PR locally.
> 
> It is more simple; all be it less technically correct.
> 
> If you're all in favour I will push instead of this one and then refactor `TemplateProvider` again.
> 
> It should be noted there are other differences between `v1` and `v2` too (namely `CustomResourceStatus.isReady()`). Personally I don't think we can guarantee `v3` etc won't need similar differences and hence I still lean toward the technically correct implementation.

+1 @robpblake proposal to use simplier approach.

manstis:
@robpblake @vaibhavjainwiz @lucamolteni @r00ta Implementation replaced with the simplified approach.

vaibhavjainwiz (approved):

shard-operator-parent/shard-operator-core/src/main/java/com/redhat/service/smartevents/shard/operator/core/resources/CustomResourceStatus.java
```diff
@@ -38,22 +37,6 @@ public final Optional<Condition> getConditionByType(final String conditionType)
         return conditions.stream().filter(c -> conditionType.equals(c.getType())).findFirst();
     }
 
-    @JsonIgnore
-    public final boolean isReady() { // Don't use this in V2 as READY constant doesn't exist https://issues.redhat.com/browse/MGDOBR-1369
```
manstis, line 42: Only used by `v1`.

shard-operator-parent/shard-operator-core/src/main/java/com/redhat/service/smartevents/shard/operator/core/resources/CustomResourceStatus.java
```diff
@@ -38,22 +37,6 @@ public final Optional<Condition> getConditionByType(final String conditionType)
...
-        return conditions.stream().anyMatch(c -> ConditionTypeConstants.READY.equals(c.getType()) && ConditionStatus.True.equals(c.getStatus()));
-    }
-
-    @JsonIgnore
-    public final boolean isReadyV2() {
-        return conditions.stream().allMatch(c -> ConditionStatus.True.equals(c.getStatus()));
-    }
-
-    @JsonIgnore
-    public final boolean isAugmentingTrueOrUnknown() {
```
manstis, line 52: Only used by `v1`.

shard-operator-parent/shard-operator-core/src/main/java/com/redhat/service/smartevents/shard/operator/core/resources/CustomResourceStatus.java
```diff
@@ -107,10 +90,30 @@ public void markConditionTrue(final String conditionType) {
     }
 
     @JsonIgnore
-    public final void setStatusFromBridgeError(BridgeError bridgeError) {
```
manstis, line 110: Needs different implementations for `v1` and `v2`.

shard-operator-parent/shard-operator-v1/src/test/java/com/redhat/service/smartevents/shard/operator/v1/resources/CustomResourceStatusTest.java
```diff
@@ -0,0 +1,62 @@
+package com.redhat.service.smartevents.shard.operator.v1.resources;
+
+import org.junit.jupiter.api.Test;
+
+import com.redhat.service.smartevents.shard.operator.core.resources.ConditionReasonConstants;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class CustomResourceStatusTest {
```
manstis, line 9: Test specific to `v1` implementation.

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/BaseResourceStatus.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @JsonIgnore
+    public final boolean isReady() {
+        return getConditions().stream().allMatch(c -> ConditionStatus.True.equals(c.getStatus()));
+    }
+
+    @JsonIgnore
+    public final void setStatusFromBridgeError(BridgeError bridgeError) {
+        markConditionFailed(DefaultConditions.CP_DATA_PLANE_READY_NAME,
```
manstis, line 25: @lucamolteni This `Condition` name may need changing depending on what you do for "timeout".

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/CamelIntegrationStatus.java
```diff
@@ -1,48 +1,31 @@
...
 import java.util.Set;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.redhat.service.smartevents.shard.operator.core.resources.Condition;
 import com.redhat.service.smartevents.shard.operator.core.resources.ConditionStatus;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class CamelIntegrationStatus {
+public class CamelIntegrationStatus extends BaseResourceStatus {
```
manstis, line 11: Changed to extend sub-class to avoid duplication of `isReady()` logic and `getConditionByType()`

shard-operator-parent/shard-operator-v2/src/test/java/com/redhat/service/smartevents/shard/operator/v2/resources/CustomResourceStatusTest.java
```diff
@@ -0,0 +1,62 @@
+package com.redhat.service.smartevents.shard.operator.v2.resources;
+
+import org.junit.jupiter.api.Test;
+
+import com.redhat.service.smartevents.infra.v2.api.models.DefaultConditions;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class CustomResourceStatusTest {
```
manstis, line 9: Test specific to `v2` implementation.

manstis:
@vaibhavjainwiz I've added a new commit to split version specific `CustomResourceStatus` methods into version sub-classes.

Could you please re-review. I am happier with this PR now I've split the version-specific code.

r00ta (approved):

vaibhavjainwiz:
Its my point of view so just follow it if you also feel so.
`CustomResourceStatus` should only be a pure POJO class which contains `List<Conditon>` and its setter/getter method.
All other method should be a part of a utility class or service/interface(`CustomResourceStatusService`) and v1/v2 should have their specific implementation of this interface.

manstis:
Hi @vaibhavjainwiz 

> `CustomResourceStatus` should only be a pure POJO class....

I don't object; in fact I think there are a few places in RHOSE where things we serialise are not POJOs and contain logic.

I think it reasonable to consider a wider-reaching refactor of all scenarios.... but I think it's outside the scope of this PR.

vaibhavjainwiz:
> Hi @vaibhavjainwiz
> 
> > `CustomResourceStatus` should only be a pure POJO class....
> 
> I don't object; in fact I think there are a few places in RHOSE where things we serialise are not POJOs and contain logic.
> 
> I think it reasonable to consider a wider-reaching refactor of all scenarios.... but I think it's outside the scope of this PR.
> 
> IMO anything containing a method annotated `@JsonIgnore` smells like an opportunity to refactor in the way you suggest.
> 
> Search our codebase for use of `@JsonIgnore` and you'll see the extent of the issue!

yes.. there is lot of scope of refactoring in code. I didn't point it out earlier because major focus is to minimize code changes in v1 but as you are make some refactoring in v1 thats why I raise it now.
You could ignore this comment if you think its not important.

manstis:
@vaibhavjainwiz I've created: https://issues.redhat.com/browse/MGDOBR-1407

vaibhavjainwiz (approved):

ChughShilpa (approved):

shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/resources/BaseResourceStatus.java
robpblake: @manstis Please correct me if I'm wrong, but this is related to `Exception` handling and not time-out. See https://javaoperatorsdk.io/docs/features#automatic-retries-on-error and https://javaoperatorsdk.io/docs/features#setting-error-status-after-last-retry-attempt . We don't throw an `Exception` when we time out, we just mark a condition as `Failed` (or will do). Equally the `Exception` retry limit could be reached before the timeout limit.

In any case, we would need to do something like:

1. Provide a mapping between a specific `Condition` and `BridgeError` somehow; or 
2. Have the Controller decide in the implementation of `updateErrorStatus` which `Condition` should be marked as `Failed`

I don't think we have a ticket for mapping `Exception` to `Condition`, so I'd suggest we remove this from `v2` , create the ticket for mapping `Exception` to `Condition` and leave the `updateErrorStatus` method empty within the `ManagedProcessorController`.

manstis: @robpblake Correct.. but the `v2` Controllers have no exception handling really.

[One](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedProcessorController.java#L146-L154) has some legacy code from how we did it in `v1` and the [other](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/blob/main/shard-operator-parent/shard-operator-v2/src/main/java/com/redhat/service/smartevents/shard/operator/v2/controllers/ManagedBridgeController.java#L161-L164) nothing.

How exceptions are propagated from the `DataPlane` to the `ControlPlane` in `v2` remains erm, undetermined.

The existing code set `ConditionTypeConstants.READY` to `False`. Which is completely wrong in `v2`.

So I made the change in this PR and [suggested](https://github.com/5733d9e2be6485d52ffa08870cabdee0/sandbox/pull/1492/files#r1072141411) to @lucamolteni it would need review.

manstis: I can however remove the legacy use in `ManagedProcessorController`, as you propose, and create a ticket for the error mapping, also as you propose. :+1:

manstis: New commit added.

manstis: https://issues.redhat.com/browse/MGDOBR-1412

robpblake (approved):
Thanks @manstis

manstis:
@lucamolteni I removed you as a "Reviewer" as you're not very well.

You may still however be interested in the conversation on here.

